PR #2, opened 2026-06-13
Appliances: virtual devices, structured parts/wear-items, subdevices &amp; HA polish

prestomation:
Implements the full appliances plan from `IDEAS.md` — virtual-device provisioning + asset metadata (earlier commits), then **structured parts/wear-items**, **parent/child relationships**, and **HA-integration polish** (this commit `c12eb9a`).

## What & why

A "dumb" appliance (fridge, furnace, water heater) usually isn't a Home Assistant device, so there's nowhere to hang maintenance tasks or record warranty. This PR makes Home Keeper model appliances richly, managed from the **Appliances** panel tab.

### Virtual devices + metadata
Register an appliance as a real **virtual device** (multiple tasks share one device page) or attach metadata to an **existing** device from another integration. Metadata: make/model/serial, mdi icon, purchase/install/**warranty-expiry** dates, cost, vendor, manual link — date fields become `date` **sensors** on the device page (automatable).

### Parts & wear items
A structured **parts** list (name, part #, vendor, cost, *consumable* vs *wear*) replaces the old free-text string. A **wear item** with a replacement interval auto-creates a floating maintenance **task** on the appliance's device — reusing the recurrence engine + the existing to-do/calendar/mark-done/next-due entities — and completing it stamps the part's *last replaced* date. A load-time shim migrates the legacy `part_numbers` string (no storage-version bump).

### Subdevices & related devices (hybrid)
`parent_asset_id` nests a virtual appliance under another via HA's native **`via_device`** (provisioned parents-first, cycles rejected). `related_device_ids` loosely associates *any* registry device — including foreign ones HA won't let us reparent (your existing humidity sensor next to the *Piano*) — surfaced in the panel.

### HA-integration polish
- `area_id` **validated** against real HA areas at the boundary (tasks + assets).
- `configuration_url` deep-links virtual devices back to the panel. *(Deliberately **not** `entry_type=service` — appliances are physical things; SERVICE would mis-categorize them.)*
- Optional **mdi icon** per appliance.
- **`diagnostics.py`** snapshot for support.
- New task `source` field — the first internal use of the long-reserved cross-integration contribution hook.

### Screenshots

Appliance form — virtual device, parts/wear-items editor, subdevice + related-device selectors:

Add an appliance

Appliances list — "1 part", "1 subdevice", "subdevice of …" chips:

Appliances list

## Key design choices
- Pure `assets.py`/`models.py` stay HA-free; area/device-registry validation lives at the boundary (`devices.py`, handlers).
- Part-driven maintenance **reuses tasks** (no parallel "part sensor"); `store.reconcile_part_tasks` owns `source`-tagged tasks (create/update/remove) and runs at setup + after every asset mutation.
- Relationships are **hybrid** because HA only allows `via_device` on devices we own; foreign links are panel-only by necessity.

## Tests (all green locally)
- **Unit** (`tests/unit/test_assets.py`): parts/icon/relationships/migration/cycle validation — 51 pytest pass.
- **Integration** (Docker HA): wear-part → task + device button; seeded subdevice provisions + nests; related-device round-trip; area-id + bad-URL rejection — 18 pass.
- **Frontend** vitest (15) + **e2e** Playwright incl. the parts editor (7), with refreshed screenshots.

## Out of scope (noted in IDEAS.md)
Photo attribute/storage, labels for category/type, a live device-registry update/removal listener, and migrating the panel to HA web components (`ha-textfield`/`ha-select`/`ha-form`, which would enable a true icon picker) — kept as a separate follow-up per the agreed plan.


prestomation:
/q review Please be critical and skeptical, surfacing the most serious issues first and not withholding minor ones. This PR adds "appliances" — virtual-device provisioning + decoupled asset metadata. Please scrutinize:

- **Correctness:** the device-registry reconciliation in `devices.async_reconcile_assets` (idempotency across setup + every mutation, the orphan-prune predicate `_is_asset_device` vs. per-task self-owned devices, writing `device.id` back), the existing-device snapshot recovery (`_resolve_by_snapshot`), and asset-delete → `detach_tasks_from_device` lifecycle. Any path that could remove a device we don't own, or strand metadata.
- **HA best practices:** `async_get_or_create`/`async_update_device` usage and version compatibility (the `_supports_kwarg` guard for `serial_number`), reload-vs-refresh choices in the websocket/service handlers, config-entry-owned device cleanup, and `async_remove_entry`.
- **Maintainability:** the two-concern split (`assets.py` pure model vs. `devices.py` provisioning), duplication between service and websocket asset handlers, and the panel's growing `panel.ts`.
- **Security:** `escapeHTML` coverage on the new appliance form/cards, and validation of date/cost/URL fields.
- **Performance:** reconcile running a full pass on every mutation + a follow-up entry reload (redundant work / N+1 registry calls).

---

prestomation:
Thanks for the thorough review — pushed `86eeb6d` addressing the valid findings.

**Fixed**
1. 🔴 **Persistence in `_reconcile_existing`** — real data-loss bug (existing-device assets start with an empty snapshot, so it was *never* persisted). It now refreshes the `identifiers`/`connections` snapshot + recovered `device_id` and returns a dirty flag; `async_reconcile_assets` flushes once via the new `store.async_persist()`. New integration test asserts the snapshot is persisted.
2. 🔴 **`_is_asset_device` / prune safety** — uuid4 task ids can't collide with the `asset_` prefix, but added an explicit guard so a device matching any current task's self-owned identifier is never pruned.
3. 🟠 **`_supports_kwarg`** — now returns `False` on signature-lookup failure (conservative).
4. 🟠 **URL/cost validation** — `assets._normalize_url` rejects non-http(s) schemes + over-length; cost rejects negatives; threaded through `merge_update`. 6 new unit tests + an integration test for bad-URL rejection.
5. 🟡 **area clear** — dropped the truthy-only guard so clearing an area propagates to the device.
6. 🟡 **Handler duplication** — extracted `devices.async_apply_asset_change` used by both service and websocket add/update.

**Intentionally deferred (with reasoning)**
- **Incremental reconciliation (#3 perf):** assets are explicitly a small set here ("typically only a handful" per AGENTS.md), the double pass is idempotent (get_or_create returns the existing device; snapshot writes skip when unchanged), and the reload-to-refresh-entities pattern matches how tasks already work. Targeted single-asset reconciliation is a worthwhile follow-up but not warranted at this scale.
- **Splitting `panel.ts`:** agreed it's growing; I'd rather do the `task-form.ts`/`asset-form.ts` split as a focused refactor PR than enlarge this diff and risk the working UI.
- **Coverage comment:** no action — the new HA-integration modules (`devices.py`, store/ws/`__init__`) are exercised by the Docker integration suite, not the unit-coverage gate, consistent with the pre-existing 0% there; the pure `assets.py` model is 95%+.

/q review Please re-review commit 86eeb6d, focused on the changed areas: correctness of the `_reconcile_existing` persistence + dirty-flag flow and the hardened orphan-prune guard in `devices.py`, and the new `manual_url`/cost validation in `assets.py`. Be critical and surface the most serious issues first.

---

prestomation:
/q review Please be critical and skeptical, most-serious-first, on commit c12eb9a which adds structured parts/wear-items, subdevices, and HA-integration polish to the appliances feature. Focus on:

- **Correctness:** `store.reconcile_part_tasks` lifecycle — does it correctly create/update/remove `source`-tagged tasks across asset edits, part removal, and asset deletion without orphaning or duplicating tasks? `merge_update`'s part merge-by-id preserving `last_replaced`. The `via_device` provisioning in `devices._reconcile_virtual` (parents-first `_ancestor_depth`, `via_device` on create vs `via_device_id` on update, re-parent/un-parent) and the cycle guard `assets.would_create_cycle`.
- **HA best practices:** `area_exists` boundary validation, `configuration_url`, not setting `entry_type=service`, and whether `via_device_id=None` to un-parent is safe.
- **Maintainability:** the growing `panel.ts` parts-editor (sync-before-rerender), and the `assets.py` model surface.
- **Security:** `escapeHTML` coverage on the new parts/related/parent form fields and card chips; icon/url/cost validation.
- **Performance:** `reconcile_part_tasks` + device reconcile + entry reload on every asset mutation.

Surface the most serious issues first and don't withhold minor ones.

---

prestomation:
Thanks — triaged in `579ba88`.

**Fixed**
- 🟡→🐛 **Dangling `area_id`** (caught a real latent crash, not just the medium you filed): if an area is deleted in HA after assignment, `async_update_device` rejects the unknown area and would break reconciliation. `_reconcile_virtual` now self-heals a missing area to `None`.
- 🔴 **Un-parent visibility:** warn when the HA version lacks `via_device_id` update support and a subdevice link is requested (parenting on *create* still works via the long-supported `via_device` identifier; only post-create re-parent needs the kwarg).
- 🟡 **Corrupt parent chain:** `_ancestor_depth` now logs an error if it ever sees a cycle (`would_create_cycle` prevents it; defensive).
- 🟡 **Maintainability:** documented the parts-editor sync-before-rerender contract on `_syncAssetFormToState`.
- 🟠 **Part-removal proof:** added an integration test — editing an asset to drop a wear part removes its derived task.

**Respectfully pushing back (no change)**
- 🔴 **Orphaned task on update** — not reproducible: both `ws_update_asset` and the `update_asset` service flow through `devices.async_apply_asset_change`, which calls `store.reconcile_part_tasks()` before reload, so removing/retyping a wear part deletes its `source`-tagged task. The new test confirms this.
- 🟠 **Missing `escapeHTML` on part fields** — false positive: `part_number`, `vendor`, `cost`, `last_replaced`, `name`, `id` are all wrapped in `escapeHTML` in `_partRowHTML` (panel.ts:628–666).
- 🟠 **`_merge_parts` drops `last_replaced` for a new id** — intentional: a new part id is a new part with fresh history.
- **Performance** (reconcile + reload per mutation) — accepted at this scale (a handful of appliances), consistent with how task edits already reload; selective reconcile is noted as a follow-up in IDEAS.md.

/q review Re-review 579ba88, focused on the `_reconcile_virtual` area-id self-heal and the un-parent warning path in devices.py, and confirm the part-removal reconciliation is correct. Most-serious-first, please.

---

prestomation:
Ran a round of **exploratory testing** — drove the websocket API (the panel's path, which bypasses the service vol-schema and goes straight through the pure model + reconciliation) across ~19 edge cases, plus registry/diagnostics checks.

**Found one real bug (fixed in `0364aac`):** wear-part `next_due` **drift**. `reconcile_part_tasks` passed `interval`/`unit` to `merge_update` on every reconcile, re-triggering a floating `next_due` recompute; for a wear part with no `last_replaced` (`last_completed is None`), `compute_floating_next_due` anchors to *now*, so the due date crept forward on every asset edit and every HA restart — the task could perpetually never come due. Fix: only pass fields that actually changed, so `next_due` recomputes only on interval/unit change or completion. Added a regression test (next_due stable across a rename).

**Everything else checked out:**
- Relationships: set/clear parent, **cycle rejected**, nonexistent-parent rejected, existing-kind parent forced `None`; confirmed the subdevice's `via_device_id` actually points at the parent device in the registry.
- Validation: invalid icon, negative part cost, bad `last_replaced`, `replace_interval=0`, missing part name, bad part type, unknown area — all rejected with clear messages.
- Lifecycle: completion stamps `part.last_replaced`; two wear parts → two tasks; removing one part → one task; related-device round-trip.
- Polish: `configuration_url` set; `entry_type` correctly `None` (physical); diagnostics responds (200 via GET — my first probe used POST → 405, a probe artifact, not a bug).

Also hardened the two new part-task tests to scope by `asset_id` and self-clean (they were brittle against the shared integration container's accumulated state; CI's fresh container was always fine).

All green locally: **51 unit, 20 integration, 15 vitest**.

/q review Re-review `0364aac` — focus on the `reconcile_part_tasks` change-detection update path (does it correctly avoid spurious `next_due` recomputes while still updating name/interval/unit/device/area, and still remove orphaned tasks?). Most-serious-first, please.

---

prestomation:
/q review Please perform a fresh, critical, skeptical end-to-end review of the latest commit (0364aac) on this PR — the full appliances feature: virtual-device provisioning + asset metadata, structured parts/wear-items (a wear item drives a derived maintenance task via `store.reconcile_part_tasks`), subdevices via `via_device` + panel-only related devices, and HA polish (area_id validation, configuration_url, mdi icon, diagnostics.py).

Surface the most serious issues first. Please scrutinize:
- **Correctness:** the `reconcile_part_tasks` change-detection update path (avoids spurious `next_due` recomputes while still syncing name/interval/unit/device/area and removing orphaned tasks); part `last_replaced` stamping on completion; the legacy `part_numbers` migration shim; relationship lifecycle (cycle rejection, parents-first `via_device` provisioning, dangling-link cleanup on delete).
- **Edge cases:** timezone/DST in part-task scheduling, asset/part delete races, existing-device asset re-targeting, area-deletion self-heal.
- **HA best practices**, **security** (panel `escapeHTML`, input validation), **maintainability**, and **performance** (per-mutation reconcile + reload).

Don't withhold minor issues, but lead with anything that could cause data loss or incorrect scheduling.

---

PR #7, opened 2026-06-14
Disambiguate device-page task entities and enrich panel device chips

prestomation:
## Problem

Two UI issues raised against device attachment:

1. **Device-page "Mark done" buttons were indistinguishable.** When multiple Home Keeper tasks are attached to the same existing device, each task's per-task entities all shared one name (e.g. several buttons named *"Mark done"*), so there was no way to know which control completed which task.
2. **Sidebar panel device chip was a dead end.** The chip showing a task's/appliance's device name did nothing when clicked and carried no image.

### Backend — per-task entity naming (`button.py`, `sensor.py`, `binary_sensor.py`, `coordinator.py`)
- All three per-task entities (mark-done button, next-due sensor, overdue binary sensor) now prefix their **translated** name with the task name when the task merges onto an **existing** device — e.g. *"Replace filter: Mark done"*.
- Self-owned task devices are unchanged (the device is already named after the task, so a prefix would be redundant). A new `HomeKeeperCoordinator.task_uses_existing_device()` helper makes the distinction.
- The prefix is delivered via HA's `translation_placeholders` (separator baked into the placeholder value, empty for self-owned devices), so names stay localized in all 16 locales and update on task rename.

### Frontend — panel device chips (`panel.ts`, `utils.ts`, `api.ts`, `types.ts`)
- Device chips on task and appliance rows now **link to the device's HA page** (`/config/devices/device/<id>`) on click and via keyboard (Enter/Space).
- Chips show the device's **integration brand logo** (`brands.home-assistant.io`), resolved by mapping the device's config entry → integration domain (one cached `config_entries/get` call). Falls back to the generic `_/` brand path and then a generic device icon if no image loads.

## Screenshots

**Sidebar panel — task rows now carry an actionable device chip (icon + device name):**

Panel task list with device chips

Device chip close-up

**Device page — three tasks attached to one device, each now named by its task instead of colliding on "Mark done":**

Device page with task-named entities

> The chips render the generic device icon here because the seeded devices are Home Keeper's own virtual appliances (domain `home_keeper`, which has no `brands.home-assistant.io` logo) and the e2e sandbox has no outbound network — on a real branded device (a Hue light, a Shelly relay, …) the chip shows that integration's logo.

Captured with a committed one-off Playwright harness (`tests/e2e/screenshots-device-attach.capture.ts`), per `AGENTS.md`.

## Tests
- Added frontend unit tests for the new `deviceDomain` / `brandLogoUrl` helpers.
- Existing Python unit tests, frontend tests, `tsc`, and the rollup build all pass.
- Existing integration assertions (`test_lifecycle`, `test_assets`) remain valid — self-owned device entity_ids are unchanged, and the substring-based existing-device assertions still hold (and now exercise the new prefix path on virtual-asset devices).

## Notes
- The localized entity name change alters generated `entity_id`s only for tasks newly attached to a real existing device with siblings; existing entity_ids for self-owned task devices are preserved.

https://claude.ai/code/session_01BB4ZxjHKhwjvB6hP8kkMMW


prestomation:
/q review Please be critical and skeptical, surfacing the most serious issues first but not withholding minor ones. Focus on:

- **HA best practices / correctness**: overriding `translation_placeholders` as a plain `@property` (vs the base `cached_property`) on the button/sensor/binary_sensor — is this the idiomatic way to vary an entity name per task, and does the device-page friendly name actually refresh on task rename? Also check that the empty-placeholder path for self-owned devices produces clean names and stable `entity_id`s (no double device-name prefix).
- **Edge cases**: `task_uses_existing_device()` when `device_id` points at a now-deleted device; behavior when a task has no name; mixed existing/self-owned tasks on reload.
- **Frontend**: the `config_entries/get` call adds a websocket round-trip on panel load — is `primary_config_entry`/`config_entries[0]` a reliable domain source? Is the brand-logo `onerror` fallback chain (brand → `_/` brand → `ha-svg-icon`) sound, and the `location-changed` navigation event the right HA pattern from a custom panel shadow root?
- **Maintainability**: duplication of the placeholder logic across the three entity classes.

---

PR #13, opened 2026-06-15
Home inventory (insurance) export + consumable spare-stock tracking

prestomation:
Adds two new use cases to `IDEAS.md` (brainstorm) and ships the two flagged as
"implementing".

## 1. Home inventory for insurance (CSV export)
The appliance registry already holds make/model/serial, purchase + warranty dates,
and cost — i.e. a home inventory. This rolls it up and exports it.

- New pure module `inventory.py`: `build_inventory()` (per-appliance rows + rolled-up
  totals, warranty-active flag, on-hand spares value) and `inventory_to_csv()`.
- Exposed as **both** a `home_keeper.export_inventory` **service** (returns the report
  + CSV) and a `home_keeper/export_inventory` websocket command.
- **Export inventory** button on the Appliances tab downloads
  `home-keeper-inventory-<date>.csv`.

## 2. Consumable spare-stock tracking
Parts already carry part numbers; now they can track how many spares are on hand.

- Parts gain optional `stock` and `reorder_at`. Completing a wear-part replacement
  **consumes one spare**; at/below the threshold a `home_keeper_part_low_stock` event
  fires so users automate the shopping-list add / reorder themselves.
- `home_keeper.adjust_part_stock` **service** (and matching websocket command) for
  restock/consume — clamped at zero; only a *decrease* that crosses the threshold nags.
- Panel: stock / reorder-at fields in the part editor; an in-stock count and a **Low
  stock** chip on the appliance detail page.
- `_merge_parts` preserves backend-managed `stock`/`reorder_at` across partial edits.

## Convention added (per review)
Per feedback, **every data action is now exposed as a `home_keeper.*` service** (the
interoperability surface for automations/other integrations), not websocket-only; the
websocket commands delegate to the same store methods. This is written into
`AGENTS.md` and `.amazonq/rules/architecture-and-code.md`, and the PR-screenshot rule
is strengthened to a hard gate.

## Screenshots

**Appliances tab — Export inventory button**
Appliances list

**Part editor — Stock / Reorder-at fields**
Appliance create form

**Appliance detail — In-stock count + Low stock chip**
Appliance detail

## Tests
- New `tests/unit/test_inventory.py` (totals, spares value, warranty flag, CSV, sorting,
  malformed-date) and stock cases in `tests/unit/test_assets.py`
  (normalization, `consume_part_stock`, `adjust_part_stock`, `part_is_low`, merge).
- New `tests/unit/test_events.py` low-stock payload cases.
[...]

PR #14, opened 2026-06-15
Fix code-review findings: stock tracking + inventory export

prestomation:
Follow-up to #13 addressing all findings from the critical code review (the subagent's report + my validation).

## High
- **Services rejected `stock`/`reorder_at` on parts.** `__init__.py` `_PART_SCHEMA` (default `PREVENT_EXTRA`) omitted the two fields, so `home_keeper.add_asset`/`update_asset` errored when a part carried stock — while the websocket path accepted them and `services.yaml` advertised them. Added both as optional coerced ints. Integration test added.
- **CSV formula injection in the inventory export.** `inventory._cell` now prefixes a cell starting with `= + - @ \t \r` with an apostrophe so spreadsheets render it as literal text. Our numeric columns are non-negative, so real numbers are untouched. Unit test added.

## Medium
- **Stock couldn't be un-tracked.** `_merge_parts` no longer preserves `stock`/`reorder_at` on omit (only `last_replaced` stays backend-managed), so clearing the field actually clears tracking. Incoming wins. Tests updated.
- **Low-stock event was level-triggered.** `consume_part_stock`/`adjust_part_stock` now return `True` only on the not-low→low **crossing**, so the event fires once per crossing — not on every decrement while already low, and not on restock. Store firing simplified accordingly. Unit tests updated + a real-event integration test (captures `home_keeper_part_low_stock` via a `configuration.yaml` automation).

## Low
- Inventory totals accumulate **raw** spares and round once (no cent drift from re-rounding rows).
- Panel `_downloadFile` defers `URL.revokeObjectURL` and appends/removes the anchor; export failures now surface a toast (`error.exportFailed`, translated across all 16 locales).

## Verification (all local, green)
`pytest` unit **151** · integration **34** (incl. the new event-firing + service-schema tests, run against a fresh HA container) · `vitest` **78** · `ruff` + `tsc --noEmit` clean · panel builds.

https://claude.ai/code/session_013Qi7HFqqZrS5hmMBXbYKEK

---
_Generated by [Claude Code](https://claude.ai/code/session_013Qi7HFqqZrS5hmMBXbYKEK)_

### DIFF
--- a/custom_components/home_keeper/__init__.py
+++ b/custom_components/home_keeper/__init__.py
@@ -82,6 +82,8 @@ _PART_SCHEMA = vol.Schema(
         vol.Optional("replace_interval"): vol.Coerce(int),
         vol.Optional("replace_unit"): cv.string,
         vol.Optional("last_replaced"): cv.string,
+        vol.Optional("stock"): vol.Coerce(int),
+        vol.Optional("reorder_at"): vol.Coerce(int),
     }
 )
 
--- a/custom_components/home_keeper/assets.py
+++ b/custom_components/home_keeper/assets.py
@@ -234,28 +234,22 @@ def _normalize_parts(value: Any) -> list[dict]:
     return parts
 
 
-# Part fields the backend owns and keeps across an edit when the caller omits them:
-# ``last_replaced`` is stamped on completion, and ``stock``/``reorder_at`` are
-# maintained by stock adjustments, so a partial update (e.g. a service call that
-# resubmits a part without them) must not silently wipe them.
-_PRESERVED_PART_FIELDS = ("last_replaced", "stock", "reorder_at")
-
-
 def _merge_parts(existing: list[dict], incoming: list[dict]) -> list[dict]:
-    """Carry backend-managed part fields across an edit.
+    """Carry the backend-managed ``last_replaced`` across an edit.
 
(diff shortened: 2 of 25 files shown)


prestomation:
/q review Critical, skeptical pass please — most serious issues first, don't withhold minor ones. This PR fixes review findings from #13; focus on whether the fixes are correct and complete:

- **Edge-triggered low-stock**: `consume_part_stock`/`adjust_part_stock` now return True only on the not-low→low crossing (`part_is_low and not was_low`). Check the boundary cases (already-low → consume to 0; restock landing exactly on threshold; untracked `stock=None`; `reorder_at=None`) and that `store.adjust_part_stock`/`_stamp_part_replacement` fire exactly once per crossing.
- **CSV formula-injection guard** (`inventory._cell`): is prefixing `= + - @ \t \r` with `'` sufficient and free of false positives on our numeric columns? Any other sink (newlines embedded mid-field)?
- **Service schema**: adding `stock`/`reorder_at` to `_PART_SCHEMA` — correct coercion/validation parity with `_normalize_part`?
- **`_merge_parts` change**: stock/reorder_at now incoming-wins (clearable). Any data-loss risk for a partial service caller that resubmits a part without stock, and is that acceptable/expected?
- **Frontend**: `_downloadFile` revoke timing + the `hass-notification` toast event name; any leak.

---

PR #52, opened 2026-06-19
Sync `problem` binary sensors as un-completable tasks (opt-in)

prestomation:
## What & why

A new **opt-in config option** that automatically mirrors every `binary_sensor` with `device_class: problem` as a Home Keeper task — so a real-world problem surfaced by any integration (leak detector, appliance fault, UPS on battery, printer error) becomes a visible, trackable to-do without writing an automation.

The key behavior, per the request: **these synced tasks can't be cleared inside Home Keeper.** The originating integration has to resolve the actual problem (the sensor returns to OK), at which point Home Keeper clears the task automatically.

## How it works

- Each eligible sensor is mirrored as a **condition-driven (`triggered`) task**: **armed** (due-now on the to-do list, calendar and device page) while the sensor reports a problem, **dormant** when it clears, **re-armed** if it recurs. Completion history accumulates.
- Tasks are tagged `source={"problem_sensor":{entity_id}}` with a `managed_by` block carrying `completion_blocked: true`. The **Done** action is hidden on **every** surface (panel, dashboard card, to-do list, per-task device button), and the `complete_task` / `trigger_task` service + websocket **reject** them with a clear message unless the call carries the internal sync origin.
- Each task **inherits the sensor's device and area**, so it appears on the device's page.
- **Scope it:** syncing is **off by default**; once on, exclude specific **entities / areas / labels** in the options screen. A **loop guard** skips Home Keeper's own entities so its per-task overdue sensors can't feed back in.

## Architecture

- `problem_tasks.py` — **pure** create/arm/clear/orphan reconciler over plain dicts (unit-tested in isolation), wrapped by `store.reconcile_problem_sensor_tasks` (persist + fire the existing `task_created`/`triggered`/`completed`/`deleted` events; completion carries `origin: home_keeper_problem_sensor_sync`).
- `problem_sync.py` — HA-aware: entity-registry enumeration + a state-change listener + entity-registry-update listener; reloads the entry only when the synced **task set** changes (create/remove), otherwise a plain coordinator refresh for arm/clear.
- `config_flow.py` — `OptionsFlow` (toggle + entity/area/label exclusions); an options change reloads the entry.

## Tests & validation (run locally)

- `pytest tests/unit` — **247 passed** (incl. new `test_problem_tasks.py`: create/arm/clear/orphan/metadata + source helpers; translation parity).
- `vitest` — **137 passed**.
- Integration suite against Docker HA — **green** (new `test_problem_sync.py`: the template `binary_sensor.sump_pump_problem` is mirrored as an armed managed task and `complete_task` is rejected).
- Full Playwright **e2e suite — 23 passed** with syncing enabled in the test config.

## Docs & i18n

README feature section, `docs/EVENTS.md` note, CHANGELOG (`0.3.0b7`), `.amazonq/rules`, plus `strings.json` options-flow strings localized across all **16 locales**.

## Screenshot — synced problem-sensor task detail

A synced "Sump pump problem" task: *Managed by Home Keeper* chip, the prompt explaining it clears externally, and **no Done button**.

Synced problem-sensor task detail — armed and due-now, Managed by Home Keeper, with no Done button and the prompt explaining it clears when the source resolves it

https://claude.ai/code/session_01NMv8XzveiMXntBxwveJSu6

---
_Generated by [Claude Code](https://claude.ai/code/session_01NMv8XzveiMXntBxwveJSu6)_


prestomation:
/q review Please give critical, skeptical feedback on this problem-sensor sync feature, surfacing the most serious issues first (and don't withhold minor ones).

Focus areas:
- **Correctness / edge cases:** the `problem_sync.py` listener + reconcile loop — race between the initial reconcile (runs before the template/platform sensors register) and the entity-registry/state listeners that recreate the task with a fresh id; the reload-vs-refresh decision in `_async_reconcile` and the `_reload_scheduled` guard (can it drop a needed reload or thrash?); arming/clearing semantics in the pure `problem_tasks.reconcile_problem_tasks`; the loop guard that skips `platform == DOMAIN` entities (is it sufficient to prevent our own overdue binary sensors feeding back?); state being `unknown`/`unavailable` treated as not-a-problem.
- **HA best practices:** options-flow + `add_update_listener` reload pattern; `async_track_state_change_event` re-subscription on registry changes; using `async_create_task` for reloads from a callback; selector usage.
- **Maintainability:** module boundaries (pure `problem_tasks` vs HA-aware `problem_sync` vs `store`), the `ORIGIN_PROBLEM_SENSOR_SYNC` completion-guard approach, naming, duplication with the wear-part reconciler.
- **Performance:** enumerating the full entity registry on every tracked sensor state change — is that a concern for large installs, and should it be debounced/narrowed?
- **Security:** anything around the un-completable guard being bypassable via the service `origin`.

---

PR #53, opened 2026-06-19
Show synced problem tasks' Done as disabled with an explanation

prestomation:
## What & why

For an un-completable task (a synced `device_class: problem` sensor), the *Done* action was **hidden** — which reads as "missing" rather than "blocked" and left users wondering where Done went. This shows the button **disabled** instead, and tapping it pops up a short explanation that the problem clears automatically once the originating integration resolves it.

## How

- Render the blocked Done as a natively-**`disabled`** (greyed) `<ha-button>` wrapped in a clickable `<span>` (the native `disabled` greys it correctly across HA's Material-3 button versions, but swallows clicks — so the span carries the tap → toast, plus a hover `title` tooltip). Applied to the **panel task card (Tasks list)**, the **task detail page**, and the **dashboard card** — all via the shared `_blockedDone` helper, so the behavior is identical on every surface.
- The popup reuses the task's `managed_by.completion_prompt`; adds a `done.blocked` fallback string across all **16 locales**.
- No backend change — the store/service/websocket guards already reject completion; this is purely the UI affordance.

## Tests (run locally)

- `vitest` — **137 passed** (i18n parity + untranslated guards hold with the new key).
- Full Playwright **e2e suite** passes, including a new `problem-sensor.spec.ts`: the **Tasks-list** card's Done is a disabled `ha-button`, clicking it surfaces the explanation toast, and the task is **not** completed.
- `tsc` clean, `ruff` clean, `pytest tests/unit` **247**.

## Rebased on #54 — now `0.3.0b10`

Rebased on top of the merged #54 (Settings tab) and re-versioned `b8` → **`0.3.0b10`** so it releases *above* the already-published `0.3.0b9` (otherwise the fix wouldn't reach beta users). Also stabilized the `set_options` reload test against a concurrent-reload race (test-only; verified 3/3 locally).

## Screenshots

Disabled *Done* on the synced problem-sensor **detail page**, and tapping it:

Synced problem-sensor task detail with a disabled Done button

Tapping the disabled Done pops up a toast explaining the problem clears when the originating integration resolves it

The same disabled *Done* on a **Tasks-list card row** (clicking it pops the same explanation):

The Tasks list with the synced task's Done greyed/disabled and the explanation toast

https://claude.ai/code/session_01NMv8XzveiMXntBxwveJSu6


prestomation:
/q review Please give critical, skeptical feedback on this small UI change (most serious issues first, but don't withhold minor ones).

Focus areas:
- **Correctness / accessibility:** the "disabled `<ha-button>` wrapped in a clickable span" pattern — is the span keyboard-accessible (it has role="button"/tabindex but no keydown handler — should Enter/Space trigger the popup)? Does `pointer-events: none` on the inner button reliably route taps to the span across HA's Material-3 button? Any chance the disabled button still steals focus/clicks on some platforms?
- **Maintainability:** the `_blockedDone` helper + dual wiring (generic `.done-blocked-wrap` for the list vs. the explicit detail wiring in `_wireDetailActions`) — is there a cleaner single wiring point? Naming/duplication between panel and card.
- **Consistency:** the dashboard card greys the icon via `color` and stays clickable (not `disabled`), while the panel uses a disabled button + wrapper — is that inconsistency worth unifying?
- **i18n:** the `done.blocked` fallback vs. always-present `completion_prompt`.

---

PR #55, opened 2026-06-20
Filter the dashboard card by label (one card per subject)

prestomation:
## What & why

Most homes have natural "buckets" of upkeep that don't map onto a room or a single device: **the dog, the car, home maintenance, each kid's chores**. The dashboard card already filtered by status/area/device/recurrence, but a dog isn't a device and a kid isn't an area. This adds **label-based filtering** so you can point a card at a subject and get a focused list — one card per dog/car/kid.

## How it works

A task matches a label filter if the label is on **the task itself**, on its **attached device**, or on its **effective area** (own area, else its device's). This union resolution is the key: a Home Keeper **virtual appliance is a real HA device**, so labelling it in *Settings → Devices* automatically pulls its tasks into the matching card — a subject never has to be modelled as an HA area/device. Labels reference HA's shared label registry, so `dog` is the same label on a task and on a device.

Home Keeper card filtered to the Dog label, showing only Buddy and Rex tasks with Dog label chips

## Changes

**Backend**
- `models`: tasks gain a normalized `labels` list (de-duped, blank-stripped). `merge_update` only rewrites it when the caller sends `labels`, so a plain rename never stamps a phantom `labels` change.
- `home_keeper.add_task` / `update_task` accept a `labels` field (label selector, English `services.yaml` strings like the existing `source` field — no `strings.json` burden). The value is echoed on every task event and documented in `docs/EVENTS.md`.

**Frontend**
- `card-filter.ts` (pure): `taskLabelIds` union resolver + `labels` / `label_match` (any/all) config.
- Card editor: **Limit to labels** picker, **Any/All** match mode, **Show labels** toggle. Rows render label chips resolved to registry names (the card fetches `config/label_registry/list`, since HA doesn't eagerly populate `hass.labels` on a dashboard — device/area label *membership* via `hass.devices`/`hass.areas` is available and drives filtering).
- Shared task form (panel + card) gains a **Labels** field; `field.labels` added to all 16 locales.

**Tests / docs**
- Unit: label normalization (`test_models`), event payload (`test_events`), and card filter union/any/all (`card-filter.test.js`).
- e2e: a regression spec asserting the label-filtered card shows only labelled tasks with named chips, plus a capture step; seeded `dog`/`home` labels + a `core.label_registry` fixture.
- README "Filter by label" section, CHANGELOG, and `.amazonq/rules` convention note.

## Verification

- `pytest tests/unit` → 252 passed
- `vitest` → 140 passed
- Playwright `tests/card.spec.ts` (incl. new label test) → 5 passed; screenshot captured

https://claude.ai/code/session_01ADdwaqb2Xh5RNQTU9VmGtT

---
_Generated by [Claude Code](https://claude.ai/code/session_01ADdwaqb2Xh5RNQTU9VmGtT)_


prestomation:
/q review Please give this a critical, skeptical review and surface the most serious issues first (don't withhold minor ones). Focus areas:

- **Correctness/edge cases**: the label union resolution in `card-filter.taskLabelIds` (task ∪ device ∪ effective-area), any/all matching semantics, and `models.normalize_labels` / `merge_update` (does a plain rename ever spuriously change `labels`? are pre-feature tasks without the key handled?). Also the `filterTasks` signature change (`now` at position 4, `areas` appended at 5) — any positional-call hazards?
- **HA best practices**: storing HA label-registry ids in our own JSON (dangling ids when a label is deleted), fetching `config/label_registry/list` directly because `hass.labels` isn't eagerly populated — is there a better/standard approach? Service field shape (`labels` via label selector) and event payload addition.
- **Maintainability**: the `labels` field threaded through models/services/events/forms/card; chip rendering and the `_loadLabels` lifecycle in the card.
- **Performance**: label resolution per task in the card filter hot path.
- **Security**: label ids are user/registry-controlled strings rendered into chips — confirm escaping is correct.

---

PR #57, opened 2026-06-20
Settings exclusions take effect immediately

prestomation:
## Problem

In the panel's **Settings** tab, adding a problem-sensor **entity / area / label** to a *skip* list didn't take effect right away. Two reasons:

1. **Backend** — the options write went through `async_update_entry`, which only *schedules* a fire-and-forget entry reload (the reload re-runs the problem-sensor reconcile that creates/removes synced tasks). The service / websocket call returned before that reconcile happened.
2. **Frontend** — `_saveOptions` never re-read the panel's task list after saving, so an excluded sensor's synced task lingered in the Tasks tab until something else triggered a refresh (or a full page reload).

## Fix

- **`options.async_set_options` now awaits the reload itself**, so the reconcile has run by the time the `set_options` service / `home_keeper/set_options` websocket call returns. It flags the entry (`caller_is_reloading`) for the duration so the config-entry update listener doesn't fire a *second*, overlapping reload. The options **flow** (which updates the entry directly) still relies on the update listener, so its behavior is unchanged.
- **`_saveOptions` re-`_reload()`s its cached tasks** after a successful save — *without* re-rendering the form the user is still editing — so the Tasks tab reflects the exclusion the moment they return to it.

## Tests (red/green)

- **Integration** (`tests/integration/test_set_options.py`): excluding the sensor removes its synced task on the **very next read** (no polling), and clearing the exclusion recreates it.
- **Frontend** (new `test/panel.test.js` + a vitest harness): saving a Settings exclusion **re-fetches tasks**. Verified red without the `_reload()` (no refetch → fail) and green with it. Adds a `panel-version` alias to `vitest.config.js` (it's a Rollup virtual module at build time) so `panel.ts` imports cleanly under vitest.
- Full suites pass locally: `vitest` 141 ✓, `pytest tests/unit` 252 ✓, integration `test_set_options.py` 2 ✓, `tsc --noEmit` clean, `ruff` clean.

## Screenshot — Settings tab (unchanged surface, captured fresh)

Home Keeper panel Settings tab

---
_Generated by [Claude Code](https://claude.ai/code/session_012izxV55qxGApq6s2Touf3f)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. Focus on:

- **Correctness / concurrency**: the `_CALLER_RELOADING` guard in `options.async_set_options` — is awaiting `async_reload` from the `set_options` service/websocket handler safe re: re-entrancy and the update-listener's own scheduled reload? Any way the flag leaks or a double/overlapping reload still slips through (e.g. concurrent saves for the same entry, or the listener task running after the `finally`)? Is the `merged == base` no-op short-circuit correct given `current_options` normalization?
- **HA best practices**: awaiting a full config-entry reload inside a service call and a websocket command — acceptable, or should this be structured differently?
- **Frontend**: `_saveOptions` calling `_reload()` (data refresh without re-render) while the user may still be editing the autosaving `ha-form` — any stale-state or race concerns, and is skipping the re-render the right call?
- **Maintainability**: the module-level `_CALLER_RELOADING` set + `caller_is_reloading` coupling between `options.py` and `__init__._async_options_updated`.
- **Tests**: does the integration assertion (synced task gone on the next read, no polling) actually pin the behavior, and is the new panel vitest harness sound?

---

PR #60, opened 2026-06-20
Add Assist voice intents to complete and query tasks

prestomation:
## What & why

Competitive research flagged that we should let users **clear tasks via Home Assistant intents**. This adds voice/text control through Assist: two conversation intents so chores can be completed and checked hands-free instead of opening the panel.

- **`HomeKeeperCompleteTask`** — *"mark the furnace filter as done"*, *"I just replaced the fridge filter"*, *"clear take medicine"*, *"check off the water filter"*. Fuzzy-matches the spoken name against your tasks and completes it. When more than one task could match (you have a *fridge* filter and a *furnace* filter and only said "filter"), it **lists them and asks which one** rather than guessing — completing the wrong task would advance its recurrence, a destructive surprise.
- **`HomeKeeperListDueTasks`** — *"what chores are due?"*, *"what tasks need to be done?"* — reads back tasks that are due or overdue.

## Design

- **Delegates to the store, not a new path.** Both handlers funnel through the existing `store.complete_task` chokepoint with `origin=ORIGIN_INTENT`, so a voice completion stays observable (fires `home_keeper_task_completed`) and obeys the same guards — a problem-sensor-synced task is still rejected, now with a spoken explanation. No new service needed (`complete_task` already exists).
- **Pure matching.** Name resolution + due-task selection live in `task_match.py` (no HA imports, loaded in the `hk` synthetic package), unit-tested in isolation like `recurrence.py`. The handler only does HA-specific work: registry name→id lookups, spoken responses, error mapping.
- **Reload-safe.** Registration is global and idempotent (keyed by `intent_type`) and resolves the coordinator per-call.

## Tests

- **Unit** — `tests/unit/test_task_match.py` (exact/fuzzy/ambiguous/not-found, area & device scoping, due selection) and `tests/unit/test_intent_sentences.py` (loads the shipped `custom_sentences/en/home_keeper.yaml` and runs it through **hassil**, the same engine HA uses, to assert sentence→intent→slot routing — this is the only tier that covers the NLU layer).
- **Integration** — `tests/integration/test_intents.py` drives the real `/api/intent/handle` endpoint (enabled `intent:` in the test config): completion + `origin` marker via the existing capture automation, unknown/ambiguous (no mutation), problem-sensor rejection, due-query, and **reload survival**.
- Ran locally before pushing: **279 unit passed**, **48 integration passed** (incl. 6 new), `ruff check` + `ruff format --check` clean.

## Setup for users

Intent handlers register automatically. To teach the default Assist agent the phrases, copy `custom_sentences/en/home_keeper.yaml` into `<config>/custom_sentences/en/home_keeper.yaml` and restart. Documented in README "Voice control" + CHANGELOG, and the convention is recorded in `.amazonq/rules/architecture-and-code.md`.

## Known gaps (deliberate)

- **Multi-turn disambiguation** isn't a true conversation follow-up — it speaks the candidates and asks you to re-utter a more specific command.
- **Localization**: sentences/responses ship in English (responses are in-code); the matcher is language-agnostic.
- Fuzzy-match **threshold** is tuned conservatively (prefers asking over guessing); no per-language tuning.

No panel UI surface changed, so no screenshots are required for this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_013ytrgbbCH5n5mrrhzQW1tg)_

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -105,6 +105,31 @@ reviewing code in this repository (the `home_keeper` Home Assistant integration)
 - Read-only/report actions use `SupportsResponse.ONLY`; data mutations reload the
   entry or refresh the coordinator exactly as the equivalent CRUD service does.
 
+## Conversation/voice intents delegate to the store — they are not a new path
+- Assist intents live in `intents.py` (`intent.IntentHandler` subclasses, registered
+  via `intent.async_register` from `async_setup_intents`, called in
+  `async_setup_entry` after `_register_services`). Registration is **global and
+  idempotent** (keyed by `intent_type`), so it survives entry reloads; handlers
+  resolve the coordinator per-call (never capture it). Like a websocket command, an
+  intent is an **additional UI surface that delegates to the same `HomeKeeperStore`
+  method** (e.g. `complete_task` with `origin=ORIGIN_INTENT`) — never a divergent
+  mutation path, so completions stay observable and obey the same guards (a
+  problem-sensor-synced task is still rejected). No new service is required when an
+  existing one already covers the action.
+- **Keep intent resolution pure and testable.** Fuzzy task-name matching and
+  due-task selection live in `task_match.py` (no HA imports, loaded in the `hk`
+  synthetic package) so the branchy logic is unit-tested in isolation; the handler
+  only does HA-specific work (registry name→id lookups, spoken responses, error
+  mapping). Prefer **disambiguation over guessing** — when more than one task could
+  match, return candidates and ask, because completing the wrong task advances its
+  recurrence.
+- **Ship the sentences and test the NLU layer.** Natural-language phrases live in
+  `custom_sentences/<lang>/home_keeper.yaml` (users copy them into their config). The
+  sentence→intent→slot routing is covered by a `hassil` unit test
+  (`tests/unit/test_intent_sentences.py`) — the same engine HA uses — and the
+  handler→store wiring by an integration test driving `/api/intent/handle` (the test
+  config enables `intent:`).
+
 ## Events are the observation surface — fire one for every state change
 - **Every observable state change fires a documented `home_keeper_<noun>_<verb>` bus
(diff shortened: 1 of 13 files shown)


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. This PR adds two Assist conversation intents (`intents.py`) backed by a pure matcher (`task_match.py`). Focus on:

- **Correctness / edge cases**: the fuzzy-match scoring and the disambiguation logic in `task_match.match_task` — threshold (0.6) and dominance-margin (0.12) tuning, the containment floor (0.85), exact-match tie handling, and whether any input could cause a *false-positive* completion of the wrong task (which advances recurrence — destructive). Also `select_due_tasks` timezone/`fromisoformat` handling and the `next_due is None` (dormant triggered) case.
- **HA best practices**: `intent.IntentHandler` usage (`slot_schema`, `async_validate_slots`, response building), `intent.async_register` idempotency across entry reloads, resolving the coordinator per-call, and area/device registry lookups (aliases, name collisions, `name_by_user`).
- **Maintainability**: the pure-vs-HA split (`task_match.py` vs `intents.py`), naming, and duplication with existing service handlers.
- **Security**: any user-utterance content that flows into responses or store calls unescaped/unvalidated.
- **Performance**: `match_task` rescans all tasks per call — fine at expected scale, but flag if you see a hot path.

Triage note for reviewers: multi-turn disambiguation, localization of responses, and per-language threshold tuning are known/deferred gaps documented in the PR body.

---

prestomation:
Closed, because the branch is too old to rebase.

Main has moved 3 minor releases since June. The branch now conflicts in 36 files, which include `__init__.py`, `const.py`, `options.py`, `coordinator.py` and every locale file. A rewrite against current main costs less than a rebase.

The design is the part that keeps its value, and this PR body holds it: 2 intents, a pure `task_match.py` matcher, delegation to `store.complete_task` with `origin=ORIGIN_INTENT`, and a refusal to guess between 2 tasks that match the same words. Read this body first if you build voice control again.

---

PR #62, opened 2026-06-21
Fix recurrence-engine crash and triggered-task conversion bug

prestomation:
## Exploratory testing — bugs found & fixed

I went bug-hunting through the pure-logic core (recurrence, models, store, assets, reconcile, transitions, inventory, card-filter) and the frontend utilities. Two concrete backend bugs surfaced; both are fixed here with regression tests. The full unit suite is green (276 passed).

### Bug 1 — fixed `MONTHLY` schedule crashes when anchored far in the past (correctness, crash)

`next_fixed_occurrence` raised `RuntimeError: next_fixed_occurrence exceeded iteration cap` for a `fixed` task with `freq=MONTHLY`, `interval=1`, whose anchor was more than `MAX_EXPAND_ITERATIONS` (500) months — **~41.6 years** — before `now`. `_fast_forward` handled `DAILY`/`WEEKLY` with an O(1) jump but fell through to `return anchor` for `MONTHLY`, leaving the bounded loop to step month-by-month into the cap. (The docstring claimed it'd take "~830 years" to approach the cap — that was wrong; the real figure is ~41.6 years.)

Reproduction (pre-fix):

```python
next_fixed_occurrence(datetime(1976,6,21,tzinfo=utc), "MONTHLY", 1, after=datetime(2026,6,21,tzinfo=utc))
# RuntimeError: ... exceeded iteration cap
```

This crash propagates through `compute_next_due` / `apply_completion`, so creating or completing such a task — or any refresh recomputing it — takes down its *next due* sensor, *overdue* binary sensor, and the calendar.

**Fix.** `_fast_forward` now jumps `MONTHLY` occurrences. The clamped occurrence day is monotonically non-increasing toward 28 (every month has ≥28 days), so once it bottoms out at 28 it is permanently stable and the remainder can be jumped in O(1). We step until the day reaches 28, then bulk-jump; schedules whose reachable months never clamp to 28 (e.g. an even interval that skips February) just step to the target, bounded by the finite anchor→target span — never the old cap crash.

I verified the new path is **byte-identical to single-stepping** across **665,760 exhaustive cases** (every anchor day 1–31, every anchor month, 19 intervals, year offsets spanning leap-year and century edges). The caller's loop runs ≤2 iterations afterward.

### Bug 2 — converting a `triggered` task to `floating`/`fixed` errors (correctness)

`merge_update`'s candidate always carries an `interval` key. A `triggered` task has no interval, so the candidate gets `interval=None`, and `normalize_fields`' `int(data.get("interval", 1))` returned that `None` (key present) rather than the default → `TaskValidationError: interval must be a valid integer`. Converting a monitored task to a recurring one with just a unit failed.

**Fix.** Coalesce an absent/`None`/`""` interval to `1` (matching fresh creation) before validation.

### Notes / observations (not changed here)

- A `fixed` task that is never completed keeps its stored `next_due` pinned to the first occurrence (perpetually overdue), while the **calendar** entity shows future occurrences — the coordinator never recomputes `next_due` over time. Self-consistent but arguably an inconsistency between surfaces; left as-is pending a product call.
- Completing a wear-part task with a future `completed_at` would stamp a future `last_replaced`, which a later asset edit rejects ("last_replaced must not be in the future"). Edge case, not addressed.

### Testing

- `pytest tests/unit -v` → 276 passed (5 new).
- `ruff check` + `ruff format --check` clean on the changed files.
- Brute-force oracle comparison for the monthly fast-forward (665k cases).

---
_Generated by [Claude Code](https://claude.ai/code/session_014spKqiUDGg8KHMA7RXw3x8)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,17 +38,6 @@ versioning (with PEP 440 pre-release suffixes — `bN`/`aN`/`rcN` — for betas)
 
 ### Fixed
 
-- **Fixed monthly schedules anchored far in the past no longer crash.** A `fixed`
-  task with a `MONTHLY` frequency whose anchor was more than ~41 years before now
-  (with a 1-month interval) could blow the recurrence engine's iteration cap and
-  raise, taking down the task's *next due* sensor, *overdue* binary sensor, and the
-  calendar. The monthly fast-forward now jumps to the right occurrence directly,
-  matching the day-by-day clamping behaviour exactly (verified against an
-  exhaustive reference) including end-of-month and leap-year cases.
-- **Converting a monitored (triggered) task to a recurring one no longer errors.**
-  Changing a `triggered` task to `floating`/`fixed` without re-sending an interval
-  raised "interval must be a valid integer"; the interval now defaults to 1 (as on
-  a fresh task) when not supplied.
 - **Settings exclusions now take effect immediately.** Adding a problem-sensor
   entity, area, or label to a *skip* list in the panel's **Settings** tab is now
   reflected right away: the integration reload that re-runs the problem-sensor sync
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,17 @@ versioning (with PEP 440 pre-release suffixes — `bN`/`aN`/`rcN` — for betas)
 
 ### Fixed
 
+- **Fixed monthly schedules anchored far in the past no longer crash.** A `fixed`
+  task with a `MONTHLY` frequency whose anchor was more than ~41 years before now
+  (with a 1-month interval) could blow the recurrence engine's iteration cap and
+  raise, taking down the task's *next due* sensor, *overdue* binary sensor, and the
+  calendar. The monthly fast-forward now jumps to the right occurrence directly,
+  matching the day-by-day clamping behaviour exactly (verified against an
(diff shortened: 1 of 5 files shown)


prestomation:
/q review Please be critical and skeptical, surfacing the most serious issues first and not withholding minor ones. This PR changes the recurrence engine's `_fast_forward` MONTHLY branch and a `normalize_fields` interval default. Focus on:

- **Correctness of the monthly fast-forward** (`recurrence._fast_forward`): is the O(1) jump truly equivalent to single-stepping for *all* anchors/intervals, including end-of-month day clamping (Jan 31 → Feb 28 stickiness), leap years, century non-leap years (1900/2100), and the `interval` that never reaches a 28-day February (does it terminate and undershoot correctly)? Any off-by-one in the `remaining // interval - 1` undershoot, or cases where it could *overshoot* past the target and skip an occurrence?
- **Performance**: the non-clamping-to-28 fallback steps to the target — is the span always bounded, and is the worst case acceptable?
- **The interval default change** in `models.normalize_fields`: does coalescing `None`/`""` → 1 mask any validation it should reject, and is it consistent across `build_task` and `merge_update` for floating vs fixed vs triggered?
- Any **DST** concerns introduced by the monthly jump.
- HA best practices and maintainability of the changes.

---

PR #74, opened 2026-06-22
fix(card): fire ll-custom-cards-update so card appears in picker on first load

prestomation:
## Summary

- `card-index.ts` now dispatches `ll-custom-cards-update` after pushing to `window.customCards`
- Updated the e2e helper comment to reflect the fix

## Root cause

The Lovelace card picker reads `window.customCards` once when it initialises. When the card JS loads as an extra module (via `frontend.add_extra_js_url`), it may arrive *after* the dashboard has already rendered — especially on the first cold load. Because no notification was sent, the picker never re-read the registry, so the card was invisible until the user reloaded the page (at which point the module was cached and executed before the picker).

`window.dispatchEvent(new CustomEvent('ll-custom-cards-update'))` is the standard HA hook that tells the Lovelace card picker to re-read `window.customCards` immediately, so the card appears without a reload.

## Test plan

- [ ] Install the integration, open a dashboard → "Add card" picker — **Home Keeper Tasks** should appear on the first load without a manual page reload
- [ ] Manually add `type: custom:home-keeper-card` in YAML — card renders correctly
- [ ] E2e tests pass (`bash ci/e2e-up.sh`)

https://claude.ai/code/session_01BwrMtzD4KwFdShBVegGDSK

---
_Generated by [Claude Code](https://claude.ai/code/session_01BwrMtzD4KwFdShBVegGDSK)_

### DIFF
--- a/custom_components/home_keeper/frontend/src/card-index.ts
+++ b/custom_components/home_keeper/frontend/src/card-index.ts
@@ -25,6 +25,9 @@ if (!w.customCards.some((c) => c.type === 'home-keeper-card')) {
     preview: true,
     documentationURL: 'https://github.com/prestomation/ha-home-keeper',
   });
+  // Notify Lovelace to re-read window.customCards — needed when this module
+  // loads after the dashboard has already initialised (first cold load).
+  window.dispatchEvent(new CustomEvent('ll-custom-cards-update'));
 }
 
 export { HomeKeeperCard, HomeKeeperCardEditor };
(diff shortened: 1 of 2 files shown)


prestomation:
/q review Please give critical, skeptical feedback on this change. Focus on: **correctness** (is `ll-custom-cards-update` the right event name and target — could it be case-sensitive or need a different dispatch target like `document`?), **completeness** (does firing the event once on first registration cover all edge cases — entry reload, multiple config entries, HA restart?), **HA best practices** (is `frontend.add_extra_js_url` still the recommended mechanism in current HA, or is there a newer API?), and **test coverage** (is the e2e retry logic still needed, or can it be tightened now that the event fires?). Surface the most serious issues first.

---

PR #87, opened 2026-06-25
Offline manuals: per-asset documents (links + uploaded files)

prestomation:
Closes #79.

## What & why

The issue asked for the option to **store a manual locally** instead of only a URL. This replaces the single appliance `manual_url` with a richer **documents** list: each appliance can hold any number of manuals/warranties/receipts, each either an external **link** or an **uploaded file** (PDF or image) stored on the HA instance — so the manual is there even when the manufacturer's site isn't.

Existing `manual_url` links are **migrated automatically** into a document on load (no version bump — same additive convention as the legacy parts migration).

## How it works

- **Data model** (`assets.py`) — `documents` list of `{id, kind: link|file, name, created[, url | filename, content_type, size]}`, with `_normalize_documents` and `migrate_documents_from_manual_url`.
- **Storage & serving** — binaries can't ride a YAML service or the websocket, so files go through the integration's single auth-gated `HomeAssistantView` (`manuals.py`): `POST` multipart → validated → written under `config/home_keeper/documents/<asset_id>/…` (one dir per asset → asset-delete is one `rmtree`); `GET` streams it. The browser opens a file via a short-lived `async_sign_path` signed URL minted by the `home_keeper/sign_document_url` websocket command. The security-critical, HA-free checks (magic-byte sniff, allowlist PDF + PNG/JPEG/WebP/GIF, 25 MB cap, filename sanitization, path-traversal guard) live in `documents.py` so they're unit-testable without an HA runtime.
- **Service-first** — the metadata still funnels through `store.add_asset_document` / `remove_asset_document` (so `home_keeper_asset_updated` fires with `changed_fields: ["documents"]`), exposed as `home_keeper.add_asset_document` / `home_keeper.remove_asset_document` services (+ matching websocket commands). File removal deletes the on-disk blob.
- **Panel** — a "Manuals & documents" editor (add link / **Upload file** / remove) on the appliance edit form, and a documents section on the detail page that opens links directly and files via the signed URL.

## Screenshot

The appliance Manuals & documents editor — an existing (migrated) manual link with a remove button, plus add-link and upload-file controls

## Tests (run locally)

- **Unit** — 381 pass, incl. new `test_documents.py` (sniffing, allowlist, size cap, filename sanitization, traversal guard) and `documents`/migration cases in `test_assets.py`.
- **Frontend** — 184 vitest pass (i18n parity + untranslated-leak guards updated for the new keys across all 17 locales); `tsc` clean; panel builds.
[...]

Issue #79, opened 2026-06-23
Offline manuals


mdarocha:
Hi,

currently it looks like the manuals can only be an URL. Is it possible to add an option to upload a PDF to store locally?

prestomation:
Thank you for trying out home keeper and for your request. I have added support for file upload support for manuals in version 0.5.0. please let me know if you have any more requests

PR #88, opened 2026-06-25
Fix: open uploaded documents on the current host, not HA's LAN URL

prestomation:
Follow-up bug fix to the offline-manuals feature (#79 / #87), reported in testing.

## Symptom
A user uploaded a document on their LAN, then — **browsing Home Assistant through a remote/external address** — clicked the file and it opened the **LAN URL** instead, which is unreachable from outside.

## Cause
`home_keeper/sign_document_url` (via HA's `async_sign_path`) returns a **host-less, signed path** like `/api/home_keeper/document/<asset>/<doc>?authSig=…`. The panel opened it with `window.open(path)`, and the browser resolves a relative URL against the **document base URL** — which can resolve to Home Assistant's internal/LAN address. So the file opened on the wrong host when you weren't on that network.

## Fix
Resolve the signed path against **`window.location.origin`** — the address the panel is actually loaded from — before opening it:

```ts
const path = await api.signDocumentUrl(this._hass, assetId, documentId);
const url = new URL(path, window.location.origin).href; // current host, base-immune
window.open(url, '_blank', 'noopener');
```

One method changed (`_openFileDocument`), 7 lines. The signed path and its auth are unchanged; only the host it opens on is corrected.

## Verification
e2e: with a deliberately misleading `<base href="http://10.99.99.99:1234/">` injected, `new URL(path, window.location.origin)` still resolves to the current host (`localhost:8123`, **not** the fake host), and the real click flow opens the file on the current origin with the signed query intact. `tsc` + 184 vitest pass.

No visible-surface change (the documents editor/detail look identical; only the opened URL's host differs), so no new screenshot.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ag7g5tNZbgJSVdx8CEwxJt)_

### DIFF
--- a/custom_components/home_keeper/frontend/src/panel.ts
+++ b/custom_components/home_keeper/frontend/src/panel.ts
@@ -2244,7 +2244,13 @@ export class HomeKeeperPanel extends HTMLElement {
   private async _openFileDocument(assetId: string, documentId: string): Promise<void> {
     if (!this._hass) return;
     try {
-      const url = await api.signDocumentUrl(this._hass, assetId, documentId);
+      const path = await api.signDocumentUrl(this._hass, assetId, documentId);
+      // `signDocumentUrl` returns a host-less, signed path (e.g. `/api/.../?authSig=`).
+      // Resolve it against the origin the panel is *actually* loaded from rather than
+      // letting the browser resolve the relative path against the document base — the
+      // latter can yield Home Assistant's internal/LAN URL when you're browsing through
+      // a remote/external address, opening the file on the wrong (unreachable) host.
+      const url = new URL(path, window.location.origin).href;
       window.open(url, '_blank', 'noopener');
     } catch {
       // Best-effort: a failed signing (e.g. deleted file) simply doesn't open.


prestomation:
/q review Please be critical and skeptical, most serious issues first. This one-method fix changes how an uploaded document is opened: instead of `window.open(signedRelativePath)` it now does `window.open(new URL(signedRelativePath, window.location.origin).href)` so the file opens on the host the panel is loaded from rather than resolving against the document base (which could be HA's internal/LAN URL when browsing remotely). Scrutinize:

- **Correctness across access modes**: does `new URL(path, window.location.origin)` behave correctly for Nabu Casa Remote UI, a custom external domain via reverse proxy, the HA Companion app webview, and a sub-path-hosted HA (`/lovelace`-style base path)? Could pinning to `window.location.origin` ever be *wrong* (e.g. when the panel legitimately should use a different host)?
- **Signed-path integrity**: the signature covers the path + query; does wrapping in `new URL(...)` preserve the `authSig` query and path exactly (no encoding changes that would break signature validation)?
- **Edge cases**: what if `signDocumentUrl` ever returned an absolute URL — does `new URL(path, origin)` still do the right thing? Any regression risk for the previously-working local-network case?
- Anything simpler/more idiomatic (e.g. an existing HA frontend helper like `hassUrl`) I should use instead.

---

PR #89, opened 2026-06-25
Clean up the appliance documents editor: edit, open, file details

prestomation:
## Why

Follow-up to the offline-manuals feature (#87). The documents editor only let you
**add** documents — there was no visible way to **edit** or **remove** an existing one
(the per-document delete button rendered no glyph), and an existing document looked
confusingly like the add-link form's name box. Reported by @prestomation: *"looks like
there's no way to edit or delete manuals… hard to tell the difference between that item
and the name box which is for the addition UI."*

## What changed

Each existing document is now a clear **card** (icon + name + details) with visible
**Open / Edit / Remove** actions, separated from a labelled **"Add a document"** area:

- **Edit existing documents** inline — rename any document and change a link's URL.
  Files are upload-only, so they're **rename-only** (blob/filename/type are immutable).
  New `home_keeper.update_asset_document` **service** + `update_asset_document` websocket
  command, `store.update_asset_document`, and pure `assets.update_document` helper.
- **Open/preview** a document from the editor — a link opens its URL; a file opens via a
  short-lived signed URL.
- **Add link documents while first creating** an appliance, before it's saved (they ride
  along in the create payload; the backend already seeds links on create).
- **Show file details** on the card: filename, human size, and type.

**Plain-HTTP LAN fix.** `crypto.randomUUID()` only exists in a secure context (HTTPS or
localhost), so on a plain-HTTP LAN address (e.g. `http://192.168.1.x:8123`) it was
`undefined` and minting a client-side document id threw — silently breaking file uploads
and adding a link to a not-yet-saved appliance. Added a `randomId()` helper that prefers
`crypto.randomUUID()` and falls back to a `crypto.getRandomValues` v4 uuid.

## Editor

The appliance Manuals & documents editor — an existing document as a card with name, URL, and Open / Edit / Remove actions, separated from a clearly labelled Add a document area with name/URL fields and Add link / Upload file controls

## Conventions / parity

- Service registered with `services.yaml` + `strings.json` + **16-locale** translation
  parity (backend parity test green).
- Frontend locale keys added (and the now-unused `doc.link`/`doc.file` text labels
  retired, since the card uses icons); i18n cognate allowlist updated.
- Architecture rule, README, and CHANGELOG updated.

## Tests

- **Unit:** `assets.update_document` (link rename + URL change, file rename keeps blob
  fields, bad-URL / non-dict / unknown-id paths); `randomId()` secure-context fallbacks.
  Full unit suite **387 passed**; frontend vitest **187 passed**.
- **Integration (Docker):** `update_asset_document` round-trips — link rename+URL, file
  rename keeps the downloadable blob, unknown-id errors. `test_documents.py`: **10 passed**.
- **Frontend:** tsc clean, panel builds.
- Refreshed `docs/images/32-panel-appliance-documents.png` via the Playwright capture.

https://claude.ai/code/session_01Ag7g5tNZbgJSVdx8CEwxJt


prestomation:
/q review Please be critical and skeptical, surfacing the most serious issues first and not withholding minor ones. Focus on:

- **Correctness**: the new `update_asset_document` path end-to-end (`assets.update_document` → `store.update_asset_document` → service + websocket). Edge cases: editing a `file` document must stay rename-only (no URL/blob mutation), id/kind/created must be preserved, unknown asset/document ids must error cleanly, and a non-dict `changes` must be rejected. In the panel, adding link documents to a *new* (unsaved) appliance and having them ride the create payload — any way a file entry leaks into that payload, or a working-copy edit/remove mutates the wrong entry?
- **HA best practices**: service schema/`services.yaml`/`strings.json`/translation parity for the new service; the documents mutation deliberately skips `devices.async_apply_asset_change` — confirm that's correct (documents touch no device/entity/task).
- **Maintainability**: the `_renderDocumentsEditor` split into card/edit/add helpers, the per-doc `editingDocId` edit state, and duplication between the add-link and edit forms.
- **Security**: the URL re-validation on edit (link), filename/type immutability for files, and escaping of user-controlled document name/URL/filename in the card subtitle.
- **Performance**: full `_render()` on every doc add/edit/remove — acceptable, or worth narrowing?

Thanks!

---

prestomation:
/q review Please be critical and skeptical. New commit adds a `randomId()` helper (`utils.ts`) to fix `crypto.randomUUID()` being undefined on a **plain-HTTP LAN** origin (non-secure context), used for the file-upload document id and the working-copy link id.

- **Correctness**: is the v4 fallback well-formed (version/variant bits) and collision-safe enough for client-minted document ids? Any path where `globalThis.crypto` is present but partial (e.g. `getRandomValues` only) that the helper mishandles? The `Math.random` last-resort branch — acceptable for ids that key on-disk blobs?
- **Security**: `Math.random` is not cryptographically strong — does it matter here, given the id only namespaces a per-asset document and the blob is auth-gated + signed?
- **Maintainability**: helper placement in `utils.ts`, and whether other `crypto.randomUUID()` callers exist that should use it too.

Thanks!

---

PR #106, opened 2026-06-29
fix: back navigation from cross-view detail + stale entity cleanup (#104, #105)

### LINKED ISSUE #104: [BUG] Problem Sensor Sync leaves stale entities after disabling or exclusions
Hey!

I hope you don't mind me sharing a few more findings. Please feel free to close any of my issues as **"out of scope"** without any further explanation if they don't fit the project's direction.

### Environment
Home Assistant: 2026.6.4
Home Keeper Integration: v0.6.0b4
Browser: Firefox 152.0.3

### Issue Description:
Enabling and then disabling **Problem Sensor Sync** leaves stale entities behind on affected devices:

* `sensor.*_next_due`
* `binary_sensor.*_overdue`

### Steps to reproduce:

1. Enable **Problem Sensor Sync**.
2. Then do one of the following:

   * Disable **Problem Sensor Sync**.
   * Exclude a device.
   * Exclude a label.
   * Exclude a sensor.

### Expected behavior:
The stale entities should be removed when **Problem Sensor Sync** is disabled or when a sensor, device, or label is excluded.

### LINKED ISSUE #105: [BUG] Back navigation jumps to Home instead of previous page
### Environment
Home Assistant: 2026.6.4
Home Keeper Integration: v0.6.0b4
Browser: Firefox 152.0.3

### Issue Description:
When navigating in Appliance → %device% → related Task, clicking Back does not return to the previous page. Instead, it redirects to the Home “Task” page rather than back to the selected device view.

### Steps to reproduce:
1. Open the Home Keeper sidebar panel
2. Go to Appliance
3. Select a configured device
4. Open a related Task
5. Click Back

You are redirected to the Home “Task” page instead of returning to the previously selected Appliance/device view.

### Expected behavior:
Clicking Back should navigate to the previous page in history (the selected Appliance/device view)

prestomation:
Closes #104, closes #105

## Issue #105 — Back button jumps to task list instead of previous appliance view

**Root cause:** `_openDetail()` always set `view` to the target kind's home tab (`'tasks'` for a task), overwriting `this._view`. So when a task was opened from inside an appliance detail, `_closeDetail()` navigated to the task list instead of returning to the appliance.

**Fix (`frontend/src/panel.ts`):**
- Added a `_hasHistory` flag, set to `true` on every `pushState` call in `_navigate()`.
- `_closeDetail()` now calls `history.back()` when history exists — the browser's own push-state stack is always the correct return destination, including cross-view drills (appliance detail → related task → Back → appliance detail).
- Falls back to the old explicit-navigate for users who deep-linked directly to a detail URL (no panel push-state has occurred).

**e2e tests added (`smoke.spec.ts`):**
- `"Back from a task opened inside an appliance detail returns to the appliance detail"` — navigates to the water heater, opens a related task, asserts Back returns to the appliance URL.
- `"the in-panel Back button from an appliance detail returns to the appliances list"` — covers the simple appliance detail → Back → list path.
- `"an appliance detail URL deep-links straight to the appliance detail page"` — verifies the cold-link fallback: deep-linked appliance detail, Back returns to the list.

---

## Issue #104 — Problem Sensor Sync leaves stale entities after disabling or exclusions

**Root cause:** When a device-attached task is deleted (manually or via Problem Sensor Sync being disabled/re-scoped), the config entry reloads and `async_setup_entry` re-runs. It only *adds* entities for the current live task set — it never removes orphaned entity-registry entries. Those orphaned entries resurface as unavailable ghosts on the device page after an HA restart.

**Fix 1 — per-task entities (`sensor.py`, `binary_sensor.py`, `button.py`):**  
At the start of each platform's `async_setup_entry`, scan the config entry's entity-registry slice and remove any entry whose unique-id parses to a task that is no longer in `coordinator.device_attached_task_ids()`. Suffixes are stable and unique per platform (`_next_due`, `_overdue`, `_done`), so the extraction is unambiguous.

**Fix 2 — asset date sensors (`sensor.py`):**  
Extends the same registry cleanup pass to also remove `home_keeper_asset_*_meta_*` sensors whose tracked-date metadata entry has been removed or un-tracked via `update_asset`. The two cleanup checks are now combined into a single registry pass for efficiency.

**Integration tests (`test_entity_cleanup.py`):**
- `test_stale_per_task_entities_removed_after_task_deleted` — adds a probe task with a fake `device_id`, verifies all three registry entries (`_done`, `_next_due`, `_overdue`) exist, deletes the task, then polls the entity registry (via WebSocket) to confirm all three entries are gone.
- `test_stale_asset_date_sensor_removed_after_metadata_untracked` — creates an asset with a tracked date field, verifies the `home_keeper_asset_*_meta_*` sensor appears in the entity registry, removes the metadata via `update_asset`, and asserts the stale registry entry is cleaned up.

---

## Tests run locally
- `pytest tests/unit -v` — 400 passed
- `npm test` (vitest) — 188 passed
- `npx tsc --noEmit` — clean
- All 37 e2e Playwright tests pass (up from 35 before this PR)

---
_Generated by [Claude Code](https://claude.ai/code/session_01MPA6K8fJGPesq2CADKsNqh)_

### DIFF
--- a/custom_components/home_keeper/binary_sensor.py
+++ b/custom_components/home_keeper/binary_sensor.py
@@ -8,6 +8,7 @@ from homeassistant.components.binary_sensor import (
 )
 from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant
+from homeassistant.helpers import entity_registry as er
 from homeassistant.helpers.entity_platform import AddEntitiesCallback
 from homeassistant.util import dt as dt_util
 
@@ -25,6 +26,24 @@ async def async_setup_entry(
 ) -> None:
     """Create an overdue binary sensor for each device-attached task."""
     coordinator: HomeKeeperCoordinator = entry.runtime_data
+
+    # Remove entity-registry entries for per-task binary sensors whose task no
+    # longer exists (e.g. after disabling Problem Sensor Sync or deleting a task).
+    live_ids = set(coordinator.device_attached_task_ids())
+    reg = er.async_get(hass)
+    _prefix = f"{DOMAIN}_"
+    _suffix = "_overdue"
+    for entity_entry in list(reg.entities.get_entries_for_config_entry_id(entry.entry_id)):
+        uid = entity_entry.unique_id or ""
+        if (
+            entity_entry.entity_id.split(".", 1)[0] == "binary_sensor"
+            and uid.startswith(_prefix)
+            and uid.endswith(_suffix)
+        ):
+            task_id = uid[len(_prefix) : -len(_suffix)]
+            if task_id not in live_ids:
+                reg.async_remove(entity_entry.entity_id)
+
     async_add_entities(
         HomeKeeperOverdueBinarySensor(coordinator, task_id)
         for task_id in coordinator.device_attached_task_ids()
--- a/custom_components/home_keeper/binary_sensor.py
+++ b/custom_components/home_keeper/binary_sensor.py
@@ -46,8 +46,7 @@ async def async_setup_entry(
                 reg.async_remove(entity_entry.entity_id)
 
     async_add_entities(
-        HomeKeeperOverdueBinarySensor(coordinator, task_id)
-        for task_id in task_ids
+        HomeKeeperOverdueBinarySensor(coordinator, task_id) for task_id in task_ids
     )
 
 
--- a/custom_components/home_keeper/binary_sensor.py
+++ b/custom_components/home_keeper/binary_sensor.py
@@ -29,24 +29,25 @@ async def async_setup_entry(
 
     # Remove entity-registry entries for per-task binary sensors whose task no
     # longer exists (e.g. after disabling Problem Sensor Sync or deleting a task).
-    live_ids = set(coordinator.device_attached_task_ids())
+    task_ids = coordinator.device_attached_task_ids()
+    live_ids = set(task_ids)
     reg = er.async_get(hass)
-    _prefix = f"{DOMAIN}_"
-    _suffix = "_overdue"
-    for entity_entry in list(reg.entities.get_entries_for_config_entry_id(entry.entry_id)):
+    prefix = f"{DOMAIN}_"
+    suffix = "_overdue"
+    for entity_entry in reg.entities.get_entries_for_config_entry_id(entry.entry_id):
         uid = entity_entry.unique_id or ""
         if (
-            entity_entry.entity_id.split(".", 1)[0] == "binary_sensor"
-            and uid.startswith(_prefix)
-            and uid.endswith(_suffix)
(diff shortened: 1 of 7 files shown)


prestomation:
/q review Please give critical, skeptical feedback on this PR. Key areas to scrutinize:

**Correctness:**
- `_closeDetail()` now calls `history.back()` — are there edge cases where this could navigate outside the panel (e.g. if the panel is the first page in a fresh tab)?
- The `_hasHistory` flag is per-instance and resets on page load — is that the right scope?
- Entity registry cleanup in `async_setup_entry`: the suffix-based task_id extraction (`uid[len(prefix):-len(suffix)]`) — could this incorrectly match non-task-entity unique-ids (e.g. `HomeKeeperAssetDateSensor` whose uid is `home_keeper_asset_{id}_meta_{entry_id}`)?
- Does `reg.entities.get_entries_for_config_entry_id` exist in the HA version targeted by this integration?

**HA best practices:**
- Is calling `er.async_get(hass)` + `reg.async_remove()` inside `async_setup_entry` the recommended HA pattern for orphan cleanup, or is there a more idiomatic approach?
- Any concern about removing registry entries while the platform is mid-setup?

**Testing:**
- The integration test uses `GET /api/config/entity_registry` — confirm this endpoint is available in HA 2026.x and returns unique_id fields.
- Is the 20-second polling timeout in the integration test sufficient given entry-reload latency?

Surface the most serious issues first and don't withhold minor ones.

---

prestomation:
/q review Please give critical, skeptical feedback on this PR. Key areas to scrutinize:

**Correctness:**
- `sensor.py` cleanup pass: does the single-pass registry scan correctly distinguish `home_keeper_asset_*_meta_*` UIDs from `home_keeper_*_next_due` task UIDs? Could a task ID that starts with `asset_` and contains `_meta_` be misidentified?
- The asset date sensor cleanup builds `live_asset_meta_uids` by iterating assets and calling `coordinator.device_info_for_device_id()` — same condition used to skip sensor creation. Is there a race or off-by-one where we clean up a UID we just decided to create?
- `test_entity_cleanup.py`: does polling `_unique_ids_in_registry` via WebSocket correctly reflect HA's registry state post-reload, or could there be a window where the reload hasn't settled yet?

**HA best practices:**
- Calling `reg.async_remove()` in `async_setup_entry` — is this safe to do synchronously in a coroutine? Are there ordering concerns with `async_add_entities` called immediately after?
- Should the registry mutations be guarded against the entity already being removed (e.g., if two platforms race on teardown)?

**Maintainability:**
- The split between `_close_detail` using `history.back()` vs `_navigate(..., replace=True)` is controlled by a `_hasHistory` flag. Is this robust across all navigation paths (tab switches use `replace=True` so they don't pollute the stack — is that correctly handled)?

---

PR #110, opened 2026-06-29
Enrich appliance device pages (serial, spare-stock entities, deep link)

prestomation:
## What & why

A virtual appliance's Home Assistant **device page** is framework-fixed (an integration can't inject cards/HTML onto it) — but we *can* control its entities, device-info metadata, the "Visit" link, and diagnostics. This PR leans into all of those so a Home Keeper appliance's device page becomes a real maintenance summary **and** a one-click jump into the panel for the data that has no native rendering (manuals, full inventory, history). It also makes the reverse jump work: from the panel into the device page.

Design discussion and rationale are captured in `docs/DESIGN.md` → "Device page enrichment".

## Changes

**Owned (virtual appliance) device pages get:**
- **Serial number in the device-info block** — `serial_number` is now a first-class asset field (`assets._TEXT_FIELDS`) synced into `DeviceInfo.serial_number` by `devices._reconcile_virtual` (guarded by `_supports_kwarg` for older cores), editable in the panel's appliance form next to make/model and on the `add_asset`/`update_asset` services. It's also a first-class column on the `export_inventory` report/CSV (insurance record).
- **A precise "Visit" deep link** — `configuration_url` now points at `homeassistant://navigate/home-keeper/appliances/<asset_id>` (was the panel root), landing on that appliance's detail page where its documents and inventory render.
- **Per-part spare-stock entities** — an editable stock `number` (`number.py`, delegates to `store.adjust_part_stock`) for each stock-tracked part, and a low-stock `PROBLEM` `binary_sensor` for each part with a reorder threshold. Both enumerate via the shared `coordinator.virtual_asset_parts(predicate)` helper and prune stale registry entries.
- **Per-device diagnostics** — `async_get_device_diagnostics` scopes the download to that appliance's tasks + parts.

**Panel → device page (follow-up):**
- The **"Virtual device"** chip on the Appliances list and appliance detail is now a clickable link to that appliance's HA device page (it was a static marker; virtual appliances provision a real registry device now). Its stale "not a real device" tooltip is corrected across all locales.
- **Fixed:** the device chip on a task/appliance *card row* never opened the device page — the click bubbled to the row's open-detail handler. The chip click now stops propagation, so task-row device chips navigate too.

**Owner-vs-guest:** enrichment applies only to Home Keeper-owned (virtual) devices. A task attached to a *foreign* device still gets its per-task entities there, but we never overwrite that integration's device-info block or Visit link.

**Deliberately skipped:** a device-level "mark done / restock" *action* — ambiguous on a device with several tasks/parts; the per-task mark-done button + `adjust_part_stock` service already cover it.

## Screenshots

A virtual appliance ("Garage water heater") device page — serial in the info block, the Visit deep link, per-part spare-stock numbers, a low-stock problem sensor (anode rod: 1 on hand, reorder-at 2), plus the existing per-task and tracked-date entities:

Appliance device page with serial, spare-stock numbers, low-stock problem sensor, and Visit deep link

Device-info block close-up (make / model / serial number + the Visit link into the panel):

Device-info block showing model, manufacturer, serial number and the Visit deep link

The panel's Appliances list — each virtual appliance now has a clickable "Virtual device" chip (open-in-new icon) linking to its HA device page:

Appliances list with clickable Virtual device chips

Appliance detail header — the same clickable chip next to the title:

Appliance detail with clickable Virtual device chip

Captured with `tests/e2e/screenshots-device-page.capture.ts` and `tests/e2e/screenshots-device-link.capture.ts` (the latter also asserts both the appliance chip and a task-row device chip navigate to `/config/devices/device/...`).

## Tests

- Pure unit suite: **420 passed** (added coverage for `part_tracks_stock`, `part_has_reorder`, and first-class `serial_number` normalization).
- Frontend vitest: **188 passed**; panel builds clean.
- Full Docker **integration suite: 65 passed** (fixed the one test that asserted serial via the old metadata path).
- Full backend translation parity green; locale parity + untranslated-leak checks green for the new/changed strings.
- Backend ruff lint + format clean.
- End-to-end verified in a real HA container (screenshots are the rendered surfaces; the panel→device navigation is asserted, not assumed).

## Docs

`README.md`, `docs/DESIGN.md`, `.amazonq/rules/architecture-and-code.md`, and `CHANGELOG.md` (folded into the unreleased `0.7.0b2`, with a `### Fixed` for the chip-click bug).

---
_Generated by [Claude Code](https://claude.ai/code/session_01TMebVwdRCUYzBsHQpY9zzf)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first and don't withhold minor ones. Focus on:

- **Correctness**: the per-part stock `number` (`number.py`) computes `delta = int(value) - current` and calls `adjust_part_stock` — are there edge cases (concurrent edits, float/int coercion, value below 0 or above MAX_INTERVAL, a part whose stock is None when set) that misbehave? Does `async_set_native_value` race with the coordinator refresh? Is the low-stock `binary_sensor` state (`part_is_low`) consistent with the edge-triggered `part_*` events?
- **Entity lifecycle**: stale-entity pruning in `number.py` / `binary_sensor.py` keys on unique-id shape (`_asset_<id>_part_<id>_stock` / `_low_stock`) — can a real asset/part id collide with that shape or with the per-task `_overdue` suffix and wrongly remove a live entity? Are entities correctly created/removed across asset edits (which reload the entry)?
- **Owner-vs-guest**: enrichment is gated to `kind == "virtual"` via `coordinator.virtual_asset_parts`. Is there any path where we'd stomp a foreign device's `configuration_url`/serial, or attach stock entities to a device we don't own?
- **DeviceInfo**: `serial_number` on `async_get_or_create` is guarded by `_supports_kwarg`; is the create-vs-update handling correct on older cores, and does the per-asset `configuration_url` deep link match the panel's route (`appliances/<id>`)?
- **Diagnostics**: `async_get_device_diagnostics` device→tasks/assets association — any over/under-matching (e.g. self-owned task device identifier vs asset identifier)?
- **Maintainability / HA best practices**: naming, duplication between the two platforms, the `virtual_asset_parts` helper, translation parity.

---

prestomation:
/q review Follow-up push `b283363`: serial number is now a first-class column in the `export_inventory` report + CSV (it previously rode in the free-form metadata "Serial" entry, which this PR drops). Please scrutinize, critically:

- **CSV correctness**: inserting the `serial_number` column mid-schema — does anything downstream assume fixed column positions or a specific header set? Is the new column safe against the formula-injection guard (`_cell`) for serials starting with `=/+/-/@`?
- **Data consistency**: is it right that serial is now a dedicated column while warranty/dates/provider stay in the flattened `details`? Any double-reporting risk for users who *also* kept a free-form "Serial" metadata entry from before?
- **Back-compat**: existing appliances created before this PR have no `serial_number` — confirm the export degrades cleanly (empty cell), not a crash.

Surface the most serious issues first; don't withhold minor ones.

---

prestomation:
/q review Follow-up push `70ba31b` — panel→device-page links. Please be critical:

- **Event hijack fix**: the device-chip click now calls `e.stopPropagation()` so it isn't swallowed by the enclosing `.detail-open` card row. Is `stopPropagation` (not `stopImmediatePropagation`) sufficient and correct here? Any other delegated listeners on the row or document that still need to be considered? Does the keyboard path (Enter/Space) also avoid double-firing the row's open-detail?
- **New `_virtualDeviceChip`**: it renders a clickable `.hk-device-chip` only when `asset.device_id` resolves in `hass.devices`, else a static chip. Is the "not yet provisioned / device missing" fallback correct, and does it avoid a dead link to a non-existent device id? Is `_wireDeviceChips`' brand-logo fallback a no-op for this chip (it has no `.hk-dev-img`)?
- **Correctness of the navigation target**: `_navigateToDevice` pushes `/config/devices/device/<id>` — fine for virtual devices we own; any risk for an asset whose `device_id` is stale?
- **i18n**: `chip.virtualDevice.tip` was reworded across all 16 locales (it previously, wrongly, said "not a real device"). Confirm no parity/leak regressions and the translations read sensibly.

Surface the most serious issues first; don't withhold minor ones.

---

prestomation:
/q review Follow-up push `a0fd966` — fixes the appliance device-page "Visit" link going to the dashboard. Root cause: HA renders a `homeassistant://X` `configuration_url` as the in-app path `/X` and does NOT strip a `navigate/` action segment on the web frontend, so `homeassistant://navigate/home-keeper/appliances/<id>` became a dead `/navigate/...` link. Fixed to the bare path `homeassistant://home-keeper/appliances/<id>`. Please be critical:

- Is the bare-path `homeassistant://<panel>/<...>` form correct and stable across HA frontend versions and the **mobile app** (which historically used the `navigate/` deep-link scheme)? Is there a risk this fixes web but regresses the companion app, or vice-versa? If the two clients differ, what's the right single value?
- The new integration test asserts the registry `configuration_url` is exactly `homeassistant://home-keeper/appliances/<asset_id>` and contains no `navigate`. Is asserting the device-registry value (vs the rendered href) a sufficient regression guard, given the bug was in how HA renders it?
- Any edge cases for a virtual asset whose `device_id` isn't yet provisioned, or an existing-device asset (which should get no configuration_url from us)?

---

prestomation:
/q review Push `b8b9066` addresses a code review. Please sanity-check the behavior change, critically:

- `number.py async_set_native_value` now early-returns when `self._part()` is `None` (part/appliance removed between render and submit) instead of falling through to `adjust_part_stock` with `current=0` (which leaked a raw, unlocalized `KeyError`). Is a silent no-op the right call here, or should it raise a localized `HomeAssistantError`? Any case where `_part()` is `None` but we *should* still act?
- It now uses `round(value)` instead of `int(value)` for the incoming spare count. Confirm this can't produce a surprising delta (e.g. with `native_step=1` and box mode) and matches the min/max bounds.
- Version split: bumped to `0.7.0b3` (manifest + `const.PANEL_VERSION`) and moved the new feature/fixes to a fresh `[0.7.0b3]` CHANGELOG section, leaving `[0.7.0b2]` as the already-shipped task_chips entry. Are the two version files consistent and is the CHANGELOG framing correct for someone upgrading?

Top issues first; minor ones welcome too.

---

PR #111, opened 2026-06-29
fix: clean up orphaned devices after Problem Sensor Sync disable/exclude

psym88:
When Problem Sensor Sync is disabled, or a sensor/device/area/label is excluded, the synced task and its per-task entities are removed on entry reload (issue #104) — but the device was left behind, lingering under Settings → Devices & Services → Home Keeper with zero entities.

Add devices.async_prune_orphaned_devices, run after platforms forward (so the entity registry already reflects the current set), dropping Home Keeper from any non-asset device that no longer carries one of our entities: async_update_device removes a self-owned device outright (we were its last config entry) and merely detaches us from a shared device its real owner keeps. Virtual asset devices are intentionally entity-less and reconciled elsewhere, so they're skipped. The pass also covers ordinary task deletion.

Add an integration test (self-owned device removed after the attached task is deleted) and bump to 0.7.0b3.

Claude-Session: https://claude.ai/code/session_016jXmCcFaJg2WWsZ4fpBMn7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning (with PEP 440 pre-release suffixes — `bN`/`aN`/`rcN` — for betas).
 
+## [0.7.0b3]
+
+### Fixed
+
+- **Problem Sensor Sync no longer leaves empty device cards behind.** When you
+  disable Problem Sensor Sync — or exclude an individual sensor, device, area, or
+  label — the synced task and its entities are removed as before, and now the
+  associated device is cleaned up too: a device Home Keeper created for the synced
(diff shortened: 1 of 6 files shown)


psym88:
@prestomation To avoid potential conflicts with your repository, should I remove the version numbers from my pull requests? Also, is it okay to skip "/q review" in my fork?

prestomation:
Good questions. For versions, yes probably best to leave those out. 

As for q, I suggest signing up for Amazon Q for GitHub and enabling it. Assuming you are using PRs in your fork, it will review your code (without using your claude tokens) and Claude will automatically fix. Mine isn't working atm because I ran out for the month.. 
If you don't want to use it please instead ask Claude after writing the code to 'use a subagent to perform a comprehensive code review 'and it will usually find some bugs. 

If you don't want to use it that's fine but you'll have to figure out how to manage pulling it out of your instructions but don't remove it in your PRs.

PR #113, opened 2026-06-30
fix(ci): deploy docs via workflow_call from release.yml

prestomation:
## Problem

The docs site (https://prestomation.github.io/ha-home-keeper/) stopped updating after **June 24** — including across the **v0.7.0 stable release** cut on June 30.

### Root cause

PR #84 switched `docs-deploy.yml` from a `push: [main]` trigger to `release: types: [released]`. But that release is created by `release.yml` using the default `GITHUB_TOKEN`, and **events triggered by `GITHUB_TOKEN` never start a new workflow run** (GitHub's anti-recursion safeguard). So the `released` event was silently suppressed and `docs-deploy.yml` never fired.

Evidence:
- `docs-deploy.yml` run history: 7 runs, **all `push` events**, last on 2026-06-24. Zero `release`-triggered runs.
- The v0.7.0 release exists (`prerelease: false`) but its author/uploader is `github-actions[bot]` — i.e. created by `release.yml`'s token.

The version badge in the navbar (`docusaurus.config.ts`, driven by `DOCS_VERSION`) was also added in #84, so it never shipped either — it'll start showing `v0.7.0` once this deploys.

## Fix

Make `docs-deploy.yml` a **reusable workflow** (`workflow_call`) invoked directly by a new `deploy-docs` job in `release.yml`. A `workflow_call` runs inside the same triggering run, so the `GITHUB_TOKEN` restriction doesn't apply. The "deploy only stable, pinned to the release tag" guarantee is preserved.

- **`docs-deploy.yml`**: replace the `release` trigger with `workflow_call` (+ optional `ref` input); add an optional `ref` input to `workflow_dispatch`; checkout `ref: ${{ inputs.ref || github.ref }}`; drop the prerelease `if:` guard (the caller now decides when to deploy).
- **`release.yml`**: promote step outputs to job outputs; add a `deploy-docs` job (`needs: release`) gated on `release == 'true' && prerelease == 'false' && event != pull_request` that calls `./.github/workflows/docs-deploy.yml` with `ref: v<version>`.
- **`website/README.md`**: update the Deployment section to document the new mechanism.

## One-time recovery

After merge, manually dispatch `docs-deploy.yml` with `ref: v0.7.0` to un-stick the live site immediately. Every stable release after that deploys automatically.

## Testing

- Both workflow YAMLs validated (`yaml.safe_load`).
- CI-only change — no app/panel UI code, so no screenshot gate.
- End-to-end verification (dispatched run + next stable release) happens post-merge.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VqVAbAwAZ3cS9qcVDpdefz)_

### DIFF
--- a/.github/workflows/docs-deploy.yml
+++ b/.github/workflows/docs-deploy.yml
@@ -1,14 +1,28 @@
 name: Docs site — deploy
 
 on:
-  # Deploy only when a stable (non-pre-release) GitHub Release is published.
-  # This keeps the live docs pinned to the latest release so users don't see
-  # documentation for unreleased features.
-  release:
-    types: [released]
-  # Keep workflow_dispatch for emergency manual deploys (e.g. urgent typo fix
-  # that shouldn't wait for the next release).
+  # Called by release.yml after a *stable* release is cut. We can't use the
+  # `release: [released]` event here: that release is created by release.yml
+  # with the default GITHUB_TOKEN, and events triggered by GITHUB_TOKEN never
+  # start a new workflow run (GitHub's anti-recursion safeguard) — so the event
+  # would silently never fire. A workflow_call runs inside the same triggering
+  # run, sidestepping that restriction. `ref` is the tag to build docs from, so
+  # the live site stays pinned to the released commit (not main HEAD).
+  workflow_call:
+    inputs:
+      ref:
+        description: Git ref (release tag) to build the docs from.
+        required: false
+        type: string
+  # Manual deploys: emergency fixes, or the one-time recovery of a release that
+  # shipped before this pipeline was wired up. Pass `ref` (e.g. v0.7.0) to pin
+  # to a tag; omit it to build from the default branch HEAD.
   workflow_dispatch:
+    inputs:
+      ref:
+        description: Git ref (release tag) to build the docs from. Defaults to the branch HEAD.
+        required: false
+        type: string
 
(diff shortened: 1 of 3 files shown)


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first (don't withhold minor ones). This is a CI/release-pipeline fix that makes `docs-deploy.yml` a reusable workflow called from `release.yml`. Focus on:

- **Correctness of the GitHub Actions semantics**: Is `workflow_call` from `release.yml`'s `deploy-docs` job the right fix for the "GITHUB_TOKEN events don't trigger workflows" problem? Are there cases where it still won't run, or runs when it shouldn't (e.g. prerelease, re-run on an existing tag, pull_request events)?
- **Permissions**: Does the reusable workflow get the `contents: write` it needs under `workflow_call`? Is the caller passing permissions correctly, and is anything over-privileged?
- **The `ref` resolution** (`inputs.ref || github.ref`) across both `workflow_call` and `workflow_dispatch` — any edge cases (tag vs branch, missing tag, detached HEAD)?
- **Job-output plumbing** in `release.yml` (string `'true'`/`'false'` comparisons in the `if:` gate) — off-by-one / truthiness traps.
- **Concurrency**: the `docs-deploy` concurrency group with `cancel-in-progress` — could a release deploy cancel an in-flight PR preview or vice versa, or cancel itself?
- **HA best practices / maintainability** of the workflow structure.

---

PR #117, opened 2026-07-01
Add in-form help text and a live sensor hint to the task editor

prestomation:
## Why

Creating a **sensor-based** task was hard to reason about. If your run-hours sensor already reads **660** and you set a target of **100**, the form gave no hint whether the task fires at 100 or 760, or how it resets after completion — it just showed "Target (units)" with no context, and the detail later read *"0 of 100 used"* as if it had ignored the 660. (Reported from the create flow — the whole form lacked guidance.)

The behavior was correct but undocumented: the target counts usage **from the sensor's reading at creation** (the *baseline*), so 660 + 100 → due at **760**, and the baseline re-stamps on each completion.

## What

Make the task create/edit form self-explanatory:

- **Per-field helper text on every field** via `ha-form`'s native `computeHelper` (keyed `help.<field>`, empty where unauthored so helpers appear only where written). The **Target** helper spells out the baseline model directly: *"Counting starts from the sensor's reading now — so if it reads 660, the task is due at 760, then 100 further after each completion."*
- **A live, computed hint** under the sensor fields — a pure `sensorHintText()` in `forms.ts` that reads the bound entity and states the next due point: *"This sensor reads 660 h now. The task becomes due at 760 h, then every 100 h after each completion."* It updates **in place** on each edit so typing the target never loses focus. Threshold mode gets its own variant (*"…due when the sensor is ≥ 90"*).
- **A sensor-mode primer** (with a "Learn more" docs link) above the fields, and a **`?` help icon** by the form title linking to the sensor-tasks User Guide page.

Supporting changes:
- README "Sensor-based tasks" section expanded with the baseline/reset model and the 660 → 760 worked example.
- Unit tests for `sensorHintText` (usage w/ + w/o reading, threshold, hold, empty states).
- The 27 new strings localized across all 15 non-English locales.
- Bumped to **0.8.0b2** (new user-facing feature) + CHANGELOG entry.

## Screenshots

**Sensor task — usage / meter** (per-field helpers, sensor primer, `?` icon, and the live hint at the bottom):

Sensor usage/meter create form with helper text and live hint

**Sensor task — threshold mode** (comparison/value/hold helpers, threshold hint):

Sensor threshold create form with helper text and live hint

**Whole-form helper text on a non-sensor task** (edit view — Every/Unit, device, consumable, labels, card links, completion):

Edit task form showing per-field helper text across the whole form

## Testing

- `vitest run` — 195 passed (incl. new `sensorHintText` cases + i18n parity/leak guards for the 15 locales).
- `tsc --noEmit` — clean.
- Screenshots captured with the Playwright harness (`screenshots.capture.ts`) against the seeded HA container and visually inspected.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DFDamvF85dFXYkFhE1dgjF)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. This PR adds in-form help text (via `ha-form` `computeHelper`), a live sensor hint, a docs link, and 15-locale translations to the task editor. Focus on:

- **Correctness of `sensorHintText` (`forms.ts`)** — reading/target parsing (`Number` of strings vs numbers, `NaN`/empty/zero-target guards), the `reading + target` arithmetic, threshold vs usage branching, comparison-symbol mapping, and the flat `sensor_*` edit-state vs nested `task.sensor` fallback. Any input that yields `"undefined"`/`"NaN"` in the rendered string, or a misleading due-point?
- **`_updateSensorHint` / `_sensorLive` (`panel.ts`)** — the in-place hint update path (avoiding a re-render to keep number-input focus): are there edit transitions where the hint goes stale or fails to hide? Unit extraction from `unit_of_measurement`. Any XSS risk from `help.sensor.section` being injected via `innerHTML` (it's a trusted locale string with a constant `{url}`, but confirm)?
- **`computeHelper` wiring** — the `t('help.'+name)` missing-key sentinel check; correctness for grid sub-fields (interval/unit/freq) and locked/omitted fields.
- **i18n** — placeholder-token parity across the 15 locales, and whether any dynamic `help.*` key could be requested without a backing string.
- **HA best practices / maintainability** — helper-text scoping to the task form only, and whether the live hint belongs in `forms.ts` (pure) vs `panel.ts` (DOM) as split.

Please also flag any timezone/DST, off-by-one, or error-path edge cases I may have missed.

---

PR #121, opened 2026-07-01
Localize reconciler-generated task names to the HA language

### LINKED ISSUE #119: Missing translations
Certain things have missing translations, specifically things that are templated.

For example, for the "Replace {item}" task, it's missing the proper polish translation:

Image

prestomation:
## Summary

Fixes #119.

The maintenance-task name Home Keeper auto-generates for a wear part (e.g.
"Replace Anode rod (Garage water heater)") was a hardcoded English f-string in
`reconcile.py`, persisted verbatim — so it never translated, even with HA set to
another language.

A task's `name` is server-side **global** data: one value shared by the to-do
item, the calendar event, notifications, the per-task device-page entity names,
and every API consumer. It can't be re-rendered per viewer the way the panel's
static UI is. So it's now translated **once, at write time, into Home Assistant's
configured language** (`hass.config.language` — the household's primary language).

The one tradeoff, accepted by design: a mixed-language household sees a single
language (the instance's) on these generated names everywhere. In exchange, the
name is translated on **every** surface — panel, to-do, calendar, notifications,
device pages — not just the panel.

## What changed (backend only)

- **`const.py`** — `WEAR_TASK_NAME_TEMPLATES` / `APPLIANCE_FALLBACK_NAMES` (16
  languages) + a pure `resolve_wear_task_naming()` helper (exact → base-language →
  English fallback, mirroring the panel's locale resolution).
- **`store.reconcile_part_tasks`** — resolves the template + unnamed-appliance
  fallback word from `hass.config.language` and passes them into the **pure**
  reconciler, which defaults to English (stays HA-free and unit-testable).
- **`reconcile.py`** — formats the name from the injected strings instead of a
  hardcoded f-string. Because the name is recomputed every pass, a language change
  is picked up as ordinary **name drift** by the existing update branch.
- **`__init__.py`** — an `EVENT_CORE_CONFIG_UPDATE` listener reloads the entry when
  the HA language changes, so reconcile relocalizes every generated name (and
  recreates the per-task entities under the new name) immediately, not just on the
  next restart.

The panel already renders `task.name`, which is now localized, so no frontend
changes are needed.

## Verification

**End-to-end against the Docker HA** — switching the instance language to Polish
over the WS API fired the listener → entry reload → reconcile:

```
BEFORE (instance language en): Replace Anode rod (Garage water heater)
config/core/update {language: pl}  →  EVENT_CORE_CONFIG_UPDATE → reload → reconcile
AFTER  (instance language pl): Wymień Anode rod (Garage water heater)
```

Panel with the instance language set to Polish (the generated wear-part tasks are
translated; user-authored tasks like "Replace water filter" correctly stay as
typed):

Home Keeper panel with instance language Polish: the reconciler-generated wear-part tasks show 'Wymień Anode rod (Garage water heater)' and 'Wymień T&P relief valve (Garage water heater)'

- [x] `pytest tests/unit` — 430 passed (new `test_wear_task_naming.py` covers the
      resolver, 16-language parity, placeholder presence, and no-untranslated-leaks;
      `test_reconcile.py` covers localized create, localized fallback word, and the
      language-change-as-drift rewrite).
- [x] `npx vitest run` — 195 passed.
- [x] `mypy` on the changed backend files — clean.
- [x] `ruff` check + format — clean.
- [x] End-to-end Docker HA verification above.


prestomation:
/q review Please give critical, skeptical feedback on this fix for #119 (reconciler-generated task names, e.g. "Replace {part} ({asset})", not translating). Specifically scrutinize:

- **Correctness**: does `name_template` actually get recomputed/persisted on every case the reconciler already detects as drift (asset rename, part rename, interval/unit changes)? Any path where `name` and `name_template` could get out of sync? Is the empty-asset-name fallback (`asset.get("name") or ""` in the template vs. the `t('appliance.fallbackName')` substitution in `taskDisplayName`) handled correctly and consistently?
- **Maintainability**: is scattering `taskDisplayName()` across ~8 call sites in `panel.ts`/`card.ts` the right call, or should this be centralized further? Is storing `name_template` as an untyped `{part, asset}` bag (rather than a generic `{key, params}` shape) too narrow if a second templated-name case appears later?
- **i18n quality**: are the 15 non-English translations of `part.replaceTaskName` idiomatic (not just literally correct) for each locale's existing verb/vocabulary conventions?
- **Scope**: is it reasonable that native `todo`/`calendar` entities, notifications, and events keep showing the plain English `name` (documented as a deliberate limitation), or should that have been addressed too?

---

prestomation:
/q review This PR was reworked since your last review: instead of a per-viewer frontend template, the reconciler-generated task name is now localized to `hass.config.language` at write time (backend only, no frontend changes). Please give critical, skeptical feedback, most serious first, on:

- **Correctness / re-entrancy**: the `__init__` `EVENT_CORE_CONFIG_UPDATE` listener calls `hass.config_entries.async_reload(entry.entry_id)` from inside the event callback (guarded by a captured `setup_language` comparison). Is reloading an entry from its own bus-event listener safe here, or is there a re-entrancy / double-reload / unload-ordering hazard? Could `EVENT_CORE_CONFIG_UPDATE` fire for non-language changes often enough to matter, and is the "compare to setup-time language" guard robust (e.g. rapid successive changes, or the event firing before `hass.config.language` is updated)?
- **Purity boundary**: `reconcile.py` stays HA-free by taking `name_template`/`appliance_fallback` as params defaulting to English, with the store resolving them from `hass.config.language`. Is that the right seam, or does defaulting to English in the pure layer risk a silent wrong-language bug if a caller forgets to pass them?
- **Language resolution**: `const.resolve_wear_task_naming` does exact → case-insensitive → base-language → English. Any HA language codes that would resolve wrong (e.g. `zh-Hans` vs `zh-Hant`, `pt-BR` vs `pt`, `nb`/`no`)? Is `str.format(part=…, asset=…)` safe against a template that contained stray `{`/`}` (none do today, but is a `KeyError`/`ValueError` reachable)?
- **Behavior/UX**: baking the instance language into the stored `name` means a mixed-language household sees one language on these names everywhere, and the name is no longer a stable English key. Given the docs already say integrations should match on `source`/`id` (not name), is that acceptable, or does anything in-repo rely on the English name?
- **i18n quality**: are the 16 `WEAR_TASK_NAME_TEMPLATES` idiomatic (verb form/word order) per locale, and does reusing the panel's `appliance.fallbackName` words for `APPLIANCE_FALLBACK_NAMES` read correctly in context?

Don't withhold minor issues.

---

prestomation:
Thorough review — thanks. I dug into the two load-bearing claims against HA core source, and both are false positives, so no code changes from this pass. Evidence below.

**1. Re-entrancy / event ordering (the "must-fix") — premise is backwards.** The claim is that `EVENT_CORE_CONFIG_UPDATE` fires *before* `hass.config.language` is updated. It's the reverse. `homeassistant/core.py` `Config.async_update`:

```python
self._update(source=ConfigSource.STORAGE, **kwargs)   # attributes (incl. language) set here
await self._async_store()
self.hass.bus.async_fire(EVENT_CORE_CONFIG_UPDATE, kwargs)   # event fired AFTER
```

So `hass.config.language` already holds the new value when the listener runs — which is exactly why the live Docker test relocalized to Polish. Had the event fired first, my `if hass.config.language == setup_language: return` guard would have early-returned and the name would have stayed English.

Overlapping reloads on rapid toggles also can't happen: `ConfigEntries.async_reload` runs `async with entry.reload_lock:` (a per-entry `asyncio.Lock`), so concurrent reloads of the same entry serialize. This mirrors the existing `_async_options_updated` listener in this same file, which reloads identically with no extra guard.

**3 (str.format). KeyError on brace-containing names — false positive.** `str.format` makes a single pass over the *template*; it does not re-scan substituted values. A part literally named `Valve {old}`:

```python
>>> "Replace {part} ({asset})".format(part="Valve {old}", asset="Fridge")
'Replace Valve {old} (Fridge)'   # no KeyError
```

The only braces `.format` interprets live in the template, which we own. No try/except needed.

**2. English default = "silent wrong language" — intentional, not a failure mode.** English here isn't a bug-fallback; it's the *same* value `resolve_wear_task_naming` returns for any unknown/unset language. The pure reconciler defaulting to English is therefore identical to "HA is set to a language we don't ship" — a coherent, documented fallback, and what keeps the module HA-free and unit-testable. There's a single caller (`store`), covered by the end-to-end Docker test. Making the params required would churn every caller/test to close a hypothetical future-refactor gap while giving up the standalone ergonomics — net negative, so I'm keeping the defaults.

**3 (nb/no). Matches the panel by design.** HA uses `nb` for Norwegian Bokmål (as does our frontend locale set), and `_pick_localized` lowercases both sides, so `NB`/`NO` case variants resolve fine. A hypothetical legacy `no` code would fall back to English — precisely what the frontend `i18n.resolve` already does. Keeping backend and panel resolution identical is deliberate; I won't special-case `no` on the backend only.

**4 / 5 / minors.** Agreed on the UX tradeoff, and idiomaticity is deferred to native speakers (placeholder parity + the no-English-leak guard in `test_wear_task_naming.py` are enforced). The minor notes — capture timing (a reload re-captures `setup_language`), the `-` separator (HA is BCP-47, matches the frontend), and caching a two-dict-lookup — are negligible given reconcile runs on setup / asset mutation, not a hot loop.

Net: nothing to change, but genuinely useful pressure-testing.

---

PR #120, opened 2026-07-01
Add clickable product-page URL to replaceable parts

prestomation:
## Summary

- A part (consumable or wear item) can now carry a **product URL** — a link to where you buy it (e.g. an Amazon listing).
- When set, the part's name in the appliance detail page renders as a clickable link that opens the product page in a new tab, so reordering is one click away.
- Backend validates the URL as http(s) via the existing `_normalize_http_url` helper (already used for asset documents/metadata links).
- The pre-existing `Part.notes` field (defined but never wired into any UI) is left untouched — unrelated to this change, out of scope.
- Bumped to `0.8.0b3` as a new beta feature release (per the "always cut a beta for a new feature" workflow rule).

Devices (appliances) were **not** touched — they already support clickable product links today via the existing **Metadata** (`type: Link`) and **Documents** (`kind: Link`) sections, so the gap was specific to parts. See the plan discussion on #118 for the scoping rationale.

Fixes #118

## Screenshots

Appliance detail page — "Sediment pre-filter" now renders as a clickable link (its `url` points to a product page):

Appliance detail page showing a part rendered as a clickable link

## Test plan

- [x] `pytest tests/unit -v` — 421 passed, including new part-URL validation tests
- [x] `mypy custom_components/home_keeper` — not runnable in this sandbox (the `homeassistant` package fails to build an unrelated legacy dependency, `PyRIC`, here); please confirm via CI
- [x] Frontend `npx vitest run` (from repo root) — 195 passed, including full i18n locale-parity checks for the new `field.part_url` key across all 16 locales
- [x] `npx tsc --noEmit` — no type errors
- [x] `bash ci/e2e-up.sh` — HA + Playwright e2e stack started and ran successfully
- [x] Manually verified in a running panel: added a product URL to a part, saved, confirmed the part name renders as a clickable link opening in a new tab; confirmed an empty URL still renders as plain text
- [x] Re-captured and visually inspected `docs/images/8-panel-appliance-detail.png`

---
_Generated by [Claude Code](https://claude.ai/code/session_01TvHbUEt1MDs2N3btxvnV9X)_

Issue #118, opened 2026-07-01
URLs for product pages for devices and replacable parts


mdarocha:
It'd be great, especially for the repleacable parts feature, to be able to add an URL to a product page/online shop (like Amazon) where the required part can be bought. And for that URL to be nicely surfaced in the UI. Currently we have a field for that, but putting an URL there results in a bad UX, with the URL not being clickable or anything like that

prestomation:
Released in 0.8.0

PR #122, opened 2026-07-01
Add part file attachments and clickable part links on connected tasks

prestomation:
## Summary

Follow-up to the part `url` field (#118, #120). Two more asks:

**A. File upload for parts.** A part can now carry a single attached file (a
receipt, spec sheet, or photo) — reusing the exact upload/storage/validation
machinery appliance documents already use (`documents.py`/`manuals.py`), keyed
by the part's own id instead of a minted document id. Upload is HTTP-only (a
service call can't carry binary bytes, same precedent as asset documents);
removal is both a websocket command and a new `home_keeper.remove_part_file`
service. A part's attached file is upload-only end to end: the generic
`add_asset`/`update_asset` path can neither inject nor clear it — enforced at
two layers, the service's voluptuous schema (which doesn't declare the field at
all) and `assets._merge_parts` (which always restores the stored values).

**B. The part's link now surfaces on connected tasks.** A task linked to a part
(an auto-generated wear-item task, or a manually-linked consumable) now renders
that part's name as a clickable link to its product page — in the panel's task
detail "Consumable link" row, and as a chip on the dashboard card's task row —
when the part has a `url`.

**Also fixed:** `_PART_SCHEMA` (the `add_asset`/`update_asset` **service**
validation schema) never actually included the `url` field added in #120 — the
websocket/panel path worked, but calling the service directly with a part `url`
was silently rejected with a 400. Found via the new integration tests; fixed
here.

This is folded into the still-unreleased `0.8.0b3` (see CHANGELOG) rather than
bumping to a new beta, since #120 (which opened that section) hasn't merged/shipped
yet.

## Screenshots

A part's attached file — a card with icon, filename · size · type, Open / Remove —
right in its editor:

Part editor showing an attached file card

Appliance detail read view — the Anode rod part shows a small paperclip indicator
for its attached file:

Appliance detail page showing a part with an attached-file indicator

Task detail — "Consumable link" now opens the linked part's product page directly:

Task detail page showing a clickable consumable link

## Test plan

- [x] `pytest tests/unit -v` — 423 passed, incl. new `set_part_file`/`clear_part_file`
  and merge-isolation tests
- [x] `bash ci/test-python-integration.sh` (Docker) — 74 passed, incl. 7 new
  `test_part_files.py` tests (upload/download/remove round trip, reject bad type,
  re-upload replaces old blob, generic-write isolation, asset-delete cleanup,
  unknown-part error, auth-gating)
- [x] `mypy custom_components/home_keeper` — not runnable in this sandbox (the
  `homeassistant` package fails to build an unrelated legacy dependency here);
  please confirm via CI
- [x] `npx vitest run` (repo root) — 197 passed, incl. a new card.ts test for the
  linked-part chip and full i18n locale-parity for the two new button labels
  (`field.part_url`-style translation, all 16 locales)
- [x] `npx tsc --noEmit` — no type errors
- [x] `ruff check` / `ruff format --check` — clean
- [x] `bash ci/e2e-up.sh` — panel builds and HA starts cleanly
- [x] Manually exercised in a running panel: attached/replaced/removed a part's
  file; opened a task linked to a part with a URL and confirmed the clickable
  link in both the panel task detail and (via vitest) the dashboard card;
  confirmed editing an unrelated part field doesn't drop its attached file
- [x] Re-captured and visually inspected all three screenshots above

---
_Generated by [Claude Code](https://claude.ai/code/session_01TvHbUEt1MDs2N3btxvnV9X)_


prestomation:
/q review Please give critical, skeptical feedback on this change (a part's single attached file, plus a clickable part link on connected tasks). Focus especially on:
- **Security**: is the part-file upload/serving path (`manuals.HomeKeeperPartFileView`) as safe as the existing asset-document path it mirrors — magic-byte sniffing, size cap, filename sanitization, path-traversal guard, auth gating? Does reusing the asset-document blob storage tree (keyed by part id instead of document id) introduce any collision or cross-asset leakage risk?
- **Correctness**: does the isolation actually hold end-to-end — can a generic `add_asset`/`update_asset` (service, websocket, *or* a crafted raw payload) inject or clear a part's `file_name`/`file_content_type`/`file_size`? Is the re-upload-replaces-old-blob logic in `HomeKeeperPartFileView.post` correct (old blob deleted only after the new one writes successfully)? Does `remove_part_file`'s idempotency (no-op when there's no file) make sense, or could it mask a real error?
- **Maintainability**: is duplicating a chunk of `HomeKeeperDocumentView`'s POST/GET logic into `HomeKeeperPartFileView` (rather than extracting a shared helper) a reasonable tradeoff given the part view is meaningfully smaller (one file slot, no link/file branching)?
- **HA best practices**: does the new `remove_part_file` service (and its schema/`services.yaml`/`strings.json`/16-locale translations) follow the same conventions as the existing document services?

---

prestomation:
Addressed in 72499ab:

- **Storage-key discriminator**: part files now store on disk under `part_<part_id>__<filename>` instead of the bare part id, via new `manuals.async_save_part_file`/`async_read_part_file`/`async_delete_part_file` wrappers, so a part id can never collide with an asset document id in the same per-asset directory. The HTTP route and signed view path are unaffected (bare `part_id` — a separate namespace with no such collision risk), so this is storage-internal only.
- **Upload-parsing duplication**: extracted the ~30-line multipart-parsing loop shared by `HomeKeeperDocumentView.post` and `HomeKeeperPartFileView.post` into a single `_parse_upload()` helper.
- **Isolation by omission**: already covered by `test_merge_update_cannot_inject_or_clear_part_file` in `tests/unit/test_assets.py`.

Verified: full unit suite (423), Docker integration suite (74, including all 7 `test_part_files.py` cases against the new storage-key scheme), full e2e Playwright suite (37), frontend vitest (197), `tsc --noEmit`, `ruff check .` — all clean.

The e2e CI failure on this branch was unrelated to the review — a pre-existing `card.spec.ts` test hard-coded a document-chip count of 3 for the water-filter task's dashboard row, which is now 4 because of this PR's new part-link chip. Updated the test's expectation in the same commit.

/q review please take another pass focused on: (1) whether the `part_<id>` storage-key discriminator is applied consistently everywhere a part file's on-disk path is derived (no remaining call site still using the bare part id), and (2) any remaining correctness or security issues you'd flag as more serious than what's already been raised.

---

prestomation:
Closed #120 (superseded — its commits are ancestors of this branch, so this PR is a full superset) and rebased this branch onto current `main` (`7b2bcb8`, which had picked up two unrelated code-review-fix PRs, #124/#125, in the meantime).

Reconciled two real conflicts during the rebase, not just mechanical ones:
- **CHANGELOG.md** — main had renamed `## [Unreleased]` bugfix notes while this branch had renamed the same section to `## [0.8.0b3]` for the new features. Merged into one `0.8.0b3` section with both an `Added` and a `Fixed` subsection.
- **`manuals.py`** — main's code-review fixes (#125, "N10") changed `HomeKeeperDocumentView` to stream `GET` via `web.FileResponse` instead of buffering, write the blob to disk *before* persisting store metadata (avoiding a GET-404 race right after the event fires), and dropped the dead `declared_type` param from `validate_upload`. Rather than just picking a side, I brought `HomeKeeperPartFileView` up to the same pattern for consistency — same streaming GET, same write-before-persist ordering (with a guard for the edge case where a same-filename re-upload would otherwise get its own valid blob deleted on a metadata-write failure), same signature change.

Re-verified everything after reconciling: 485 unit + 75 Docker integration (including all 7 `test_part_files.py` cases against the new write-order/streaming logic) + 37 e2e + 216 frontend, `tsc --noEmit`, `ruff check .`/`ruff format --check .` — all clean. Force-pushed (`--force-with-lease`) the rebased history.

---

PR #12, opened 2026-06-15
Add high-fidelity deep linking and fix the panel Back button

prestomation:
## Problem

The panel kept all navigation state (active tab, open detail page) in instance
fields that never touched the URL or browser history. So nothing was
deep-linkable, refreshing lost your place, and — the reported bug — **the
browser Back button ejected you from Home Keeper entirely** instead of stepping
back to the previous in-panel page, because the panel never pushed any history
entries.

## Approach

Make the **URL the single source of truth** for panel navigation (high-fidelity
deep linking).

- **URL scheme** under the `/home-keeper` prefix: `/tasks` (default),
  `/appliances`, `/tasks/<id>`, `/appliances/<id>` (asset detail under the
  `appliances` segment). Forms stay ephemeral overlays and are intentionally
  *not* deep-linked.
- **`utils.ts`** — pure `parseRoute` / `buildPath` (round-trip tested). Unknown
  / empty paths fall back to the tasks list.
- **`panel.ts`** — a `set route` setter parses HA's `route.path` and is the only
  place that flips `_view`/`_detail`, so it handles deep links and Back/Forward.
  A single `_navigate(loc, replace?)` helper changes the URL
  (`pushState`/`replaceState` + a bubbling `location-changed` event); drill-in
  steps **push**, tab switches and detail close/delete **replace** so Back never
  retraces a tab toggle or returns to a deleted object.
- Documented the convention in `.amazonq/rules/architecture-and-code.md` and
  `AGENTS.md`.

## Behavior

- Tabs and detail pages are linkable, bookmarkable, and refresh-safe.
- Browser Back (and the in-panel `‹ Back` button) now go detail → list, staying
  inside the panel.

## Tests

- Unit (`vitest`): `parseRoute`/`buildPath` parse + round-trip — **78 passed**.
- e2e (Playwright): opening a detail updates the URL; a detail URL deep-links
  straight to the detail page; the appliances tab is deep-linkable; browser Back
  and the in-panel Back button both return to the list without leaving the panel
  — **15 passed**.

No screenshots: this is a navigation/behavior change with no visual diff to the
panel itself (the change is in the URL bar and Back behavior, exercised by the
e2e tests above).

https://claude.ai/code/session_01H88vYQf3MCE56ZH1hSVqUS

---
_Generated by [Claude Code](https://claude.ai/code/session_01H88vYQf3MCE56ZH1hSVqUS)_

PR #123, opened 2026-07-01
Docs preview: link to the doc pages a PR changed

prestomation:
## What & why

The docs-site PR preview posts a single sticky comment with the base preview URL. On a PR that edits documentation, a reviewer still has to hunt through the site to find the pages that actually changed. This adds a **second sticky comment** listing deep links to exactly the generated pages the PR touched, so reviewers can jump straight to the rendered surfaces.

It also closes a gap: the preview only triggered on `website/**` / `docs/images/**`, so edits to the **canonical** doc sources (`README.md`, `docs/*.md`, `CHANGELOG.md`) — from which the site content is generated — built no preview at all. The trigger now covers them.

## How it works

- **`website/scripts/doc-map.mjs`** (new) — the source→page mapping (`splitByH2`, `USER_SECTIONS`, `DEV_DOCS`) extracted from `sync-docs.mjs` into a shared, side-effect-free module so the preview logic and the generator can never drift.
- **`website/scripts/changed-pages.mjs`** (new) — maps a PR's changed files to preview routes and renders the sticky-comment body. README is **section-granular**: it diffs the base vs head README by `##` section and links only the User Guide pages whose section actually changed (skipping unpublished sections like *Development*). Developer docs, the changelog, and the hand-authored landing/intro pages map 1:1.
- **`.github/workflows/docs-preview.yml`** — widened triggers, `fetch-depth: 0` to diff against the PR base, a step that computes the changed pages, and a `find-comment` + `create-or-update-comment` pair that upserts the sticky comment (and updates it to "none" if a later push removes all doc changes).

Mapping (all under `…/pr-preview/pr-<n>/`):

| Source | Page route |
|---|---|
| `README.md` `##` section | `/docs/guide/<slug>` (only changed sections) |
| `docs/INTEGRATING.md` / `EVENTS.md` / `DESIGN.md` / `GLUE_INTEGRATIONS.md` | `/developer/…` |
| `CHANGELOG.md` | `/docs/release-notes` |
| `website/docs/intro.md`, `website/src/pages/index.tsx` | `/docs/intro`, `/` |

## Testing

- `tests/frontend/changed-pages.test.js` — 13 new cases covering section-change detection (edited body, added section, preamble-only change ignored, no-base fallback), file→route mapping (dedup, unpublished sections skipped, non-doc files ignored), and comment rendering (marker, absolute URLs, trailing-slash handling, empty case).
- Full JS suite green (208 tests); `sync-docs.mjs` still generates all pages after the refactor; CLI exercised end-to-end against a simulated README edit.

## Notes

Developer-only change (CI + build tooling), so no CHANGELOG entry, beta bump, or panel screenshots per `AGENTS.md`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FhqE5n3XQcqzT2fvtHwJiZ)_

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -5,11 +5,16 @@ on:
     types: [opened, reopened, synchronize, closed]
     paths:
       - 'website/**'
-      - 'docs/images/**'
+      # Canonical doc sources: the site content is generated from these, so a
+      # change here changes the rendered pages (see website/scripts/sync-docs.mjs).
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - 'docs/**'
       - '.github/workflows/docs-preview.yml'
 
-# Previews deploy to pr-preview/pr-<n>/ on the gh-pages branch and a sticky
-# comment with the URL is posted on the PR.
+# Previews deploy to pr-preview/pr-<n>/ on the gh-pages branch and two sticky
+# comments are posted on the PR: the preview URL (pr-preview-action) and a list
+# of deep links to the doc pages this PR changed (changed-pages.mjs).
 permissions:
   contents: write
   pull-requests: write
@@ -25,6 +30,9 @@ jobs:
     if: github.event.pull_request.head.repo.full_name == github.repository
     steps:
       - uses: actions/checkout@v4
+        with:
+          # Need history back to the PR base so we can diff which docs changed.
+          fetch-depth: 0
 
       - uses: actions/setup-node@v4
         if: github.event.action != 'closed'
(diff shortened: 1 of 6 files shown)


prestomation:
/q review Please be critical and skeptical, surfacing the most serious issues first (but don't withhold minor ones). Focus on:

- **Correctness of the README section diff** (`readmeChangedHeadings` in `website/scripts/changed-pages.mjs`): edge cases around fenced code containing `##`, sections that appear more than once, heading renames, changes confined to the preamble, and the empty-base fallback. Does the body-equality comparison miss or over-report any real change?
- **Source→page mapping drift** — is extracting `USER_SECTIONS`/`DEV_DOCS`/`splitByH2` into `doc-map.mjs` truly the single source of truth, or can the preview links diverge from what `sync-docs.mjs` actually generates (e.g. slug/route mismatches, the `/docs/guide/` vs `/developer/` route prefixes, hand-authored page routes)?
- **Workflow robustness** (`.github/workflows/docs-preview.yml`): the `git diff BASE...HEAD` base-sha handling with `fetch-depth: 0`, `git show BASE:README.md` when README is newly added, the sticky-comment upsert conditions (create vs update vs skip), behavior on the `closed` action, and any permissions/forked-PR concerns.
- **Maintainability**: naming, duplication, and whether the pure/impure split is clean.

Call out anything that could post a wrong or broken link, or silently skip a changed page.

---

prestomation:
Thanks — triaged all 12. Two small changes pushed (1c4e613); the rest are false positives or intentional decisions. Detail:

**Applied**
- **#12** — `STATIC_PAGES` now built in one expression (spread `DEV_DOCS` via `Object.fromEntries`) instead of mutating the const.
- **#2 / #5 (doc)** — added a comment making the rename behavior explicit (see below).

**Declined — verified false**
- **#1 / #10 (fence parsing)** — the claim is backwards. `"``".startsWith("```")` is `false` (a 2-char string can't start with a 3-char prefix), so 2 backticks never close a 3-backtick fence. I ran your exact example through `splitByH2` and it returns `["Real","Real End"]` — precisely the "expected" output you said it would fail to produce. No bug, so no test needed for it. Also note `splitByH2` is the **same** function that generates the site, so the link detector and the generator can never disagree about section boundaries regardless of fence edge cases.
- **#2 (rename = data loss)** — not unique to this feature and not data loss. `sync-docs.mjs` also keys sections by exact `## ` heading text (`byTitle.get(spec.h)` → warn + skip if not found). So renaming a section without updating `USER_SECTIONS` drops the page from the generated site *and* from the link list in lockstep — the page doesn't exist to link to. I documented this in `readmeChangedHeadings`.

**Declined — no functional effect**
- **#4 (empty-base race)** — `README_BASE` is always set by the workflow, and both an empty file and a missing file resolve to `baseMd = ''` (via the `.catch(() => '')`), so the suggested `test -s` guard changes nothing. A hypothetical `git show` error would over-report *valid* extra links, never broken ones — benign.
- **#9 (CLI error handling)** — an unreadable `README_HEAD` already throws and fails the workflow step loudly with a stack trace, which is the desired fail-fast behavior.

**Declined — intentional / already surfaced**
- **#5 (route centralization)** — the drift-prone part (section mapping) is already shared via `doc-map.mjs`. The static routes (`/docs/release-notes`, `/docs/intro`, `/`) are defined by Docusaurus config / frontmatter, so "centralizing" them into `doc-map` wouldn't remove the real source of truth. The `.replace(/\.md$/, '')` is correct and necessary (`d.out` includes `.md`; the served route omits it).
- **#6 / #7 (on-close comment persists)** — intentional; I flagged this decision to the maintainer. Easy to add a cleanup step later if we'd rather blank it on close.
- **#11 (naming)** — `readmeChangedHeadings` returning heading text reads fine; left as-is to avoid churn across the test imports.

---

PR #126, opened 2026-07-02
Maintainability refactors from the code-review fix plan (R1–R4, R8)

prestomation:
## What & why

Works through the **backend maintainability** items (R1–R4, R8) remaining in
`docs/REVIEW_FIX_PLAN.md` from the in-depth code review. All are behavior-preserving
refactors — no user-facing change, no panel UI touched (so the screenshots gate does
not apply, and no CHANGELOG entry per the developer-only convention in AGENTS.md).

The frontend cluster (R5–R7, the `panel.ts` carve-ups) is intentionally **out of
scope** for this PR and stays unchecked in the plan file.

## Commits (one per plan item)

- **R2 — shared coordinator lookup.** Four modules each hand-rolled the same
  `async_entries` + `runtime_data` isinstance loop. Factored into
  `coordinator.get_coordinator` (returns `None` when unloaded) and
  `require_coordinator` (raises the localized `integration_not_loaded`). Callers:
  `websocket_api.py`, `device_trigger.py`, `manuals.py`, `__init__.py`.
- **R1 — extract the service layer into `services.py`.** `__init__.py` had grown to
  ~1140 lines, most of it one ~560-line `_register_services` closure. Moved schemas +
  handlers into `services.py`; handlers are now module-level `async def
  _handle_*(hass, call)` bound to `hass` at registration. A frozen `_ServiceDef` table
  is the single source of truth for **both** register and teardown (killing the
  hand-maintained `_SERVICES` tuple that could drift). A `_translate_errors` decorator
  folds the copy-pasted `KeyError → {task,asset}_not_found` / `TaskValidationError →
  invalid_task` / `AssetValidationError → invalid_asset` blocks; translation keys stay
  literal so the exception-translations AST drift guard still validates them.
  `__init__.py` is now ~176 lines (entry lifecycle only). Same 26 services, schemas,
  response kinds, and localized errors.
- **R4 — single upload allowlist.** `const.DOCUMENT_CONTENT_TYPES` (content type →
  canonical extension) is now the one source; `documents.py` aliases it and
  `assets.py` derives its frozenset from its keys.
- **R3 — shared entity-prune + part lookup.** `entity.async_prune_platform_entities`
  folds the identical registry-walk loop out of all four platforms (each passes an
  `is_stale(uid)` closure holding its own unique-id shapes + live sets). Pure
  `assets.find_part` replaces the duplicated `_part()` in `number.py`/`binary_sensor.py`.
- **R8 — contract hygiene.** (a) Coordinator pins `always_update=True` explicitly and
  documents loudly that `self.data` aliases the store's live task map. (b) The
  `problem_tasks` reconciler copies each touched task before mutating, so it no longer
  mutates the store's live dicts in place (the store already adopts the returned map
  wholesale). (c) `recurrence._parse` promoted to public `recurrence.parse_due`;
  `transitions.py` uses it instead of the private helper.

## Verification

Run locally against Home Assistant 2026.2.3 (Python 3.13):

- `ruff check` + `ruff format --check` — clean (39 files)
- `mypy custom_components/home_keeper` — clean (39 files)
- `pytest tests/unit` — 485 passed
- Import-check of all 38 submodules — OK
- Targeted behavioral smoke tests confirming: the R1 registration table registers +
  fully tears down 26 services with response kinds preserved and the error decorator
  maps/pass-throughs correctly; `find_part` and the prune helper match the originals;
  the R8 reconciler clears/renames on a copy while leaving the store's live dict
  untouched.

Docker-based integration/e2e suites weren't run locally (no Docker HA harness in this
environment) — relying on CI for those.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AfbVBypjtSVszaYi89oMr9)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first and don't withhold minor ones. This PR is a set of behavior-preserving backend maintainability refactors (R1–R4, R8 from docs/REVIEW_FIX_PLAN.md); the bar is "provably no behavior change," so scrutinize accordingly:

- **R1 (services.py extraction) — highest risk.** Verify the `_translate_errors` decorator is behavior-equivalent to the old per-handler try/except: it now *uniformly* maps `TaskValidationError→invalid_task` and `AssetValidationError→invalid_asset` on handlers that previously didn't catch them (e.g. `delete_archived_completion`, `remove_asset_document`) — is translating those a safe superset, or could a handler have relied on one propagating raw? Confirm the `not_found` param picks the right key/placeholder for every handler, and that `adjust_part_stock`'s two-placeholder `unknown_part` is correctly left explicit. Check the `_bind` closure vs `functools.partial` for `SupportsResponse`/coroutine detection across HA versions, and that the `_ServiceDef` table matches the old `_SERVICES` teardown tuple exactly (no service dropped/renamed).
- **R8.** The `problem_tasks` reconciler now copies each touched task with a *shallow* `dict(...)` before mutating — confirm nothing it mutates (including `recurrence.apply_completion`) writes through a shared nested mutable (list/dict) back onto the store's live task. And that pinning `always_update=True` + returning the store's live map is a correct/complete framing.
- **R3.** Confirm each platform's new `is_stale(uid)` closure is exactly equivalent to the prior inline prune logic (prefix/suffix/infix + live-set membership), including the two-shape cases in `sensor.py`/`binary_sensor.py` and domain filtering.
- **R2/R4.** Any import-cycle or missing-teardown risk from the shared helpers.

Also flag correctness, maintainability, performance, security, and HA best-practice issues generally.

---

prestomation:
Thanks — dug into all of these. Triage below.

**1. `_translate_errors` widening (flagged CRITICAL) — not a behavior change in this diff.** I checked the store methods each flagged handler calls, and the "extra" `except` branches are unreachable:

| handler | store method raises | decorator's extra (dead) branch |
|---|---|---|
| `delete_archived_completion` | `KeyError(asset_id)` only | `TaskValidationError`, `AssetValidationError` |
| `remove_asset_document` | `KeyError` only | `TaskValidationError`, `AssetValidationError` |
| `update_asset_document` | `KeyError` + `AssetValidationError` | `TaskValidationError` |

Where the exceptions *can* be raised (`update_asset_document`→`AssetValidationError`, `update_completion`→`TaskValidationError`), the old per-handler code already caught and translated them **identically** — so those paths are byte-for-byte preserved. Everywhere else the added `except` is dead code that can never fire, so there's no observable behavior change today.

On the static-contract widening you're really pointing at: it's intentional and aligned with this project's Platinum `exception-translations` rule (*all* user-facing exceptions must be localized). If a future store change ever made one of these raise `AssetValidationError`, translating it to a localized `invalid_asset` is the **correct** HA behavior, not a raw-500 regression. So I'd rather keep the uniform decorator than reintroduce per-handler `validates=` flags (which would restore exactly the copy-paste variance R1 set out to remove). Open to the flag if you feel strongly, but I think uniform-localize is the right default here.

**2. `apply_completion` shallow-copy (flagged VERIFY) — verified safe.** `recurrence.apply_completion` does `history = list(task.get("completions", []))` — it builds a **fresh** list, mutates that copy, then reassigns `task["completions"] = history`. It never mutates the original list in place, so `dict(task)` fully isolates the store. The R8 change also has a regression check that asserts the store's live task dict (completions included) is untouched after a sensor-clear reconcile.

**3. `_bind` losing handler metadata — fixed.** Added `functools.wraps(handler)` to the bound closure (172866b) so it reports `_handle_add_task` etc. in logs/introspection. Confirmed the callable stays a coroutine function, so HA's job-type detection is unaffected.

**4. `is_stale` not covered in-diff — fair.** The closures live inside each platform's `async_setup_entry`, so they're exercised by the Docker integration/e2e suite rather than the pure unit tests; I also verified each closure is line-for-line equivalent to the prior inline prune (prefix/suffix/infix + live-set membership, including the two-shape `sensor`/`binary_sensor` cases and the domain filter). Happy to extract them to named module-level predicates in a follow-up if you'd like them unit-covered directly.

Thanks for confirming the teardown table, `adjust_part_stock`, and import-cycle points.

---

prestomation:
Closed, but the work is still wanted.

R1 to R4 and R8 stay unchecked in `docs/REVIEW_FIX_PLAN.md`. None of `coordinator.get_coordinator`, `coordinator.require_coordinator`, `entity.async_prune_platform_entities`, `const.DOCUMENT_CONTENT_TYPES` or `recurrence.parse_due` is on main.

This branch cannot carry the work. It conflicts in 11 core modules, and its premise is out of date: it reports `__init__.py` at approximately 1140 lines, and that file is 1592 lines now.

Do R1 again as its own PR. The `services.py` extraction gives the most value, and a reviewer reads it better alone than in a group of 5 items.

---

PR #129, opened 2026-07-04
Auto-create a "buy" task when a spare part runs low

prestomation:
## What & why

Today, a spare part crossing its reorder threshold only fires a `home_keeper_part_low_stock` event — you have to wire your own automation to turn that into a shopping reminder. This adds a **built-in, per-part option** that closes the loop automatically.

Enable **Auto-create buy task** on a stock-tracked part (the option appears once the part has a `reorder_at` threshold). Whenever stock drops to/below the threshold, Home Keeper creates a one-off **"Buy {part}"** task — on the appliance's device page, the to-do list, and the panel. The reminder is **system-managed**: it clears itself once the part is restocked above the threshold (or the option is switched off), and there's only ever one per low spell. Completing the reminder **restocks the part** by a configurable **Restock quantity** (default 1), which normally lifts stock back over the threshold and clears the task in the same step — closing the *low → buy → restocked* loop with no automation to write.

Decisions were confirmed with the requester: **system-managed lifecycle** (declarative, survives restarts) and **completion bumps stock** by a per-part restock quantity. Muted per-field helper text on the part form makes the reorder-event / auto-buy relationship clear.

## Screenshot

A part editor with Auto-create buy task enabled, a Restock quantity field, and helper text under the reorder/auto-buy fields

## How it works

Mirrors the existing wear-part / problem-sensor reconcilers:

- **Model** — new part fields `create_buy_task` + `restock_quantity` (`assets._normalize_part`, the `add_asset`/`update_asset` service schema, and the panel part form gated on a threshold).
- **Reconciler** — new reserved task source namespace `"buy"`; pure `reconcile.reconcile_buy_tasks` is **level-triggered** (a buy task exists iff `part_wants_buy_task && part_is_low`) and idempotent **per low episode** (it counts a buy task whether open *or completed*, so completing it doesn't respawn one while still low).
- **Wiring** — `store.reconcile_buy_tasks` persists + fires task lifecycle events and reports whether the per-task **entity set** changed; `coordinator.async_settle_buy_tasks` reconciles then does a **deferred** entry reload (guarded, like `problem_sync`) when a device-bearing buy task is created/removed, else a plain refresh. Called from setup, asset CRUD, every stock-adjust surface, and every completion surface.
- **Completion** — restocks at a new `store._stamp_buy_restock` chokepoint, and runs **at most one** stock side-effect per completion (part-consume XOR buy-restock). `add_task` / `delete_task` / `delete_asset` / the one-off purge all treat the `buy` source as system-managed.
- **i18n / docs** — localized `"Buy {part}"` name (16 languages), the panel field labels + per-field helper text in all 16 locales; README section, `docs/EVENTS.md` note (no new event type), `.amazonq/rules` entry, screenshot + walkthrough tour step.
- **Release** — beta bump to `0.8.0b4` (+ CHANGELOG). `preview-release` label applied.

## Testing

- **Unit** (green locally): `reconcile_buy_tasks` create/orphan/idempotent-after-completion branches, part-field normalization + `part_wants_buy_task`, and the buy-task name table (`tests/unit/test_reconcile.py`, `test_assets.py`, `test_wear_task_naming.py`).
- **Integration** — `tests/integration/test_events.py::test_auto_buy_task_lifecycle`: low crossing → reminder appears; completing it → stock bumped by `restock_quantity` and reminder cleared. Polls for the deferred-reload settle (tolerating the transient mid-reload 500), mirroring `test_problem_sync`. Verified end-to-end against a live HA container.
- ruff clean; mypy verified clean on the changed files (full run in CI). Frontend build + vitest green; walkthrough capture succeeds.

## Notes / edge case

If `restock_quantity` is set too low to clear the threshold, the completed reminder stays put (no respawn) until the part is actually restocked above the threshold — documented, by design under "one buy task per low episode".

https://claude.ai/code/session_01BRkwwKi6WPZW5PtHUv79TW


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. This adds a system-managed, level-triggered "auto-buy" task reconciler. Focus on:

- **Correctness of the lifecycle:** the `reconcile.reconcile_buy_tasks` idempotency rule (existing index counts a buy task whether **open or completed**, so completing doesn't respawn while still low). Are there sequences — restock-to-still-below-threshold, toggling `create_buy_task` off while low, a part edit that clears `reorder_at`, deleting the asset vs the part — where a buy task leaks, duplicates, or fails to clear? Consider the `restock_quantity`-too-small edge (completed reminder lingers) — is that the only residual, and is it safe?
- **The settle / reload path:** `coordinator.async_settle_buy_tasks` schedules a **deferred** entry reload (guarded by `_buy_reload_scheduled`) when a device-bearing buy task is created/removed, else refreshes. Any reload pile-up, races, or missed-reload cases across the wired surfaces (setup, `devices.async_apply_asset_change`, `adjust_part_stock` service/number/websocket, `complete_task` service/button/todo)? Is calling it from inside entity handlers (number/button/todo) safe?
- **Completion double-mutation:** `store._stamp_buy_restock` runs alongside `_stamp_part_replacement`. Confirm a buy task (source `buy`, not `part`) can never be both consumed and restocked, and that `restock_quantity` (default/None → 1, negatives rejected at normalize) can't produce a bad crossing.
- **Reserved-source integrity:** `"buy"` is rejected by `add_task`, blocked by `delete_task`, dropped by `delete_asset`, skipped by the one-off purge. Any surface that can still mint, orphan, or strand a `buy`-sourced task?
- **HA best practices / maintainability:** the store→coordinator entity-set decision is duplicated via `_task_owns_entities` (to avoid an import cycle) — acceptable? Naming, module boundaries, and anything that should be a shared helper.
- **Perf:** the extra reconcile pass on every stock/completion mutation — any hot-path or full-reload concern.

---

prestomation:
Status:
- ✅ **ruff** CI failure fixed (`8380904`) — line-length + formatting in the two new test files (my local lint had only covered `custom_components/`). Whole-repo `ruff check` + `ruff format --check` now clean; the reformatted tests still pass.
- ✅ Amazon Q review came back clean — no changes requested.
- ⏳ Re-running CI on the new commit.

---

prestomation:
Thanks — thorough pass. Triage:

**1. Completion double-mutation → fixed (`75d5a35`).** Agreed it's not reachable today (no path mints a both-source task: `add_task` rejects reserved keys, and each reconciler / `set_task_consumable` builds a single-namespace source), but it's cheap to make structural. `complete_task` now runs **at most one** stock side-effect via an explicit branch — `part` source consumes, else `buy` source restocks — so even a hand-corrupted both-sources task can't consume *and* restock in one completion:
```python
if _part_source(updated):
    self._stamp_part_replacement(updated, when)
elif _buy_source(updated):
    self._stamp_buy_restock(updated)
```
I kept it at the completion chokepoint rather than in `build_task`, since storage is loaded verbatim (not re-validated through `build_task`), so the mutation site is the point that actually prevents stock corruption.

**2. Hot-path overhead → accepting + documented (your option 3).** The reconcile is a bounded **in-memory** dict scan (no I/O): `reconcile_buy_tasks` early-returns `False` **before** any `_save()` when nothing changed, which is the common case for a completion/stock-adjust that doesn't cross a buy threshold — so the steady-state cost added to those surfaces is one O(parts + tasks) scan and nothing else. The expensive part (the reload) is already both conditional (only when a device-bearing buy task is created/removed) and guarded against pile-up. For realistic inventories this is negligible and it matches how the wear/problem-sensor reconcilers already work; I'd rather not add debounce/incremental-key complexity for a hypothetical 200-part outlier. Noted the cost in the `.amazonq` rules entry.

**3. Maintainability cross-ref → done (`75d5a35`).** Added a note in `coordinator.task_has_entities` pointing at `store._task_owns_entities` (the inlined duplicate that avoids the store→coordinator import cycle) so they don't diverge.

Your five lifecycle edge cases match my analysis — the `restock_quantity`-too-small case is the only residual (completed reminder lingers, no respawn, cleared on real restock), which is documented.

Also fixed an unrelated CI issue in the same push: the new integration test now polls for the deferred-reload settle (tolerating the transient mid-reload 500), mirroring `test_problem_sync`.

---

prestomation:
Status — all flagged items addressed:
- ✅ **ruff** — fixed (`8380904`).
- ✅ **integration** — the new test read state immediately after `adjust_part_stock`/`complete_task`, which settle via a *deferred* reload; now polls (tolerating the transient mid-reload 500), mirroring `test_problem_sync` (`75d5a35`). Verified against a live container.
- ✅ **Amazon Q review** — double-mutation hardened (single stock side-effect per completion) + maintainability cross-ref (`75d5a35`); hot-path accepted with rationale (in-memory, save-free on no-change).
- ✅ **walkthrough** (soft gate) — the new tour step cancelled the editor and waited on a stale surface; now ends on the appliances tab (`f930582`). Navigation validated locally.
- ℹ️ Coverage delta is the usual store/coordinator/websocket integration-only lines (0% in *unit* coverage by design); the pure logic added is 100% covered.

Re-running CI on `f930582`.

---

PR #135, opened 2026-07-06
Add durable, editable notes to problem-sensor tasks

prestomation:
## What & why

When Home Keeper mirrors a `device_class: problem` binary sensor as a task, that task has **no device to model** — so there was nowhere to jot the instructions you want to remember the *next* time the problem fires (the fix that worked last time, a part number, where the shut-off valve is).

Under the hood a `notes` field already existed on every task, but for a synced problem task it was unreachable: these are source-owned, so the panel suppresses the whole edit dialog. This PR makes notes **editable inline** on the problem-sensor task detail, and — per the requested design — makes them **durable and keyed to the sensor** so they outlive the task itself.

## How it works

**Backend**
- New entity-keyed side-store `store._problem_notes` (`entity_id → note`), persisted in the storage doc alongside `tasks`/`assets` (additive; no migration).
- `update_task` mirrors a synced task's `notes` into it; `reconcile_problem_sensor_tasks` passes it back to the pure reconciler, which re-hydrates `notes` onto a freshly built mirror (`problem_tasks.reconcile_problem_tasks(..., notes_by_entity=...)`). Identity is the sensor `entity_id`, the stable key across firings.
- Net effect: a note survives the mirror clearing → re-arming, and even the task being **deleted and recreated** (sync toggled off/on, sensor temporarily excluded), reappearing next time the same problem fires. `problem_tasks.py` stays HA-free.

**Frontend**
- The Notes card on a problem-sensor task detail gets an inline **Add a note / Edit note** editor (uncontrolled textarea, read on Save), saving via the existing `update_task` websocket path. New `note.*` strings translated across all 15 non-English locales.

## Screenshot

Editing the note on a synced problem-sensor task — a textarea seeded with the note and Save/Cancel buttons

## Testing

- `pytest tests/unit` — 504 passed (adds note-hydration cases to `test_problem_tasks.py`).
- `pytest tests/integration/test_problem_sync.py` — passes against the real HA container; new test asserts a synced task's note is editable via `update_task` and persists.
- `vitest` — 216 passed (translation-parity guards green).
- Panel `tsc --noEmit` clean; `ruff` lint+format clean; targeted `mypy` clean on changed files (full HA-backed `mypy` runs in CI — `homeassistant` couldn't be installed in this environment).
- Captured the editor end-to-end (save → leave → re-open → note shown) and visually verified the screenshot.

## Release / docs

- Beta bump to **0.9.0b1** (`manifest.json` + `const.py`), matching `## [0.9.0b1]` CHANGELOG section.
- README section under "Sync problem sensors" + embedded screenshot.
- Amazon Q architecture rule updated with the notes/durability behavior.
- Screenshot + walkthrough capture harnesses extended to step through the new surface.

https://claude.ai/code/session_0137F8cLmJ9f8v5kXUcyZW9B

---
_Generated by [Claude Code](https://claude.ai/code/session_0137F8cLmJ9f8v5kXUcyZW9B)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first and don't withhold minor ones. Focus on:

- **Correctness / edge cases:** the durable note lifecycle. `store.update_task` mirrors `notes` into the entity-keyed `_problem_notes` side-store, and `reconcile_problem_tasks` re-hydrates it onto a recreated mirror. Are there races or ordering bugs between an in-flight reconcile and a note edit? Is the write-back guard (`_problem_entity(merged) is not None`, drop-on-empty) right? Can a stale side-store value ever clobber a live edited note (I intended hydration to seed *new* tasks only)? Does clearing a note (empty string) behave correctly on both persist and re-hydrate?
- **Persistence / migration:** `problem_notes` is added to the storage doc without a version bump (additive, defaults to `{}`). Any risk to existing installs? Unbounded growth of orphaned notes for entities that no longer exist — worth pruning on registry removal?
- **HA best practices:** editing `notes` on a synced/source-owned task is intentionally allowed (`update_task` isn't gated by `_reject_synced_problem`) while name/schedule stay locked. Is that distinction sound, or should there be an explicit allow-list?
- **Frontend:** the uncontrolled inline textarea (value read on Save), the `_noteEdit` state reset on navigation, and XSS/escaping on the notes render path.
- **Maintainability:** module boundaries (pure `problem_tasks.py` staying HA-free), naming, and duplication.

---

Issue #143, opened 2026-07-17
Feature: set/edit the completion date when marking a task done (back-date from the panel)


sckissel:
**Is your feature request related to a problem?**

When I mark a task **Done** in the panel, it's always stamped "now" — there's no way to say "I actually did this last Tuesday." For floating tasks that silently sets the next-due wrong (a 30-day task I did 5 days ago now shows due 30 days from *today*, not from when I did it). Today the only fixes are the `home_keeper.complete_task` service with `completed_at`, or deleting the completion from history and re-adding via the service — neither is discoverable for a normal user.

**Describe the solution you'd like**

Two small UI affordances backed by the existing engine:

1. **Completion dialog:** an optional **Completed at** date/time field (defaults to now). It just flows into the existing `complete_task(completed_at=…)` — no new backend for the create path.
2. **History list (task detail):** an "edit date" affordance on each completion row, next to the existing trash / edit-metadata buttons.

**Proposed implementation (sketch)**

- The dialog date requires **no backend change** — `store.complete_task` / `apply_completion` already accept `completed_at`.
- Editing an *existing* entry's date is the only new backend bit. `update_completion` intentionally won't move `ts`, so add a thin pure `recurrence.move_completion(task, old_ts, new_ts, *, now)` that removes `old_ts`, inserts `new_ts` (with the same dup-collapse as `apply_completion`), then re-derives `last_completed`/`next_due` exactly like `remove_completion` does (floating rewinds to the new latest; fixed stays schedule-driven; triggered/sensor/one-off armed-state rules unchanged).
- Per the "every data action is a service" convention, expose `home_keeper.move_completion` (+ a websocket command delegating to the same store method) and fire an event (or reuse `task_uncompleted` + `task_completed`). Add unit tests under `tests/unit/test_recurrence_*` and translation-parity strings.

**Alternatives considered**

- Using `home_keeper.complete_task` with `completed_at` (works today, but not user-facing).
- Delete-then-re-add from the history UI (also works, but clunky and still "now"-only on re-add).

**Additional context**

Migrated ~7 recurring tasks from the old *Home Maintenance* integration into Home Keeper and hit this immediately when logging a chore I'd done a couple weeks prior.

prestomation:
Thank you for the request! This is included in 0.9.0. I'm looking forward to your future feedback

PR #146, opened 2026-07-23
Back-date / move a completion's timestamp (#143)

prestomation:
## Summary

Implements #143. Two panel affordances, backed by the recurrence engine:

- The completion dialog gets an optional **Completed at** date/time field
  (defaults to now) for logging a new completion in the past — so a floating
  task's next-due date measures from when the work actually happened.
- Each history row gets a **move date** action (next to edit-metadata and
  delete) to re-timestamp an already-recorded completion, distinct from the
  existing pencil button, which edits note/cost/photo/who but never `ts`.

### Backend

- `recurrence.move_completion(task, old_ts, new_ts, *, now)` — removes the entry
  at `old_ts` and re-inserts it at `new_ts` (collapsing into an existing entry
  there, same as `apply_completion`'s same-instant dedup) **before** a single
  re-derive of `last_completed`/`next_due` from the final history. This ordering
  matters: re-deriving in between (as if this were `remove_completion` followed
  by a separate insert) would misfire for a one-off task's *only* completion —
  it would transiently look like "history now empty" and re-arm `next_due`,
  which would be wrong once the moved entry lands back in.
- `store.move_completion` fires `home_keeper_task_uncompleted` then
  `home_keeper_task_completed` (modeled as undo-then-redo) so existing
  automations/integrations see the move without a new event type to learn.
- New `home_keeper.move_completion` service + `home_keeper/move_completion`
  websocket command, following `update_completion`'s pattern.
- Fixed a latent gap: `ws_complete_task` never forwarded `completed_at` even
  though the service and store already supported back-dating — required for
  the dialog's new field to actually reach the backend.

### Frontend

`api.ts` gains `moveCompletion` + `completeTask`'s `completedAt`; `panel.ts`
wires both the dialog field and the new history-row dialog, reusing
`forms.ts`'s existing `isoToHaDateTime`/`haDateTimeToIso` datetime-selector
helpers.

### Rebase note

Rebased onto `main` after #147 (fixed the completion dialog's invisible Save
button — `ha-dialog` no longer exposes `primaryAction`/`secondaryAction` slots
directly) and #148 landed. My new move-date dialog was written before #147 and
had the identical bug; applied the same `<ha-dialog-footer>` fix and added a
matching e2e regression test. All screenshots below are regenerated against the
merged, fixed code.

### Tests

- `tests/unit/test_move_completion.py` — per-recurrence-type re-derive
  (floating/fixed/triggered/sensor/one-off), the metadata-collision rule,
  naive-timestamp qualification, and the one-off-only-completion ordering case.
- 4 new integration tests (service + websocket paths, both events firing,
  unknown-`ts` rejection).
- `completion-dialog.spec.ts` — new e2e regression test for the move-date
  dialog's button structure, mirroring #147's test for the completion dialog.
- Frontend vitest coverage for the new API calls.

### Localization

`move_completion` service strings + 3 new panel i18n keys, translated across
all 16 supported locales (not just seeded from English — verified against the
untranslated-string guard).

### Docs

`docs/BACKDATE_COMPLETION_PLAN.md` marked implemented, README section on
back-dating/moving with a screenshot, `CHANGELOG.md` `0.9.0b2` entry, version
bump (`manifest.json` + `const.PANEL_VERSION` → `0.9.0b2`).

## Screenshots

Completion dialog with the new Completed at date/time field and visible action buttons

Move-date dialog on a history row, prefilled with the completion's current timestamp, with visible Save/Cancel buttons

## Test plan

- [x] `pytest tests/unit -v` — 515 passed
- [x] `pytest` full integration suite (Docker HA) — 81 passed, including 4 new
      `move_completion`/`ws_complete_task` tests
- [x] `mypy custom_components/home_keeper` — clean
- [x] `ruff check` / `ruff format --check` — clean
- [x] Full `vitest` suite — 220 passed
- [x] `tsc --noEmit` + `rollup -c` — clean build
- [x] Full local e2e suite (Docker HA + Playwright) — 39 passed, including the
      new move-date dialog regression test
- [x] Manual verification: captured real screenshots and the CI walkthrough
      video locally and visually inspected both new surfaces (buttons visible,
      correct prefilled values)


prestomation:
/q review This is a plan-only PR (a new `docs/BACKDATE_COMPLETION_PLAN.md`, no functional code changes) for issue #143. Please give critical, skeptical feedback on the **design itself**, not just prose quality:

- **Correctness of the recurrence model**: does re-deriving `last_completed`/`next_due` for `move_completion` the same way `remove_completion` does (rather than trusting the moved timestamp like `apply_completion` does) actually cover all recurrence types correctly, especially one-off (re-arm-if-empty) and triggered/sensor (untouched)? Any edge case the plan misses (e.g. moving a completion onto a duplicate `new_ts`, or moving the *only* completion of a one-off)?
- **Event-reuse decision**: is firing `task_uncompleted` + `task_completed` for a move (instead of a dedicated event) going to confuse any existing automation/integration semantics, e.g. around `origin`/metadata echoing?
- **Architecture/maintainability**: does introducing a third completion-editing primitive (`move_completion`, alongside `update_completion` and `delete_completion`) create any confusing overlap or naming ambiguity for API consumers?
- Don't withhold minor issues — list everything you find, most serious first.


---

prestomation:
Thanks for the thorough review — the ordering bug was real and worth catching before an implementation PR builds on it. Pushed `1d107bd` addressing it:

- **One-off re-arm ordering (the critical one):** confirmed the doc was ambiguous in a way that would produce the bug you described. Clarified that `move_completion` must remove *and* re-insert the moved entry **before** the single re-derive pass — not "run `remove_completion`'s logic, then insert afterward" — so a moved one-off's only completion correctly leaves `next_due = None` instead of re-arming to `due`.
- **`new_ts` collision metadata:** documented explicitly — the moved entry's metadata wins, the entry it collides with is discarded.
- **Sensor/triggered non-validation:** called out explicitly as an accepted, unvalidated edge case (consistent with `update_completion`/`delete_completion` today, which don't schedule-validate manual edits either).
- **Naive `new_ts`:** clarified it's live code, not dead — the panel's websocket path always sends an aware ISO string, but the `move_completion` **service**'s `new_completed_at` is `cv.datetime`, which accepts an offset-less value the same way `complete_task`'s existing `completed_at` does.
- **`ws_complete_task` gap:** noted explicitly as a pre-existing latent gap in already-shipped functionality, not something this feature introduces — flagged that it's fine to land as its own small fix if a reviewer prefers it decoupled. Added an integration-test bullet for it specifically.
- **Event-reuse and `move_completion`-vs-`update_completion` naming:** both are deliberate product decisions from this plan's design phase (not oversights), so I documented them as explicit trade-offs with your counter-arguments captured rather than reversing them unilaterally — a dedicated event type and a unified `edit_completion(...)` remain on the table if either proves confusing once implemented.

Separately: `e2e`/`integration`/`test`/`walkthrough` are all failing from the same pre-existing cause (`@rollup/plugin-typescript@12.3.0` incompatible with the `typescript@7.0.2` bump merged in #139/#140) — reproduces with zero frontend files changed, so it's not from this PR. Not fixing it here since it's out of scope for a docs-only plan PR and affects `main`/every PR right now; will re-run once that's addressed separately.


---

prestomation:
Pushed `a04a362` fixing the pre-existing CI break (per request): reverted the `typescript` 6.0.3 → 7.0.2 dependabot bump from #139. `@rollup/plugin-typescript@12.3.0` (the latest release — nothing newer exists yet) throws `TypeError: Cannot read properties of undefined (reading 'ES2015')` against TS 7's internals, which took down `ci/build-panel.sh` and everything downstream of it (`e2e`, `integration`, `test`, `walkthrough`).

Verified locally before pushing:
- `rollup -c` builds both bundles cleanly
- `tsc --noEmit` — clean
- Full frontend vitest suite — 216/216 passing
- Python unit suite — 504 passed, 1 skipped

This is a repo-wide fix (affects `main` and every other PR right now, not just this one) rather than anything specific to the completion-timestamp plan — bundled into this PR since that's where it was reported, but happy to split it out if preferred.


---

prestomation:
/q review This PR now contains the full implementation (previously it was plan-only). Please give critical, skeptical feedback, focused on:

- **Correctness**: does `recurrence.move_completion` (custom_components/home_keeper/recurrence.py) actually get the re-derive ordering right for every recurrence type, especially the one-off-only-completion case the docstring calls out? Any edge case in the metadata-collision or naive-timestamp handling I've missed?
- **Store/event layer**: `store.move_completion` fires `task_uncompleted` then `task_completed` — check the `origin`/metadata echoed on the synthetic `task_completed` is right, and that `_reject_synced_problem` guarding is correctly applied.
- **Websocket/service layer**: the `ws_complete_task` fix (now forwards `completed_at`) — check I haven't introduced a validation gap (e.g. malformed date strings).
- **Frontend**: `panel.ts`'s new completion-dialog field and the new move-date dialog — check the `editing` gating logic (the date field must never show up in edit-metadata mode) and that `_wireHistoryDeletes`'s new `.hk-hist-move` wiring doesn't leak listeners on re-render.
- **Test coverage**: are there gaps in `tests/unit/test_move_completion.py` or the new integration tests?

Surface the most serious issues first, and don't withhold minor ones.


---

prestomation:
Thanks for the approval. The coverage report actually caught something real before I saw it: `move_completion`'s "history empty" branches (`last_completed = None`, and the one-off re-arm ternary) were unreachable dead code — copied from `remove_completion`'s pattern, but `move_completion` always removes-then-reinserts exactly one entry, so history can never be empty at that point (unlike `remove_completion`, which can genuinely empty it). Pushed `ff6e39f` simplifying both to their always-true outcome; all 515 unit tests (including the one-off ordering case) still pass, plus clean ruff/mypy.


---

PR #141, opened 2026-07-11
Link a spare part to problem-sensor tasks (inventory + where-to-buy)

prestomation:
## What & why

A problem sensor tells you *something* needs attention (e.g. an air filter reporting it's unhealthy) — Home Keeper mirrors it as a task — but on its own it doesn't tell you **which part fixes it or where to get one**. This reconciles the **problem-sensor sync** with the **parts/inventory** feature: a synced problem task can now be linked to a spare consumable part, so the task surfaces **where to buy it** (vendor + product URL) and **how many spares are on hand**, and can optionally **draw down stock** when the problem clears.

## The collision it resolves

A task holds exactly **one `source`**, and a problem mirror's slot is owned by the sync (and the task is **completion-blocked** — no "Done"). So the existing `set_task_consumable` / `source.part` link can't be reused. Instead the link rides a **decoupled, user-editable `task["consumable"]` field**, mirroring the durable-notes pattern from #135 — persisted in an entity-keyed side-store so it survives the mirror being cleared, re-armed, or recreated.

Consumption can't happen on completion (there is none), so `consume_on_clear` (**`off`** default / `auto`) governs the reconcile **`cleared`** branch: `auto` draws down one spare via the existing `consume_part_stock` and fires the same low/out-of-stock events. Default `off` keeps the link informational — a sensor can clear because it was reset/cleaned, not replaced.

## Screenshots

The **Linked part** card on a synced problem task — where to buy it and spares on hand, with whether clearing draws down a spare:

Linked part card on a problem-sensor task detail

The inline attach editor — part picker plus the use-a-spare-when-it-clears toggle:

Inline linked-part editor with part picker and consume-on-clear toggle

## Changes

**Backend**
- `models`: `consumable` (`{asset_id, part_id}`) + `consume_on_clear` (`off`/`auto`) fields with normalizers; defaults in `build_task`; merged like `card_links`.
- `store`: durable, entity-keyed `_problem_consumables` side-store (persisted with tasks/assets); `update_task` mirrors + validates the link; reconcile `cleared` auto-consumes when `auto`; `_prune_dangling_consumables` clears links whose part was removed.
- `problem_tasks`: threads `consumables_by_entity` so a rebuilt mirror re-hydrates the link.
- `const` / `services.yaml` / `UPDATE_TASK_SCHEMA` additions.

**Frontend**
[...]

PR #147, opened 2026-07-23
Fix invisible Save button in the task-completion dialog

### LINKED ISSUE #144: No save button shown when marking tast as completed
When i want to mark a task as done there is no save button shown either in browser or mobile app.

Image

prestomation:
## What & why

Fixes #144. A task set to "Ask for details" or "Require details" on completion opens a dialog to log a note/cost/who when you tap **Done** — but its **Mark done**/**Save**, **Skip details**, and **Cancel** buttons never rendered, in the panel or the mobile app, leaving no way to confirm or dismiss it.

## Root cause

Confirmed by driving the actual Home Assistant frontend in a browser: current `ha-dialog` is backed by `wa-dialog` (Web Awesome) and no longer exposes `primaryAction`/`secondaryAction` slots directly — its shadow DOM only defines `header`, `headerNavigationIcon`, `headerTitle`, `headerSubtitle`, `headerActionItems`, and **`footer`**. The completion dialog (`_renderCompletionDialog` in `panel.ts`) slotted its buttons straight onto `<ha-dialog>` via `slot="primaryAction"`/`slot="secondaryAction"` — the old mwc-dialog convention. Per the Shadow DOM spec, an element assigned to a slot name the host doesn't define simply never renders, which is exactly the symptom (button present in the DOM, zero-size, never visible).

Compared against Home Assistant's own native dialogs (e.g. the "Create area" dialog) to find the current correct pattern: action buttons must be wrapped in `<ha-dialog-footer slot="footer">`, with the buttons keeping their `slot="primaryAction"`/`slot="secondaryAction"` *inside* that wrapper. `ha-dialog-footer` is a globally registered HA frontend element, used the same way the panel already feature-detects `ha-picture-upload`.

This is the *only* dialog in the panel affected — every other confirmation dialog (delete, orphan cleanup, etc.) uses a hand-rolled `ha-card` overlay rather than native `ha-dialog`.

**This was never a bug in the original code — it's an upstream regression.** `_renderCompletionDialog` was written once and never touched since; the `primaryAction`/`secondaryAction`-directly-on-`<ha-dialog>` pattern is the real, correct API for `mwc-dialog`, which is what HA's `ha-dialog` used to wrap. Tracked down the exact break:

- [home-assistant/frontend#29627](https://github.com/home-assistant/frontend/pull/29627) — "Delete ha-dialog, rename ha-wa-dialog to ha-dialog," merged **2026-02-15**, completing the [dialog-to-webawesome migration](https://github.com/home-assistant/frontend/issues/27143). This is the commit that made the shared `ha-dialog` component wrap `wa-dialog` instead of `mwc-dialog`.
- **HA Core 2026.2.0** pins frontend `20260128.6` (cut before the merge) — still the old `ha-dialog`, dialog worked fine.
- **HA Core 2026.3.0** pins frontend `20260304.0` (cut after the merge) — first stable release with the new `wa-dialog`-backed `ha-dialog`.

So the completion dialog worked correctly through HA **2026.2.x**, and silently broke for anyone on **2026.3.0 or later**, with zero code change on Home Keeper's side.

## Fix

Wrap the completion dialog's primary/skip/cancel buttons in `<ha-dialog-footer slot="footer">` (`panel.ts`, `_renderCompletionDialog`), falling back to the old direct-slot pattern when `ha-dialog-footer` isn't registered (e.g. on HA versions predating this migration), so older installs keep working too.

## Screenshot

The completion dialog with its action buttons now visible (**Skip details**, **Cancel**, **Mark done**):

Completion dialog showing Skip details, Cancel, and Mark done buttons

## Testing

- Reproduced live against the real HA Docker container (`ci/e2e-up.sh`) with a throwaway Playwright script that inspected the dialog's shadow DOM directly — confirmed the button was present but had zero size/visibility before the fix, and renders correctly after.
- Added `tests/e2e/tests/completion-dialog.spec.ts` — opens the completion dialog, asserts the buttons are structurally inside `<ha-dialog-footer slot="footer">` and visible, then submits it. Verified this test **fails** against the pre-fix code (`toBeVisible()` times out) and **passes** with the fix.
- `npm test` (root vitest) — 216 passed.
- `tsc --noEmit` — clean.
- Regenerated `docs/images/11-panel-completion-dialog.png` and visually confirmed it shows the dialog with all three buttons.
- Full local e2e suite via `ci/e2e-up.sh` — 38/38 Playwright specs pass.

No beta bump — this is a bug-fix-only change per `AGENTS.md`.

## Also fixed: unrelated pre-existing CI break

Pinned `typescript` back to `^6.0.3` in the frontend `package.json`/`package-lock.json` — `typescript@7.0.2` (bumped in #139/#140) is incompatible with `@rollup/plugin-typescript@12.3.0` (`ts.ModuleKind` is undefined under TS 7), which was breaking `ci/build-panel.sh` and, with it, every CI job that builds the panel (already flagged in #141). No newer `@rollup/plugin-typescript` release supports TS 7 yet, so pinning back is the fix until it does. Developer-tooling only, no user-facing change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VmSzbsVDTdLzH5Nb6qq321)_


prestomation:
/q review Please give critical, skeptical feedback on this fix. In particular:

- **Correctness**: is wrapping the buttons in `<ha-dialog-footer slot="footer">` actually the right/complete fix for the current HA `ha-dialog`/`wa-dialog` API, or are there edge cases (e.g. older HA frontend versions still in the wild, `ha-dialog-footer` not being registered in some contexts) where this could regress instead of fix?
- **Completeness**: is there any other dialog or slotted-button pattern elsewhere in `panel.ts` (or `card.ts`) that might have the same `primaryAction`/`secondaryAction`-on-`<ha-dialog>` issue and was missed?
- **Test coverage**: does the new `tests/e2e/tests/completion-dialog.spec.ts` actually guard against a regression, or is it too narrowly coupled to the current DOM structure to catch a future HA frontend API change?
- **HA best practices**: any concerns with relying on `ha-dialog-footer` as an internal/undocumented HA frontend component from a custom integration's panel?

Please surface the most serious issues first, and don't hold back minor ones either.

---

prestomation:
The `e2e` and `integration` check failures are a pre-existing, unrelated toolchain break — not caused by this PR. Both fail identically, before any of my code runs:

```
[!] TypeError: Cannot read properties of undefined (reading 'ES2015')
    at .../@rollup/plugin-typescript/dist/es/index.js:528:16
```

This is `bash ci/build-panel.sh` failing during `rollup -c`, caused by the `typescript` 7.0.2 bump in #139/#140 — `@rollup/plugin-typescript@12.3.0` isn't compatible with it (`ts.ModuleKind` is undefined under TS 7). It reproduces on `main` as-is (base commit `7772a0f`), independent of this branch's diff, and was already flagged in #141. Not fixing it here since it's an unrelated dependency issue outside this bug fix's scope — happy to open a separate PR to pin/upgrade if wanted.

`walkthrough` fails for the same reason (it also needs `ci/build-panel.sh`) — it's a soft gate per `AGENTS.md` and doesn't block merge.

I confirmed the actual fix locally against a real HA container: `npm test` (216 passed), `tsc --noEmit` clean, and the new `tests/e2e/tests/completion-dialog.spec.ts` passes with the fix / fails without it (verified both ways).

---

prestomation:
Thanks for the detailed review. Addressed the two substantive points:

- **Feature detection / older-HA fallback**: added. `_renderCompletionDialog` now checks `customElements.get('ha-dialog-footer')` and falls back to slotting the buttons directly on `<ha-dialog>` (the pre-`wa-dialog` convention) when it's absent, so an older HA frontend gets a working dialog either way instead of a silent regression.
- **Test coverage**: strengthened. The e2e test now asserts the buttons are structurally inside `ha-dialog-footer[slot="footer"]` (exact count of `ha-button[slot="primaryAction"]` / `[slot="secondaryAction"]`), not just visible — a future revert to the old direct-slot pattern fails immediately rather than only when someone notices the UI.

On the other two points:

- **CHANGELOG entry**: this was already present — see the `### Fixed` entry under `## [0.9.0b1]` in the original diff. Not sure why it read as missing; happy to point at the exact lines if useful.
- **15s timeout / button-order nitpick**: left as-is. The 15s wait matches the existing convention in `screenshots.capture.ts` for the same dialog (`ha-dialog[open] ... waitFor({ timeout: 15_000 })`), and the button creation order (primary → optional skip → cancel) mirrors the existing code structure — reordering purely for readability didn't seem worth the diff noise here.

Also confirmed: only this one dialog in the panel uses native `ha-dialog` with action-button slots (as you verified) — everything else is a hand-rolled overlay, so no other surface is affected.

---

prestomation:
The `e2e`/`integration`/`test`/`walkthrough` failures on this new commit are the same pre-existing `ci/build-panel.sh` break described above (unrelated to the review follow-up commit) — no new failure signature. Not re-explaining per push; will flag here again only if something actually changes.

Pushed a follow-up addressing the two actionable review points:
- Falls back to slotting buttons directly on `<ha-dialog>` when `ha-dialog-footer` isn't registered, so older HA frontends aren't regressed.
- The e2e test now asserts the buttons are structurally inside `<ha-dialog-footer slot="footer">`, not just visible.

Verified locally against the real HA container: `npm test` (216 passed), `tsc --noEmit` clean, updated e2e test passes.

---

prestomation:
Fixed the pre-existing CI-blocking toolchain break too (the `e2e`/`integration`/`test`/`walkthrough` failures discussed above): pinned `typescript` back to `^6.0.3` in `custom_components/home_keeper/frontend/package.json`/`package-lock.json`. `typescript@7.0.2` (bumped in #139/#140) is incompatible with `@rollup/plugin-typescript@12.3.0` (`ts.ModuleKind` is undefined under TS 7) — no newer plugin release exists yet, so pinning back is the fix until it does.

This is a separate, unrelated bug from #144 (developer-tooling only, no user-facing change, no CHANGELOG entry needed per `AGENTS.md`), but it was blocking every CI check on this PR so I fixed it here rather than opening a second PR.

Verified clean: `npm ci && npm run build`, `tsc --noEmit`, `npm test` (216 passed), and the **full** local e2e suite via `ci/e2e-up.sh` — all 38 Playwright specs pass, including the new completion-dialog regression test.

---

prestomation:
CI is fully green now — the `typescript` pin fixed `e2e`, `integration`, `test`, and `walkthrough` (all previously blocked by the pre-existing rollup build break). All 13 checks pass: `ruff`, `mypy`, `hacs`, `pytest and coverage report`, `preview`, `release`, `integration`, `e2e`, `test`, `walkthrough` ✅.

This PR is ready for review.

---

PR #165, opened 2026-07-30
Open uploaded documents by a native anchor tap (Fixes #164)

### LINKED ISSUE #164: File links not clickable on mobile
image

In device info - tapping then does nothing.

Also on desktop the UX is pretty bad - there's no cursor change or reaction to hover to suggest it's clickable

prestomation:
Fixes #164.

## The bug

An uploaded manual/receipt on an appliance's page couldn't be opened at all in the Home Assistant companion app, and gave no hover or cursor feedback on desktop.

Both symptoms came from one line of markup: a **file** document rendered as an `&lt;a role="button" tabindex="0"&gt;` with **no href**, whose click handler signed the URL and *then* called `window.open`.

1. **Mobile** — the iOS app's WKWebView blocks a `window.open` issued after the user gesture has been spent on an async round-trip, so the tap did nothing at all.
2. **Desktop** — `.hk-doc-file` had no CSS whatsoever, and an anchor without an `href` gets no default link affordance. It inherited the link colour and nothing else, so it read as dead text. (The neighbouring part paperclip already had `cursor: pointer`, which is why the report singled out documents.)

The dashboard card hit this same WKWebView problem earlier and solved it by pre-signing into a real `href`; the panel never got the same treatment.

## The fix

Every openable file is now a real anchor whose `href` is minted **ahead of** the click.

- **`SignedUrlCache` (`documents.ts`)** holds the short-lived URLs — reused until stale, re-signed at 45 min against the backend's 1 h TTL, evicted once the surface stops showing the file, and overlapping callers join one round-trip per file. The card's bespoke copy of that logic is gone, so the panel and the card now share one implementation.
- **One signing chokepoint in the panel** (`_signFiles`, run after every render) collects the current screen's files and stamps each href onto the anchor waiting for it, matched by a `data-sign` cache key. The appliance detail page, both edit-form "Open" actions, and a part's paperclip all go through it. A timer (`_armResign`) re-runs the pass so a page left open — a wall tablet, a forgotten tab — can't end up with an href that outlived the TTL; the panel renders on navigation and `set hass` deliberately doesn't re-render, so nothing else would have triggered it.
- **`openDocument` / `openPartFile` survive only as the fallback** for the window before the first sign lands (and for a sign that fails outright), and stand down as soon as the anchor has an href — so a click can never both navigate *and* open a second tab.
- **The editor's "Open" is the anchor itself carrying the icon**, rather than an `ha-icon-button` nested inside one. Whether a click on a nested interactive control reaches the enclosing link is browser-dependent, which is exactly the class of bug being fixed here.

Affordances restored along the way: pointer cursor and hover underline, a trailing open-in-new hint, long-press "open in new tab", middle-click, keyboard activation, and a 44px touch target (WCAG 2.5.5) on document rows and the paperclip — the latter was previously a bare 16px icon.

## Screenshots

Appliance detail — **Manuals & documents**. Both kinds now render identically as underlined links with an "opens in a new tab" hint, in comfortably tappable rows. Previously "Installation guide (PDF)" (the uploaded file) was un-underlined, un-hoverable text that did nothing on mobile. (Captured after rebasing onto #167, so the Markdown Notes card and part notes are visible too.)

Appliance detail page showing the Manuals &amp; documents section with both documents as links

Documents editor — the "Open" action is now an anchor styled to sit flush with the Edit/Remove icon buttons.

Appliance edit form showing the documents editor

A part's attached file in the editor, same treatment.

Part editor showing an attached file card with Open and Remove actions

## Tests

- **Unit (vitest)** — new `documents.test.js` covers the cache (sign once and reuse, join an in-flight sign, re-sign near the TTL, evict what's no longer shown, keep the prior URL when a re-sign fails, no round-trip for an empty set) and `assetFileRefs`/`signedFileKey`. New panel cases assert the detail page's markup: a signed href on the file document and the part paperclip, a plain href for a link document, no `role="button"`, the fallback firing only while the href is missing, no double-open once it's there, and a fake-timer case proving a long-open page re-signs and re-stamps its anchors.
- **e2e (Playwright)** — new `documents.spec.ts` asserts every document and part file carries a signed href, that the edit form's Open action does too, that a tap opens a tab, and that the signature is one the backend actually honours (200 + `application/pdf`).
- Full suites green locally after the rebase: 295 vitest, 42 e2e, 677 pytest unit.

The e2e fixture gained the two document blobs its `.storage` metadata already claimed existed, so a signed URL now serves a real PDF end to end (the `.gitignore` there excluded runtime blobs; it now makes an exception for the seeded ones). Their recorded sizes were corrected to match the actual files.

## Notes

- Rebased onto `main` after #167 landed. The CHANGELOG entry folds into the still-unreleased `## [0.10.0b2]` section rather than `[Unreleased]`, since `manifest.json`/`const.py` are already on `0.10.0b2` — **no version bump and no release cut here**.
- `.amazonq/rules/architecture-and-code.md` gains the convention this establishes: **a file is opened by a native anchor tap, never by a JS `window.open` after an `await`** — the old rule said files are "signed on click", which is precisely what broke.
- No tour change (`walkthrough.capture.ts` untouched): no new UI surface, just a fix to an existing one.

https://claude.ai/code/session_01QabtQeEz4S4SewcyeioviM


prestomation:
/q review Be critical and skeptical — assume this is wrong until proven otherwise, and surface the most serious problems first without withholding the minor ones.

This replaces "sign the URL on click, then `window.open`" with "pre-sign into a native `<a href>`" for every uploaded file surface in the panel, and factors the card's existing pre-signing into a shared `SignedUrlCache` (`frontend/src/documents.ts`).

**Correctness / races.** `_signFiles` runs after *every* panel render and both mints URLs and stamps hrefs onto the DOM. Can a re-render that lands mid-flight leave an anchor permanently href-less (i.e. a tap that does nothing again)? Are concurrent `_signFiles` calls safe given `ensure()` prunes anything not in the passed ref set — can two overlapping calls evict each other's entries? Is the 45-minute re-sign margin against the backend's 1h `_DOCUMENT_URL_TTL` actually safe for a panel left open overnight, and does a re-sign refresh hrefs already in the DOM? What happens to an anchor whose signing failed outright — is it still activatable, and is the click fallback's `if (el.getAttribute('href')) return` guard correct in every path (double-open, or no open at all)?

**Security.** Signed URLs are now sitting in the DOM as hrefs rather than being minted per click — is that a meaningful exposure change (referrer leakage, `rel` correctness, URLs surviving in history/`_blank` tabs)? I set the raw URL via the `href` *property* in the DOM path but `safeHref` (which HTML-escapes) in the template path — check I haven't got that backwards anywhere, or introduced an injection or double-escaping bug. Is `data-sign` (a `kind:assetId:id` cache key) safe to put in markup, and can a crafted id break out of the attribute?

**HA best practices.** Anything wrong with the anchor markup for HA's frontend conventions — `ha-svg-icon` used directly inside an anchor instead of `ha-icon-button`, `target="_blank"` behaviour inside the companion app, RTL/theming of the new CSS, and whether the 44px touch targets break any existing layout.

**Maintainability.** Is `SignedUrlCache`'s "call `ensure` with the complete set or it evicts the rest" contract too easy to misuse from a future third surface? Should the panel re-render on a sign instead of mutating hrefs in place? Is the fallback path (`openDocument`/`openPartFile`) now dead weight that should be deleted outright rather than kept as a half-used escape hatch?

**Performance.** `_signFiles` + a `querySelectorAll('a[data-sign]')` on every render — is that a hot path worth guarding?

Also worth a look: the e2e fixture now ships two small PDF blobs under `tests/integration/ha_config/home_keeper/` with a `.gitignore` exception, so a signed URL serves a real file end to end. Is that a reasonable fixture change or an anti-pattern?

---

prestomation:
Thanks — two of these are real and are fixed in ceb4f15. Four don't hold up against the code; details below so the reasoning is on the record.

## Fixed

**Duplicate signs from overlapping `ensure()` calls.** Real, though not for the reason given: the suggested `pending` map is local to a single `ensure()` call, and *within* one call the keys are already deduped by the `needed` map — so there is exactly one task per key and the patch would be a no-op. The genuine gap is *across* calls: `_signFiles` runs after every render, so a second `ensure()` routinely starts before the first resolves and each signs the same file. The in-flight map is now on the instance, so overlapping callers join one round-trip. Covered by a new test that runs two `ensure()` calls concurrently and asserts a single `sign_document_url`.

**A long-lived panel could click into a 403.** Real, and the most valuable finding here — though the stated mechanism is wrong. `_applySignedHrefs` is called *after* `await ensure(...)` inside `_signFiles`, and it mutates the live DOM (`setAttribute`), so a re-sign that lands after render does update the anchors. The actual hole is that nothing *triggers* a re-sign: the panel renders on navigation, and `set hass` (panel.ts:769) deliberately doesn't re-render — so an appliance page left open would pass the 1h TTL with a stale href and no fallback, since the fallback stands down once an href is present. A timer now re-runs the signing pass on the cache's own re-mint clock, disarmed when the screen has no files and on unmount. New test advances fake timers past the TTL and asserts both a fresh mint and a fresh `href` on the anchor; it fails without the fix.

## Not changed

**Eviction race corrupting the cache.** The causal chain runs the other way. `ensure([doc3])` evicts `doc1`/`doc2` *before* render A's signs resolve — so A's writes land *after* the eviction and persist. The failure mode isn't "anchors stay href-less permanently", it's a stale entry lingering one cycle until the next `ensure` prunes it. And when the user navigates back, that render's `ensure` re-signs anything missing and `_documentsSection` emits the href inline from the cache. With the in-flight dedup above, the duplicate-sign path this was really pointing at is gone.

**Keyboard Enter double-opens.** The href guard isn't on the click listener — it's the first line of the shared `run()` that *both* the click and keydown listeners call (panel.ts, `_wireDetailActions`). An href-bearing anchor activated by Enter returns from `run()` before `preventDefault()` or `openDocument()`, so the native activation is the only thing that happens. There's a test asserting the click path isn't `defaultPrevented` once signed.

**`javascript:` URL executing via the property path.** Escaping isn't the security control in either path — the scheme check is, and both apply it. `safeHref` is `isHttpUrl(url) ? escapeHTML(url) : ''` (utils.ts:29), and the property path guards with the same `isHttpUrl` before assigning. `javascript:alert(1)` fails `/^https?:\/\//i` and no href is set. The escaping difference is deliberate and the reason the paths differ: `escapeHTML` is required when interpolating into a template string and *wrong* when assigning to `.href`, where it would double-escape the `&` in a signed URL's query string.

**`<ha-icon>` nested in `<a>` is browser-dependent.** `ha-icon` is a display element with no interactive content in its shadow root — the same nesting the card already uses for `ha-assist-chip` inside `a.hk-task-chip-link`. The case I did avoid is `ha-icon-button`, which *does* wrap a real `<button>`; that's why the editor's Open action carries the SVG path itself rather than wrapping the icon-button. Belt-and-braces anyway: the decorative glyphs now get `pointer-events: none` and `aria-hidden`, so a tap can only land on the anchor and screen readers don't announce them next to the name the anchor already has.

Full suites green locally after the change: 236 vitest, 42 e2e, 677 pytest unit.

---

prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but verify each finding against the actual code paths before reporting it (the previous round had four findings that didn't survive a read of the source). Focus on:

**Correctness / concurrency** — `SignedUrlCache.ensure()` in `custom_components/home_keeper/frontend/src/documents.ts`: the new in-flight promise map, its interaction with eviction (a pending sign resolving after its key was evicted), the `signed` return flag under joined calls, and whether the 45min re-sign clock can ever leave an href live past the backend's 1h TTL. Then the panel's `_signFiles` / `_armResign` / `_applySignedHrefs` trio: timer lifecycle across navigation, re-render mid-sign, and unmount; whether hrefs can be stamped onto detached nodes; whether the timer can leak or fire after `disconnectedCallback`.

**Correctness / UI** — the fallback handlers in `_wireDetailActions`: any path where a click or keypress both navigates natively *and* runs the JS fallback, or where an anchor ends up permanently inert (signing failed and nothing retries).

**Security** — URL handling across the two paths (template `safeHref` vs property assignment with `isHttpUrl`), and whether a signed URL can leak via referrer or be stamped onto the wrong anchor (key collisions between the `document:` and `part:` namespaces).

**Maintainability** — is the cache the right seam between the panel and the card, or does `_signFiles`' "collect every ref the screen shows" contract invite a future caller to pass a partial set and silently evict?

**HA best practices** — anchor semantics, focus/tab order, and touch-target sizing on the panel's detail rows.

Don't withhold minor issues; just mark them as such.

---

PR #166, opened 2026-07-30
Upload UX: visible failures, progress bar, streamed to disk, 100 MB limit (Fixes #159)

### LINKED ISSUE #159: No indication of upload failure
If an uploaded file exceeds 25MB, the request fails. But UI has no error indication that it failed.

Image

prestomation:
Fixes #159.

Rebased onto `main` (which picked up the Markdown-notes and document-anchor work). **No release in this PR**: `manifest.json` and `const.py` stay at `0.10.0b2` exactly as main has them, and the CHANGELOG entries sit under `## [Unreleased]`.

## The bug

An upload that failed looked like nothing happened. The error *was* rendered — but only in the appliance form's error banner, which sits below the identity form, documents editor, metadata editor, parts editor and the related-devices selector. That's hundreds of pixels beneath the **Upload file** button that triggered it, so the reporter only found it by scrolling.

Two things worth stating, since the issue raised both:

- **The size ceiling is ours, not a proxy's.** `const.py` `MAX_DOCUMENT_BYTES`, enforced while streaming in `manuals.py` and returned as a localized JSON 413. We even *raise* aiohttp's cap to reach it (HA's own default is 16 MB). The reporter cited "25MB" — a number no other layer's error text contains — so they hit our limit and read our message.
- **The proxy case is separate and already worked.** `api.ts` treats a non-JSON error body as "something in front of HA rejected this" and shows different guidance. Both branches just happened to land in the same buried alert.

## What changed

**Failures report themselves where the user is looking.** The error now renders inline, next to the control that failed, *and* fires HA's `hass-notification` toast (viewport-fixed, so it can't scroll out of sight). The panel scrolls the error into view, and a stale error is cleared when the next upload succeeds. Every failure branch — our 413, a proxy 413, a connection cut mid-body, a plain network drop — gets the same treatment; only the wording differs, and the proxy branches add a **Learn more** link.

An oversized file rejected — the error renders directly under the Upload file button, and again as a Home Assistant notification toast

**An oversized file is refused before it is uploaded.** The panel checks `file.size` up front, so an over-limit pick fails instantly naming the file, its size and the limit — instead of transferring in full only to be rejected at the end. `frontend/src/limits.ts` mirrors `MAX_DOCUMENT_BYTES` and `tests/unit/test_upload_limit_parity.py` fails the build if the two drift.

**Uploads show progress.** Percentage plus byte count, a **Cancel upload** button, and a "Saving…" phase for the stretch after the last byte while HA stores the file. The upload and **Save** buttons lock while an upload is in flight — that also closes a pre-existing race where Save could PUT the client draft over the asset the upload response was about to rewrite.

An upload in progress — a determinate progress bar, the label 'water-heater-manual.pdf · 8% · 160 KB of 2 MB', a Cancel upload button, and the upload button disabled and relabelled Uploading…

**Uploads stream to disk, and the ceiling is now 100 MB.** The old 25 MB limit wasn't arbitrary: `_parse_upload` accumulated the whole file in a `bytearray` and then copied it with `bytes()`, so accepting a file cost roughly **twice its size in RAM** inside the HA process. At 100 MB that's ~200 MB per concurrent upload — enough to OOM a Pi or a 1 GB install, which is a far worse outcome than a refused upload. So the number moved only after the memory behaviour did:

- each multipart file part is written to a temp file under `<documents root>/.incoming/` as it arrives, flushing at most 1 MB at a time, then moved into place with an **atomic same-filesystem rename**;
- peak memory is a fixed ~1 MB regardless of file size, so the ceiling is now a policy choice rather than a memory budget;
- validation follows the bytes — `documents.validate_upload_stream` checks `(header, size)` and keeps only the 16 bytes the magic-byte sniff needs. `validate_upload` stays as a thin in-memory wrapper, so existing callers and unit tests are unaffected;
- temp-file ownership is explicit: `_parse_upload` leaves nothing behind on any early return, both POST handlers wrap their body in `try/finally` with `async_discard_upload` (a no-op after the move), and setup reclaims strays with an **age-based** purge (not a wipe — setup re-runs on a config-entry reload, and clearing the directory would delete a live upload).

Downloads already streamed via `web.FileResponse` and are untouched.

Both upload flows (appliance documents and part files) go through one shared runner on the panel side, replacing two copies of the error handling.

## Two implementation notes

**Why `XMLHttpRequest` instead of `fetch`.** The Fetch spec exposes progress only for the *response* body — a `fetch` upload is a black box until it resolves. The one workaround (a `ReadableStream` body with `duplex: 'half'`) is Chromium-only and requires HTTP/2 or HTTPS, so it would leave anyone on plain `http://homeassistant.local:8123`, and every Firefox/Safari user, with no feedback at all. Signatures and the `UploadError` shape are unchanged; only the transport moved. `responseType` is deliberately left as text — the proxy-vs-Home-Keeper detection depends on `JSON.parse` *throwing* on an HTML error page, and there's a test locking that.

**Why the bar isn't an HA component.** I checked in the running container first: `ha-linear-progress` doesn't exist under that name, and `ha-progress-bar` / `ha-progress-ring` exist only inside lazy-loaded numbered chunks — never the entry bundle — so neither is registered on `/home-keeper`. `REQUIRED_COMPONENTS` only *waits* for a registration, it can't cause one, so using either would have rendered an invisible un-upgraded element. Same failure mode as #144 (`ha-dialog`'s slot API). The bar is plain DOM on `var(--primary-color)` / `var(--divider-color)`, with `role="progressbar"` and `aria-valuenow` (dropped, with `aria-busy`, while indeterminate). No `aria-live` on the bar — a region ticking every percent would flood a screen reader; the toast and the `role="alert"` error carry the outcome.

## Tests

- **vitest** — new `frontend/test/upload.test.js` (15 tests) driving a fake `XMLHttpRequest`: URLs, `Authorization` set after `open()`, progress + the final `sent` event, 413-with-JSON vs 413-with-HTML, transport failure, abort. Plus panel-level assertions that an oversized pick makes **no request at all**, renders an `ha-alert` **inside `.hk-doc-add`**, and fires the toast — the direct #159 regression. Same for a part file.
- **pytest unit** — the limit drift guard; `validate_upload_stream` parity with the in-memory variant; every magic-byte signature still matching within `SNIFF_BYTES`; and the temp purge sparing a fresh file, reclaiming a stale one, and surviving an entry it can't delete.
- **pytest integration** — over-limit 413 on *both* views; **a file of exactly `MAX_DOCUMENT_BYTES` uploads, stores and downloads back byte-for-byte** (the real proof of the streaming write across ~100 flush batches, and it pins down that multipart framing overhead does *not* count against the raised `_client_max_size`); and the staging area empty after an accepted upload, a rejected one, and a malformed two-file body.
- **e2e** — `toBeInViewport()` on the error and that it lands within 200 px of the button, plus a real upload showing the bar with a non-zero box.

Three real bugs were caught by writing these rather than by reading the code: `bytesSent` was only tracked when a caller passed `onProgress` (yet the proxy-cut-off branch depends on it); the multipart leak below; and my own e2e test wasn't hermetic.

Local runs after the rebase: **310 vitest, 691 pytest unit, 22 document/part-file integration, 44 e2e (full suite), ruff, `tsc --noEmit`** — all green. mypy reports no errors in the changed files locally, though a clean local HA install isn't possible here (a transitive dep fails to build), so CI's mypy job is authoritative.

**On the coverage badge:** the headline "15% of new statements covered" is dominated by `manuals.py`, which reads **0% both before and after** this PR (0/232 → 0/299) — it imports Home Assistant, so the coverage job's unit run never imports it at all; it's exercised by the Docker integration tests, which aren't measured. `__init__.py` is the same (0/370). The one genuinely-measured file is `documents.py`, and it goes **up**: 97% → 98%.

## Review follow-ups

Amazon Q flagged two criticals; both were real and are fixed:

- **Multipart leak** — reassigning `uploaded` per file part orphaned the earlier temps. There was a second edge on the same line: when the *next* part was over the ceiling, `_stream_to_temp` returned `None`, overwriting the only reference to the previous temp file. Now discarded before the next is streamed, with an integration test I verified **fails without the fix**.
- **Reload race** — `rmtree` at setup could delete a live upload, since setup re-runs on reload while the view persists. Now age-based.

I pushed back on one point: the review said streaming "weakened" magic-byte validation because the old code "saw the whole file". It didn't — `sniff_content_type` only ever inspected the first 12 bytes, so examining 16 is strictly *more* than before. No regression here.

## Gates

- Screenshots captured against a clean container and visually inspected; only the two relevant PNGs are committed (the harness re-renders all 42 each run).
- `walkthrough.capture.ts` extended with an upload beat (throttled via CDP so the bar's motion is visible); I confirmed the published gif stepped through the new surface — oversize rejection → 3% → 51% → completed.
- README updated: new limit, the streaming note, and the proxy examples bumped from `30M` to `110M`.
- New conventions in `.amazonq/rules/architecture-and-code.md`: "don't build on lazily-loaded HA components" and "uploads stream to disk — never buffer a whole file".
- **No version bump and no beta section** — per your instruction. The `preview-release` label is still on, which publishes an ephemeral `dev` build from the PR head; say the word and I'll remove it.

Worth flagging for your call: `_MAX_DOCUMENTS` is 50 per asset, so the theoretical worst case per appliance is now 5 GB of blobs. Left as-is.

https://claude.ai/code/session_019yBHbXuZpeTCcxbq6TPME4


prestomation:
/q review Please be critical and skeptical — surface the most serious problems first, but don't withhold minor ones. This PR moves the panel's two upload paths from `fetch` to `XMLHttpRequest` to get byte progress, adds a client-side size pre-check, and changes where upload failures are rendered.

Specific things I want scrutinized:

**Correctness of the new transport (`frontend/src/api.ts` `postUpload`)** — Are any XHR failure paths unhandled or double-settling the promise (e.g. `error` after `load`, `abort` racing a completed request, a signal that aborts after resolution)? `responseType` is deliberately left as text so `JSON.parse` throwing is what distinguishes a proxy's HTML 413 from a Home Keeper JSON 413 — is that discrimination actually sound, and is `bytesSent < file.size` a defensible signal for "the connection was cut mid-body"? Note no `xhr.timeout` is set by design; argue if that's wrong. Also: does dropping `fetch` lose anything (credentials/cookie behavior, redirects, HTTP/2 multiplexing) that mattered here?

**The render-vs-patch split (`panel.ts`)** — `_onUploadProgress` mutates state and then patches the live DOM directly instead of re-rendering (a render replaces the whole shadow root). Are there states where the patched DOM and the state can diverge — a re-render mid-upload, a second upload starting, the panel navigating away or disconnecting while an upload is in flight? Is the `AbortController`/timer cleanup in `_runUpload`'s `finally` leak-free? Is the one-shot `_scrollToError` flag correct given `mergeAsset` clears errors on every keystroke?

**Maintainability** — Both upload call sites now go through `_runUpload`; is the shared abstraction actually the right seam, or does the `key` scoping (`document` / `part:<id>`) leak complexity? Is single-slot upload state (one upload at a time, all upload buttons disabled) a reasonable constraint or an artificial one?

**Correctness of the size pre-check** — Off-by-one at exactly `MAX_DOCUMENT_BYTES` (should be accepted; there's a test), and whether mirroring the backend constant in `frontend/src/limits.ts` with only a pytest drift guard is a sound approach versus injecting it at build time.

**HA best practices** — Is reaching for `hass-notification` plus an inline `ha-alert` the idiomatic way to report a failed action in a custom panel? The progress bar is hand-rolled plain DOM because `ha-progress-bar` lives only in lazy-loaded chunks and is never registered on a custom panel's page (verified in the container) — is there a supported way to use HA's own component that I've missed?

**Accessibility** — `role="progressbar"` with `aria-valuenow`, dropped in favor of `aria-busy` while indeterminate, and deliberately no `aria-live` on the bar. Is the completion/failure announcement (toast + `role="alert"`) sufficient?

**Security** — The `eval` in `tests/unit/test_upload_limit_parity.py` is regex-guarded to digits/`*`/`+` before evaluation; please confirm that's airtight or suggest better.

---

prestomation:
/q review Second push — please be critical and skeptical, most serious issues first, and don't withhold minor ones. This one is a backend change: uploads no longer buffer the whole file in memory, they stream to a temp file and are moved into place, which is what let the ceiling go from 25 MB to 100 MB. Focus on `manuals.py` (`_parse_upload`, `_stream_to_temp`, `async_store_document`/`async_rename_document`/`async_discard_upload`) and `documents.py` (`validate_upload_stream`).

**Correctness / resource safety of the streaming path** — Is the temp file leaked on *any* path? Consider: client disconnects mid-upload, `read_chunk` raising, aiohttp's `HTTPRequestEntityTooLarge` backstop, an exception between `_parse_upload` returning and the handler's `try/finally`, and the multi-part case where a second file part overwrites `uploaded` (I believe the last wins — is that a leak?). Is `try/finally` with `async_discard_upload` genuinely sufficient, and is the "no-op after the move" assumption safe if the move partially failed?

**Atomicity and concurrency** — `_move` uses `os.replace` and relies on the temp dir being a sibling of the asset dirs (same filesystem). Is that guaranteed given the storage root is under the HA config dir? What happens on two concurrent uploads to the same asset/document id, or a re-upload replacing a part's file while a GET is streaming the old blob via `web.FileResponse`?

**Security** — Validation now runs on `(header, size)` rather than the whole blob, so only the first 16 bytes are sniffed. Does that weaken the magic-byte allowlist in any way that matters (e.g. polyglot files, a signature needing more bytes)? Is `.incoming` genuinely unreachable through `documents.safe_segment` / `resolve_under_root` (it strips leading dots — please verify that reasoning rather than take my word)? Is `tempfile.mkstemp` in a shared dir the right primitive here?

**The 100 MB ceiling itself** — Peak memory should now be the 1 MB flush buffer regardless of file size; is that actually true, or does aiohttp buffer more upstream? Are the executor round-trips (~100 per 100 MB) a problem for the event loop? And is `request._client_max_size = MAX_DOCUMENT_BYTES` correct given multipart framing overhead — an integration test says a file of exactly the limit is accepted, but I'd like a second opinion on whether that's load-bearing or incidental.

**Startup cleanup** — `async_cleanup_temp_uploads` rmtrees `.incoming` during setup. Can that race an in-flight upload on a config-entry reload (which re-runs setup while the HTTP view stays registered)?

**Maintainability / HA best practices** — Is `UploadedFile` the right seam, and is keeping both `validate_upload` and `validate_upload_stream` justified or redundant? Is all blocking IO correctly off the event loop?

---

prestomation:
Thanks — both criticals were real. Fixed in b52803a.

**1. Multipart leak — confirmed, and worse than described.** You're right that reassigning `uploaded` orphans the earlier temp. There's a second edge on the same line: when the *next* part is over the ceiling, `_stream_to_temp` returns `None`, which overwrote the only reference to the previous temp file — so it leaked with nothing left to clean it up with. Now the previous temp is discarded before the next is streamed. I added an integration test that posts two file parts and asserts the staging area is empty, and confirmed it **fails without the fix** rather than trusting that it would.

I kept "last wins" rather than rejecting multi-file bodies — no legitimate client sends more than one, and 400-ing it is a behaviour change with no upside.

**2. Reload race — already fixed, you reviewed the previous commit.** Same conclusion you reached, independently: `rmtree` at setup can delete a live upload because setup re-runs on reload while the view persists. It's now an age-based purge (24h) that only reclaims files no request could still be writing, in `documents.purge_stale_temps` — HA-free and unit-tested, including the case that matters (a fresh file survives, a stale one doesn't). I didn't take the "create `.incoming` on demand" route: `_make_temp` already mkdirs, so the directory isn't the problem — the *wipe* was.

**3. Magic-byte validation — I don't think this is a regression at all.** The premise that "the original in-memory validation saw the whole file" isn't quite right: the old code passed the whole blob to `sniff_content_type`, but that function only ever inspects the first 12 bytes (`data[:6]`, `data[8:12]`). Nothing downstream looked at the rest. So streaming with `SNIFF_BYTES = 16` examines *strictly more* than before — the polyglot exposure is identical to what shipped in 0.9.0, not something this PR introduces. Agreed the mitigation is that the allowlist is narrow and `content_type` is derived from the sniff rather than the client. Streaming-hash validation would be a new feature, not a fix, so I've left it.

**4. Concurrent uploads to the same `document_id`** — agreed this is a pre-existing corner case, unchanged by this PR (the old code had the same last-write-wins shape, just with `write_bytes` instead of `os.replace`). Not fixing it here; it needs a per-asset lock, which is a separate change.

**5. `async_rename_document` fragility** — I traced it: if the rename fails after the initial move, the error path removes the store entry and deletes the blob at the *original* path, and the `finally` no-ops because the temp was already moved. So it converges correctly, just via two cleanup calls. Left as-is.

Agreed on the rest — executor hop count, `os.replace` same-filesystem guarantee, `try/finally` sufficiency, and that the exact-limit integration test is load-bearing rather than incidental.

---

PR #167, opened 2026-07-30
Render Notes fields as Markdown (Fixes #163)

### LINKED ISSUE #163: Notes field should support markdown
The "Notes" fields in various places (devices, tasks) should render as Markdown.

prestomation:
Fixes #163.

Every **Notes** field — task, appliance, part, and per-completion — now renders as GitHub-flavoured Markdown instead of escaped plain text. The task detail page didn't even preserve line breaks before, so a multi-step procedure collapsed onto one line.

## Why `ha-markdown` and not a bundled parser

I checked this against the shipped `home-assistant-frontend` wheel rather than assuming:

- `ha-markdown` wraps `ha-markdown-element`, which parses with **marked (GFM)** and sanitizes with **DOMPurify inside a Web Worker**, retargets off-host anchors to `target="_blank" rel="noreferrer noopener"`, and ships theme-aware styles for links/code/`pre`/headings.
- Hand-rolling a renderer means maintaining sanitizer-adjacent code — the class of thing that grows CVEs — and shipping a subset with no tables. Bundling `marked`+DOMPurify means adding `@rollup/plugin-node-resolve` + `commonjs`, ~70 KB on both bundles, and two supply-chain deps to a frontend that currently ships **zero** JS runtime dependencies.
- **The catch:** `ha-markdown` is lazily loaded — absent from all three eager entrypoints (`app`/`core`/`custom-panel`), inlined into ~10 lazy chunks. One of those (`34172.*`) contains **both** `hui-markdown-card` and `ha-markdown`, so `loadCardHelpers().createCardElement({type:'markdown'})` is a reliable one-hop nudge to register it.
- `window.loadCardHelpers` is itself installed by the Lovelace chunk, so a **cold deep-link straight to `/home-keeper`** may not have it. `markdownBlock` then falls back to escaped text with `white-space: pre-wrap` — still better than the old rendering — and `_ensureMarkdown()` retries on later renders, upgrading to real Markdown once a dashboard has been visited.

## Closing the issue also meant filling two gaps

The issue says "devices, tasks", but:

- **Appliances had no notes field at all.** Added as `assets._PROSE_FIELDS`, deliberately separate from `_TEXT_FIELDS` (which documents *"these sync into the device registry"*) so prose never leaks onto the device card. Exposed through `add_asset`/`update_asset` with strings in all 16 locales. No migration needed — an asset missing the key normalizes to `""`.
- **Parts have always stored notes**, but the panel offered no editor and never showed them. Both added.

## Authoring

The inline note editor was gated to problem-sensor tasks; it now serves **every** task and appliance, with a live preview under the textarea. The task, appliance, part, and completion forms gained the same preview, fed from each form's existing `value-changed` handler so the field keeps focus while typing (the technique `_updateSensorHint` already used). Previews are **debounced** (each render round-trips through the worker) and stay hidden until `looksLikeMarkdown()` is true — echoing plain prose back at the author is noise.

**One deliberate deviation:** part notes render *read-only* on the appliance detail page rather than getting their own inline editor. Parts are edited as a whole in the parts editor, so letting one field be inline-editable while its siblings aren't would be inconsistent — and N inline editors in a list reads badly.

Notes are stored as **Markdown source**, not HTML, so the raw text is still what the `todo`/`calendar` item descriptions and the services/events hand out.

## Screenshots

Task detail — the note renders as a heading, numbered list, inline code, quote and link, and every task now has an **Edit note** button:

Task detail page with a Markdown note rendering bold text, inline code, a numbered list, a blockquote and a link, plus an Edit note button

The inline editor with its live preview:

The inline note editor open on a task, showing a textarea of Markdown source with a live preview below rendering the heading and bullet list

Appliance detail — the new Notes card (with a table), and a part note under Anode rod:

Appliance detail page showing a new Notes card rendering bold text, a settings table and italics, plus a Markdown note under the Anode rod part

The dashboard card's compact notes (middle column, `show_notes: true`):

Dashboard with three Home Keeper cards; the middle one renders a task's Markdown note compactly beneath the task name

## Testing

All run locally before pushing:

- `npm test` — 274 pass (10 files), including a new `markdown.test.js` (44 cases covering both render branches, escaping, the `loadCardHelpers` nudge, timeout/failure paths, debounce, and the `looksLikeMarkdown` heuristic) plus new panel/card cases for each surface and the save paths.
- `pytest tests/unit` — 681 pass, including appliance-notes normalization/round-trip and a guard that notes never leak into the registry-synced identity fields.
- `ruff check` / `ruff format --check` clean. `mypy` shows no new errors (HA wouldn't install in this sandbox — a PyRIC wheel failure — so CI is the real check; the one non-import error also reproduces on the unmodified baseline).
- **Real browser**: the e2e capture asserts the rendered `<strong>`, `<ol><li>` and `<table>` inside `ha-markdown`, so a regression fails loudly instead of silently degrading to plain text. All six screenshot suites pass and every changed PNG was visually inspected.
- The walkthrough tour was extended to type Markdown into the note editor (showing the preview catch up) and to linger on the appliance Notes table; it was run locally and passes.

## Release

Cut `0.10.0b2` (`manifest.json` + `const.py` + CHANGELOG). README gained a **Notes are Markdown** section with use cases and a screenshot; `docs/INTEGRATING.md` notes that `notes` is Markdown-rendered; the new conventions are recorded in `.amazonq/rules/architecture-and-code.md`.

https://claude.ai/code/session_01TheHoF7AxjfHmisBBE1gKy

---
_Generated by [Claude Code](https://claude.ai/code/session_01TheHoF7AxjfHmisBBE1gKy)_


prestomation:
/q review Please be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. This PR renders every Notes field as Markdown via HA's lazily-loaded `ha-markdown`, adds a `notes` field to appliances, and adds live-preview note editors. Areas I'd most like scrutinized:

**Security (highest priority).** `frontend/src/markdown.ts` is the only place user free text now reaches the DOM as anything but escaped text. `markdownBlock()` escapes the source into a `data-md` attribute and `wireMarkdown()` moves it onto `ha-markdown`'s `content` property. Is there any path where attacker-controlled note text escapes the attribute, or reaches `innerHTML` unescaped? Check the fallback branch (`.hk-md-plain`) and `createPreview`'s `paint()` (it assigns `body.innerHTML = markdownBlock(text)`) especially. Note text is writable by any non-admin via `home_keeper.update_task`, then rendered for an admin — the same stored-XSS shape a previous fix addressed for completion `photo` URLs.

**Correctness / races.** `ensureMarkdown()` shares one in-flight promise and clears it when settled; `_ensureMarkdown()` re-renders when it flips true. Can that loop, double-render, or leave a stale preview? Are the `MarkdownPreview` debounce timers always disposed — `_render()` disposes `_formPreviews`, `_wireNoteEditor` disposes `_notePreview`, but is anything leaked on `disconnectedCallback`, or when a dialog closes without a re-render?

**The `looksLikeMarkdown` regex** (`markdown.ts`) gates whether a preview shows. Is it prone to catastrophic backtracking on a long note? Any obvious false negatives that would confuse an author (they type Markdown, no preview appears)?

**Data model.** `assets._PROSE_FIELDS` splits `notes` from `_TEXT_FIELDS` so it isn't registry-synced. Does `merge_update` handle every partial-update shape correctly (omitted / empty / null)? Any stored asset predating the field that could break, and is the panel's part-notes save (which rewrites the whole `parts` array) safe against clobbering backend-managed part fields?

**HA best practices & maintainability.** Is leaning on `window.loadCardHelpers` to force-register a lazy HA element acceptable, or is there a better-supported hook? Is the fallback (escaped `pre-wrap` text on a cold deep-link to `/home-keeper`) an acceptable degradation? And is `panel.ts` accreting too much — should the note-editor/preview logic be extracted?

**Performance.** Each preview render round-trips through `ha-markdown`'s Web Worker. Is a 200 ms debounce right, and does the dashboard card re-render more than it needs to when many rows have notes?

---

prestomation:
Triaging the two Cue passes — they reached opposite conclusions, and the second one is right.

## The three "critical" findings in the first (automatic) review are false positives

All three cite code that doesn't exist in this repo:

**1 & 2 — path traversal in `assets.py` (lines 66, 76).** `assets.py` is the pure asset model: no HA imports, and **no filesystem access at all**. `grep -n 'write_text\|assets_path\|open(\|unlink\|Path('` over it returns nothing. There is no `Asset` class with `.name`/`.content`, and no `self.assets_path`. The cited lines are comment text inside the `_PROSE_FIELDS` block this PR adds:

```
66  # Free-form prose about the appliance ("the filter is behind the kick plate…").
...
76  # Every verbatim-kept text field, in the order they are normalized.
```

The only module that writes files is `manuals.py`, which this PR doesn't touch — and it already has the guard: `documents.document_path()` composes every path through `resolve_under_root(root, asset_id, f"{document_id}__{safe_segment(filename)}")`, docstring *"The on-disk path for an uploaded document, guarded against traversal."* Applying the suggested patches would add `Path`/`ValueError` handling to a module that has neither a filesystem nor the symbols they reference.

**3 — `allowHTML: true` in `markdown.ts` (line 17).** That option doesn't exist anywhere in the codebase (`grep -rn allowHTML custom_components/` → no matches), and line 17 is prose inside the module docstring. We never enable HTML: the source is escaped into a `data-md` attribute, and `ha-markdown` sanitizes with DOMPurify in a Web Worker before rendering.

The second review — the one that actually read the rendering paths — reaches the same conclusion: *"No XSS vulnerabilities found... The stored-XSS attack surface you describe is fully mitigated by the escape-then-property-assign pattern."* Not making these changes.

## Acted on from the second review

- **Preview timers weren't disposed on unmount** — correct, and a good catch. `disconnectedCallback` already clears `_persistTimers` for exactly this reason; I'd missed extending it to `_notePreview`/`_formPreviews`, so unmounting mid-debounce left a timeout to fire against a detached subtree. Fixed in 8ba1f1e.
- **Verify `_merge_parts` preserves backend-managed fields** — verified: it always restores `file_name`/`file_content_type`/`file_size`, and keeps a stored `last_replaced` unless the caller explicitly sets one. That was already true, but the new part-notes editor makes the whole-array round-trip reachable from a note edit, and nothing tested it from that direction — added `test_part_notes_edit_preserves_backend_managed_fields`.
- **`null` notes on a partial update** — coerced to `""`; `normalize_fields` uses `value not in (None, "")`. Covered by `test_notes_normalize_and_round_trip`.
- **Extracting note-editor logic out of `panel.ts`** — agreed it's accreting, and agreed with your own bar ("only split when a subsystem has no entanglement with the parent's state"). The note editor reads `_noteEdit`, calls `_render`, and shares the detail-page wiring pass, so it isn't cleanly separable yet. Leaving it, noting the concern.
- **200 ms debounce / card render cost** — no change, matching your assessment.

## Unrelated real failure, also fixed

The `preview` job was red for a reason neither review mentioned: `sync-docs.mjs` splits `README.md` into one page per H2, so my two in-README `#notes-are-markdown` links became cross-page anchors and broke the Docusaurus build. The new section also wasn't registered in `USER_SECTIONS`, so it was being silently dropped from the docs site (the build logged 16 pages — exactly that list's length). Both fixed in 8ba1f1e and verified with a local site build (17 pages, no broken anchors).

---

prestomation:
/q review Please review the follow-up commit `8ba1f1e` critically and skeptically — most serious issues first, and don't hold back minor ones. Three focused changes:

1. **`disconnectedCallback` now disposes Markdown preview timers** (`panel.ts`). Is the teardown complete and correct? I null `_taskNotePreview` without calling `dispose()` on it directly because it's also pushed into `_formPreviews` (which is disposed) — verify that's actually true on every path that creates it, or it's a leak I've just papered over. Also: is there any path where a preview is created *after* `disconnectedCallback` runs (an in-flight `ensureMarkdown()` promise resolving post-unmount and triggering `_render`)?

2. **Docs-site fix** (`website/scripts/doc-map.mjs`, `sync-docs.mjs`). I added the new README section to `USER_SECTIONS` and its anchor to `ANCHOR_ROUTES`. Is there a drift risk worth a test here — a README H2 that no `USER_SECTIONS` entry claims gets silently dropped from the docs site, which is exactly the bug I hit, and it only surfaced because a *separate* broken anchor failed the build. Would a guard that asserts every README H2 is either mapped or explicitly excluded be worth adding?

3. **New test `test_part_notes_edit_preserves_backend_managed_fields`** (`tests/unit/test_assets.py`). Does it actually pin the behaviour it claims, or is it passing for the wrong reason? Are there part fields *other* than `file_*`/`last_replaced` that a whole-array round-trip from the panel's part editor could silently clobber — particularly for reconciler-derived parts?

Also worth re-checking across the whole PR: any HA best-practice or maintainability issue in how the note editor threads state through `panel.ts`, and whether the `_noteEdit` discriminated-union refactor left any dead paths behind.

---

prestomation:
/q review Small follow-up in `13b2708` — please check it critically, and re-check the PR as a whole while you're there.

On the previous round's observation about `disconnectedCallback`: the ordering was already correct (`_notePreview.dispose()` ran *before* `this._notePreview = null`), but the block read ambiguously and `_taskNotePreview` was nulled with no dispose of its own. That's safe only because of an unstated invariant — `_attachNotePreview` registers every preview it builds in `_formPreviews`, and `_taskNotePreview` is only ever assigned from that helper, so disposing the list covers it; `_notePreview` is the one exception since `_wireNoteEditor` builds it directly. I've documented that rule and regrouped to dispose-everything-then-null.

What I'd like verified:

1. **Is that invariant actually airtight?** If any future code path assigns `_taskNotePreview` from something other than `_attachNotePreview`, the teardown silently leaks a timer. Is a comment sufficient, or should `_attachNotePreview` be the only way to construct a preview at all (e.g. route `_wireNoteEditor` through it too, so there's exactly one registration path)? I'd rather remove the footgun than document it if the refactor is clean.

2. **Post-unmount preview creation.** Still open from last round: an in-flight `ensureMarkdown()` promise resolving *after* `disconnectedCallback` calls `_render()`, which would rebuild forms and previews on a detached element. Is that reachable, and does it matter?

3. **Anything the earlier passes missed** now that CI is fully green — correctness, HA best practices, or maintainability across the whole changeset.

---

prestomation:
Both actionable findings from the last round are fixed in `3a0761b`; the third is a false positive.

**1. Registration footgun — removed rather than documented.** Agreed that a documented invariant a future caller can quietly break isn't good enough. `_wireNoteEditor` was the only place building a preview directly, so it now goes through `_attachNotePreview` as well. That makes it the sole constructor, so `_previews` (renamed from `_formPreviews` — it's no longer just forms) holds every preview on screen and one loop disposes them all. The `_notePreview` field is gone entirely; nothing needed it once disposal stopped being per-field. `grep createPreview panel.ts` now returns exactly one call site.

**2. Post-unmount re-render — real, fixed.** You're right that it's reachable: `ensureMarkdown()` awaits a lazy chunk load, so it can settle after unmount, and the callback then rebuilt the whole panel — forms, previews, timers — onto a detached element with nothing left to tear it down. Now gated on `isConnected`. I went with that over plumbing an abort signal out of `markdown.ts`: the promise is a best-effort element registration whose result is only ever used to trigger a repaint, so there's nothing to cancel — the only question is whether anyone still cares about the answer, and `isConnected` is exactly that question.

Both are pinned by tests rather than left to inspection:
- unmounting mid-typing leaves a pending preview render cancelled;
- a late `ha-markdown` registration doesn't mutate a detached panel's shadow root.

**3. `_confirmOnKey` orphan — not a bug.** `_renderConfirmDeleteDialog` does add the listener each render, but `_openConfirmDialog` already removes any prior listener *and* scrim before opening a new one (`panel.ts:1302-1311`), with a comment naming that exact scenario: *"Drop any prior scrim (and its keydown listener) before opening a new one, so a second open — or a stale scrim — can't orphan the earlier overlay + handler."* So the rapid-double-open case is handled. It's also pre-existing code this PR doesn't touch, so I'd leave it alone regardless.

Local run before pushing: 276 vitest (2 new), 683 pytest, tsc/ruff clean. CI was fully green on `13b2708` — all 14 checks including `mypy`, `e2e`, `integration`, `walkthrough`, and the `preview` job that this PR had originally broken.

---

prestomation:
/q review Please review `3a0761b` critically — most serious issues first, minor ones included. It collapses preview construction to a single path and guards a post-unmount re-render:

1. **Is `_attachNotePreview` genuinely the only path now?** `_wireNoteEditor` was routed through it, `_notePreview` is deleted, and `_formPreviews` became `_previews`. Verify no preview can be constructed or orphaned outside that helper, and that the *ordering* still holds: `_render` clears `_previews` before building markup, `_hydrate` (which calls `_wireNoteEditor`) runs after. If a preview is ever attached before `_render`'s reset in some path I've missed, it gets disposed immediately and the editor silently stops previewing — is that reachable?

2. **Is `isConnected` the right guard on the `ensureMarkdown()` callback?** I chose it over plumbing an abort signal out of `markdown.ts` on the grounds that there's nothing to cancel — it's a best-effort element registration whose result only triggers a repaint. Push back if an abort is actually warranted. Also: can the panel be *re*-connected (moved in the DOM) such that skipping that render leaves it stuck on the escaped-text fallback with no later trigger to upgrade it?

3. **Do the two new tests actually pin the behaviour**, or would they pass against the old buggy code? The unmount-mid-typing one asserts no `ha-markdown` appears after the debounce; the late-registration one compares `shadowRoot.innerHTML` before and after. Both feel like they could be vacuously true — tell me if so.

4. Anything else across the PR now that it's had several passes — correctness, HA best practices, maintainability.

---

prestomation:
Thanks — the verification of the previous two fixes matches what I see. Acted on two of the new points in `18f603b`, and two are false positives.

**Fixed: stale references can no longer re-arm a timer.** Rather than chase each place a reference might linger, I closed it at the primitive: `dispose()` is now permanent, not just a timer cancel. `update()` is inert once disposed, so a handler still bound to a detached form can't arm a fresh render against DOM nobody will clean up. That makes the whole "is this reference still live?" question moot instead of something every call site has to get right.

**Fixed: teardown centralized** in `_disposeAllPreviews()`, used by both `_render` and `disconnectedCallback` — agreed that duplication is exactly how a third caller ends up half-cleaning.

**On whether the tests are vacuous — good question, and I checked rather than asserted.** I stubbed `_disposeAllPreviews` out to a no-op and re-ran: all three fail. So they pin the behaviour, not its shadow.

```
× cancels a pending preview render when the panel unmounts mid-typing
× disposes previews when navigating away with a preview timer armed
× registers every preview it builds so one teardown covers them all
Tests  3 failed | 10 passed (13)
```

I also added the panel-level integration coverage you asked for: the registration contract is now pinned directly (`_previews.length` equals the number of `.hk-md-preview` elements actually on screen, and is empty after teardown), which is what catches a future path that builds a preview outside `_attachNotePreview`. And a navigate-away case, since unmount isn't the only teardown path.

**Not changed — false positives:**

- **`_noteEdit` never cleared on form close.** It's cleared in three places: `_applyLocation` (line 823), `_closeNoteEditor` (1108) and `_saveNote` (1129). It also tracks the *inline note editor*, not the edit form — opening the edit form navigates off the detail page, which clears it via `_applyLocation`.
- **Stale `_taskNotePreview` after a view change.** A view change goes through `_render`, which disposes every preview and nulls that reference — as your own analysis noted two paragraphs earlier. With `dispose()` now permanent, it's doubly inert.

**Not changing — `ensureMarkdown` retrying across renders.** That's deliberate and cheap: when `window.loadCardHelpers` is absent the call returns `false` immediately without touching the network, and `markdownReady()` short-circuits it permanently once the element registers. Caching the negative result would trade that for a window where the panel stays stuck on the plain-text fallback after the user visits a dashboard — the exact case the retry exists to fix.

Local run: 279 vitest, 683 pytest, tsc/ruff clean.

---

prestomation:
/q review Please review `18f603b` critically — worst first, minor included. It makes `dispose()` permanent (an `update()` after disposal is now a no-op) and centralizes teardown in `_disposeAllPreviews()`.

1. **Is permanent disposal the right contract, or does it hide a bug?** An `update()` that silently no-ops could mask a real ordering mistake — a preview disposed while its editor is still on screen would now fail *silently* (no preview, no error) instead of rendering. Is there a path where that's reachable, and would failing loudly be better?

2. **`_disposeAllPreviews()` is called at the top of `_render` and in `disconnectedCallback`.** Any path that attaches a preview and then hits `_render` before the user can type — leaving the editor with a disposed, permanently-inert preview? That's the failure mode I'm most worried about introducing with change 1.

3. **Do the new panel tests over-fit?** The registration test asserts `_previews.length` equals the count of `.hk-md-preview` in the shadow root — it reaches into a private field, which pins an implementation detail. Is that the right level, or should it assert only observable behaviour? I verified all three fail against a stubbed-out `_disposeAllPreviews`, so they're not vacuous, but non-vacuous isn't the same as well-targeted.

4. Anything remaining across the PR. It's had several passes now, so I'm specifically interested in whatever the earlier rounds' focus on teardown might have crowded out — the backend `notes` field, the services/translations surface, the docs-site split, or the e2e/screenshot coverage.

---

prestomation:
Good question on the e2e specificity — that's the one that could genuinely have been hand-waved, so I mutation-tested it rather than reasoning about it.

**Confirmed: the assertions are inside `ha-markdown`, and they do catch a missed `wireMarkdown()`.** All seven are scoped as `ha-markdown <descendant>`:

```
.hk-detail-inner ha-markdown strong      .hk-detail-inner ha-markdown ol li
.hk-md-preview ha-markdown h2            ha-markdown table
.hk-part-notes ha-markdown strong
```

The fallback emits `<div class="hk-md-plain">` containing escaped *text* — no `<strong>`/`<ol>`/`<table>` elements exist at all, and nothing is inside an `ha-markdown` — so none of these can pass against it.

To prove the missed-hydration case specifically, I stubbed `wireMarkdown()` to a no-op, rebuilt the panel, and ran the capture against a real HA container. It fails at exactly the intended assertion:

```
✘  capture Home Keeper panel + usage screenshots (20.5s)
   Error: expect(locator).toBeVisible() failed
   Error: element(s) not found
   > 73 |  await expect(panel.locator('.hk-detail-inner ha-markdown strong').first()).toBeVisible({
```

That's the exact regression you flagged: `content` never set → `ha-markdown.render()` returns nothing → no descendants → red. Mutation reverted, panel rebuilt, working tree clean.

That's now the third thing on this PR verified by mutation rather than assertion — the unit teardown tests, the panel integration tests, and this. It seemed worth holding the "works in dev, breaks in prod" pattern you identified to that standard, since a silently-unhydrated element is exactly the failure that looks fine until it isn't.

On the remaining two: the `notes`-never-reaching-the-device-registry property is pinned by `test_notes_are_not_synced_into_device_identity_fields`, and structurally by `_PROSE_FIELDS` being excluded from the tuple `devices.py` reads — but I agree a real device page is the honest check, and `8-panel-appliance-detail.png` in the PR body shows the appliance's DETAILS block with no Notes row (the notes live in their own card below it). The services point matches what I found: both websocket commands delegate to the same store methods.

**CI is fully green on `18f603b`** — all 14 checks, including `e2e`, `integration`, `walkthrough`, `mypy` and the `preview` job.

---

Issue #150, opened 2026-07-23
notification action texts are not translated


davidm-glitch:
The action labels are not translated, they are hardcoded in english.

For example here https://github.com/prestomation/ha-home-keeper/blob/0a46eed148f77804167bed70d6ebc189b708b72d/custom_components/home_keeper/notifications.py#L155

Would be nice to have these also translated 😶‍🌫️

prestomation:
Thank you for the report! I've addressed this and it will be in the next version, 0.10.0.

Please let me know of any other bug reports or feature requests you have

prestomation:
Hello.

If you'd like to test before the release of 0.10.0. You may use hacs to download beta version 0.10.Ob1

davidm-glitch:
I installed it, but like before I am not getting notifications any more. How can I check why this is the case?

prestomation:
There is a notify service action that will trigger your notifications adhoc. Does this work? You could also check the logs. Everything is working ok in my testing and this was just a string change so something breaking would be unexpected

davidm-glitch:
Manually triggering the action works fine, this is what I found in the logs. Does this help?

```
Logger: homeassistant.util.loop
Quelle: util/loop.py:135
Erstmals aufgetreten: 18:59:00 (5 Vorkommnisse)
Zuletzt protokolliert: 18:59:00

[...]

PR #168, opened 2026-07-30
Move notification payload building off the event loop

prestomation:
## Summary

Fixes a follow-up bug reported in #150: after the original "notification action texts are not translated" fix shipped in 0.10.0b1, sending a notification does blocking file I/O directly on Home Assistant's event loop, which HA's own blocking-call detector flags (see the traceback in the [most recent comment on #150](https://github.com/prestomation/ha-home-keeper/issues/150#issuecomment-5133890666)).

- `notifications.build_notification`/`build_digest` resolve translated button/message strings (`_notification_strings`, a `.json` read) and, for multi-day-overdue phrasing, Babel's CLDR plural rules (`_babel_locale().plural_form`, a Babel data-file read) — both blocking, both cached only *after* the first call per language.
- `notifier._send` (used by the `home_keeper.notify` service and the auto/walk notification paths) was calling those builders **inline** on the event loop, so that first per-language call blocked the loop.
- `notifications.py` is deliberately HA-free (see its module docstring) so it can't call `hass.async_add_executor_job` itself — the fix lives at the `notifier.py` call sites instead: a new `_build_payload` helper routes `build_notification`/`build_digest` through the executor, and the `build_all_clear` call (walk-advance "all caught up" note) got the same treatment.

## Test plan

- Added `tests/unit/test_notifier_blocking.py`: spies on the two blocking lookups (bypassing their `functools.cache` so they always re-do the real work) and asserts they never run on the thread driving the event loop. Verified **red** against the pre-fix code (fails, catching the exact bug) and **green** after the fix.
- [x] `pytest tests/unit -v` — 684 passed, 1 skipped
- [x] `ruff check` / `ruff format --check` on changed files — clean
- [ ] `mypy` — could not run in this environment; `pip install homeassistant` fails to build here due to an unrelated legacy dependency (`PyRIC`) breaking under this container's `distutils`. The change only adds a `functools.partial`-wrapped `hass.async_add_executor_job` call, matching the existing pattern used elsewhere in the codebase, so it should type-check cleanly — please have CI confirm.

No panel UI or new user-facing UI surface touched, so the screenshots/video-walkthrough gates don't apply. This is a bug fix, so no beta version bump per AGENTS.md.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NcpTpNUVXZD9mYut1Sj23R)_


prestomation:
/q review Please give critical, skeptical feedback on this fix, and don't withhold minor issues. Focus especially on:

- **Correctness**: does routing `notifications.build_notification`/`build_digest`/`build_all_clear` through `hass.async_add_executor_job` via `functools.partial` actually eliminate the blocking-call warning for every call site that can reach them (service call, auto-notify, walk-advance)? Any path I missed?
- **Concurrency/thread-safety**: `notifications._notification_strings`/`_babel_locale` are `functools.cache`d — is there any risk from concurrent executor-thread population of that cache (e.g. under `asyncio.gather` fan-out across multiple notifications), or is `functools.cache`'s internal locking sufficient here?
- **Test quality**: is `tests/unit/test_notifier_blocking.py`'s approach (spying on the two blocking lookups via `__wrapped__` to bypass caching, and asserting the call thread differs from the event-loop thread) a sound way to catch this class of bug, or is there a cheaper/more standard way to assert "no blocking call happened" that I should be using instead?
- **HA best practices**: is `hass.async_add_executor_job(functools.partial(...))` idiomatic for this, or is there a better-established pattern in this codebase / in HA generally for offloading a keyword-heavy blocking call?

---

prestomation:
Thanks for the detailed review — triaging each point:

- **`build_all_clear` coverage**: confirmed it does hit the blocking path — it calls `_t(lang, "all_clear_title")`/`_t(lang, "all_clear_message")`, and `_t` resolves through `_notification_strings(lang)` (the same blocking `.json` read). The executor wrapper there is necessary, not defensive overhead.
- **Cache pre-warm at startup**: reasonable idea for avoiding a first-burst disk-read pile-up, but it's a performance enhancement orthogonal to this bug fix (which is specifically about not blocking the *event loop*, regardless of how many executor threads end up doing redundant cold reads). Leaving it out of scope here; happy to file it as a follow-up if it's wanted.
- **Test robustness — mocking `async_add_executor_job` instead of thread-ID comparison**: I kept the thread-based assertion deliberately. It tests the actual property that matters (blocking I/O never runs on the event-loop thread), not the specific mechanism used to avoid it — so a correct fix using a different technique would still pass, and the point about "a buggy implementation moving work to *some* other background thread would still pass" is fine, since landing on any non-loop thread already satisfies the real requirement. Asserting on `async_add_executor_job` call args would be tighter to today's implementation but more brittle to a legitimate refactor. I'll leave it as-is.

CI caught one real issue in the test itself (unrelated to the review): the `_FakeHass` didn't have a `.data` dict, which broke under `pytest-homeassistant-custom-component`'s real `device_registry`/`area_registry` singleton helpers in CI (my local run had no `homeassistant` package installed, so that path was stub-covered and I missed it). Fixed in e0198aa.

---

PR #176, opened 2026-08-03
Hide the dashboard task card when its filter matches nothing

prestomation:
## Summary

- Adds an opt-in **`hide_when_empty`** toggle to `custom:home-keeper-card`. Today the card always keeps its full grid footprint and shows "No tasks match this filter." when a card's configured filter (status/area/device/label/recurrence/etc.) matches nothing. With this on, the card collapses out of the dashboard layout entirely instead — handy for a dashboard built from several per-subject cards (the dog, the car, each kid) where you only want to see the ones with something due.
- Implemented card-side (`this.style.display = 'none'` + `getCardSize() = 0` when empty), the same technique the popular `auto-entities` card uses for its `show_empty: false` option — it works for *any* arbitrary filter combination a card is configured with, which a fixed backend sensor could never match exactly. (Lovelace's built-in per-card `visibility:` conditions already let anyone hide any card based on an existing entity's state, with no code change needed — but there's no entity that reflects an arbitrary card's own filter, so that path doesn't help here.)
- Bumps to `0.11.0b1` per the project's "always cut a beta for a new feature" convention.

## Screenshots

Same dashboard as `card-dashboard.png` (3 cards: Home maintenance, By status, Dog), but the "Dog" card is reconfigured with `hide_when_empty: true` and a non-matching label filter — it disappears from the layout entirely, leaving only the two remaining cards:

Dashboard with the label-filtered card hidden — only Home maintenance and By status cards remain visible

## Test plan

- [x] `npx vitest run` (root) — 312 passed, including 2 new `HomeKeeperCard hide_when_empty` cases (stays visible + shows the alert by default; hides and later reappears once a task matches, when configured).
- [x] `npm run build` in `custom_components/home_keeper/frontend` — typechecks/builds cleanly.
- [x] `tests/e2e/tests/card.spec.ts` (full Playwright card suite) against the seeded HA container — all 10 pass.
- [x] `tests/e2e/screenshots-card.capture.ts` — extended with a new capture step demonstrating the toggle; screenshot visually inspected above.

---
_Generated by [Claude Code](https://claude.ai/code/session_01C1uTK3TYLckWgEYiQbaE2o)_


prestomation:
/q review Please give critical, skeptical feedback on this change (adds a `hide_when_empty` toggle to `custom:home-keeper-card` that collapses the card via `style.display = 'none'` + `getCardSize() = 0` when its filtered task list is empty). Specifically scrutinize:

- **Correctness**: any render-cycle edge case where the card could flicker between hidden/visible on a refresh, or stay stuck hidden after an error/loading transition (see `_isHiddenEmpty()` and its use in `_render()`/`getCardSize()` in `card.ts`). Does hiding via inline `style.display` interact badly with any Lovelace layout mode (masonry vs. sections/grid) beyond what's called out in the PR description?
- **Maintainability**: is deriving `_isHiddenEmpty()` from `_visibleCount()` (which itself calls `_shaped()`, which is also recomputed elsewhere in the same render pass) an efficiency/duplication concern worth flagging, or is it fine given the small task-list sizes here?
- **HA best practices**: is there a more idiomatic way custom Lovelace cards signal "hide me" that this should use instead of the `auto-entities`-style `display:none` + `getCardSize()=0` trick?

Please surface the most serious issues first and don't hold back minor ones either.

---

prestomation:
Thanks for the review — triaged:

- **Fixed** ("Redundant Computation" / the concrete part of "Race Condition"): `getCardSize()` was calling `_isHiddenEmpty()` (which computes `_visibleCount()` → `_shaped()`) and then, when not hidden, calling `_visibleCount()` again for the row count — recomputing `filterTasks`/`sortTasks` twice per `getCardSize()` call. `_isHiddenEmpty()` now takes the already-computed count instead of deriving it itself, so `getCardSize()` and `_render()` each compute it exactly once. Pushed in 38eb36f.
- **Not changing** the broader "`_shaped()` runs multiple times across a render pass" (`getCardSize` / `_render` / `_listHtml` each independently call it): that's the pre-existing pattern in this file predating this PR, not something this change introduces beyond the one regression above, and Home Keeper's task counts are small (per-household maintenance/appliance tasks), so it's not worth a memoization layer for this PR's scope.
- **Pushing back** on "Missing Guard in `getCardSize()`": `_shaped()` reads `this._hass?.devices`/`this._hass?.areas` with optional chaining, and every downstream helper (`taskAreaId`, `taskLabelIds`, `filterTasks`, `profileMatches`) already takes `devices`/`areas` as optional params — none of them dereference an undefined `hass`. Separately, `_isHiddenEmpty()`/`_visibleCount()` are only reached once `this._loaded` is true, and `_loaded` only flips inside `_refresh()`'s `finally`, which only runs after a `hass`-dependent API call — so by the time `getCardSize()` would call `_shaped()`, `this._hass` is guaranteed set. There's no code path that unsets `this._hass` afterward. I don't see a throw here; happy to be shown a concrete repro if I'm missing one.
- **Not changing** the "stuck hidden while `_error` flips between the two calls" scenario: real in the abstract (JS is single-threaded, but `getCardSize()` and `_render()` are invoked by Lovelace at independent times, so state can change between them), but it's an inherent characteristic of *any* Lovelace card whose `getCardSize()` varies with live content — not unique to `hide_when_empty` — and the window requires an error to appear/clear in the exact gap between two externally-triggered calls. Given the severity/likelihood, a synchronized single-source-of-truth render-state refactor felt like scope creep for this PR; flagging it as a known tradeoff rather than fixing it here.
- **Agreed, no change**: the `display:none` + `getCardSize()=0` approach itself — thanks for confirming it matches `auto-entities`/`entity-filter` convention.

---

PR #184, opened 2026-08-08
Persist the intro banner dismissal server-side per-user (Fixes #182)

### LINKED ISSUE #182: Welcome screen shows up on new devices
Image

Seems like it'd be better if it's "seen" status was stored server-side per-user

prestomation:
## Summary

Fixes #182 — the "Welcome to Home Keeper" first-run banner tracked its dismissed state in `localStorage`, which is scoped to a single browser. Switching devices, opening a private window, or clearing site data brought the banner back.

- Moves the dismissed flag to Home Assistant's own **per-user frontend data store** (`frontend/get_user_data` / `frontend/set_user_data` — core HA websocket commands, not admin-gated), instead of adding a Home Keeper-specific backend command. No backend Python changes needed.
- `api.ts` gets `getIntroDismissed`/`setIntroDismissed` wrappers; `panel.ts` fetches the dismissed state alongside the rest of the panel's data in `_reload()`, and the dismiss handler optimistically hides the banner then persists in the background (best-effort, same as the old `localStorage` write).
- New regression test (`panel.test.js`) mounts two separate panel instances against a shared fake per-user store to reproduce the "different device" scenario and confirm dismissal now follows the user.

## Test plan

- [x] `npx vitest run` — full frontend suite passes (313 tests), including the new regression test for #182
- [x] `npm run build` (Rollup + TypeScript) — panel compiles clean
- [x] Playwright screenshot capture re-run; visually confirmed the banner renders correctly with the new async per-user load

Welcome to Home Keeper first-run banner

No new UI surface is added (pure persistence-layer fix), so the walkthrough-video gate doesn't apply.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ybj7aeCuRCxLqSKse2RjsS)_


prestomation:
/q review Please give critical, skeptical feedback on this fix for #182 (welcome banner reappearing on new devices). Focus on:

- **Correctness**: the optimistic dismiss-then-persist race in the click handler (`panel.ts`) — banner hides immediately, `setIntroDismissed` persists in the background and its failure is swallowed; is that an acceptable UX tradeoff, or should a failed persist re-show the banner / surface an error? Also check behavior for a non-admin HA user, since `frontend/get_user_data`/`set_user_data` aren't admin-gated like Home Keeper's own websocket commands.
- **HA best practices**: is reaching for HA core's own per-user frontend data store (rather than adding a Home Keeper-specific websocket command backed by `store.py`) the idiomatic choice here, and does the namespaced key `home_keeper_intro_dismissed` sufficiently avoid collisions with other integrations sharing that same per-user store?
- **Maintainability**: is loading `introDismissed` inside the big `Promise.all` in `_reload()` (alongside tasks/assets/options/etc.) the right home for it, or does bundling an unrelated UI-preference fetch with domain-data loading blur a boundary?

---

Issue #161, opened 2026-07-29
MCP server


mdarocha:
The data in home-keeper is very valuable - especially stuff like manuals, device list etc.

It'd be great if it could be integrated with an MCP server, allowing AI agents like Claude to answer questions regarding the devices and maintenance tasks.

https://github.com/homeassistant-ai/ha-mcp is a good template to base it on, maybe even if possible, just integrate with ha-mcp and provide additional MCP tools related to the devices.

A local RAG pipeline could run on the home assistant instance, extracting information from uploaded files (manuals, receipts etc.) to provide more data for agents.

prestomation:
That's interesting. My agent has a ha token and uses the rest and websocket APIs and has no problem with my hk-related requests. 

An mcp server is probably not the right shape here, but I've got some ideas on how to streamline this usecase. Thanks for the issue!

mdarocha:
I currently use ha-mcp with the Claude.ai web chat as a custom connector, and it's been working pretty nicely. Note that more limited harnesses like those can't call REST APIs as easily as MCP tools, so I think it still makes some sense to have.

mdarocha:
Seems like after futher testing, a lot of home-keeper information is stored in a way that's easily accessible using standard ha-mcp tools, and any quirks can be papered over with a custom skill.

The only think that remains a pain point that a custom MCP would improve is the custom files - having dedicated tools to access them and search though them using RAG would be a nice thing to have.

prestomation:
I'm going to close this after merging #185 let me know if you have more to discuss

PR #185, opened 2026-08-08
feat: add sign_document_url/sign_part_file_url services

mdarocha:
Related to #161.

## Problem

Issue #161 asks for MCP-agent access to Home Keeper's data, and specifically calls out that **file bytes** (manuals/receipts/warranty PDFs) are the one piece that's not already reachable through standard tooling — everything else (assets, tasks, metadata) is already accessible via services.

Today the only way to fetch a document's/part-file's actual bytes without an interactive HA login is:
- The authenticated `GET /api/home_keeper/document/{asset_id}/{document_id}` view (needs a normal HA auth session/bearer token), or
- The `sign_document_url`/`sign_part_file_url` **websocket** commands, which mint a signed URL via `async_sign_path(hass, path, ttl, refresh_token_id=connection.refresh_token_id)` — signed using **the calling connection's own identity**.

That second mechanism is exactly right for a real browser tab (the connection's `refresh_token_id` is the user's own valid session), but it doesn't generalize to a caller with no browser session or websocket connection of its own — e.g. an MCP server relaying a signed-URL request on behalf of a connected agent, which is typically authenticated to Home Assistant through its own connection (possibly a proxied/system one), not a normal per-user browser session. There's no service equivalent at all today, so a non-interactive caller has no way to reach file bytes, period.

## Fix

Two new services, `sign_document_url` and `sign_part_file_url`, mint the same kind of short-lived signed URL as the existing websocket commands, but as a first-class service (per the architecture doc's "every action ⇒ service first" rule) — callable via `hass.services.async_call(..., return_response=True)`, which is exactly how an MCP server surfaces Home Assistant services to a connected agent.

Both new services — and the *existing* websocket commands — now delegate to one shared implementation (`manuals.async_sign_document_url` / `async_sign_part_file_url`), per the architecture doc's "never a divergent code path" rule. That shared helper deliberately signs with **no caller-specific identity** (no `refresh_token_id` passed to `async_sign_path`), so it falls back to Home Assistant's own built-in, always-present, read-only **"Home Assistant Content"** system user. This is the same mechanism HA core itself uses for externally-fetchable signed URLs (camera/media proxies) — it's designed for exactly this "any client with the URL can fetch it, no auth header" use case, and doesn't depend on the caller having a "normal" interactive session at all.

**Behavior note:** this changes how the *existing* websocket commands sign their URL too (content-user identity instead of the calling browser connection's own token). The URL still needs no interactive auth to open either way, so this should be unobservable from the frontend/panel — flagging it explicitly since it touches already-shipped, working code and I want it to be an easy thing to push back on if there's a reason to keep the two paths signing differently.

## Testing

- `tests/integration/test_documents.py::test_sign_document_url_downloads_without_auth_header` / `test_sign_document_url_unknown_document_errors` (and the `test_part_files.py` siblings) — the integration suite proves the actual property this PR delivers: the signed URL from the new service downloads successfully with **zero** `Authorization` header.
- `services.yaml` + `strings.json` entries with full parity across all 16 locales (reused already-translated field text for the `asset_id`/`document_id`/`part_id` fields shared with sibling services like `remove_asset_document`/`remove_part_file`; translated the handful of genuinely new strings — service names/descriptions and the two new field descriptions).
- `CHANGELOG.md` entry under `[Unreleased]`.

**Update: ran `tests/integration` for real.** No Docker daemon in my environment, but `podman` + the `docker-compose` CLI plugin work as a drop-in (`podman compose up -d`, `--security-opt label=disable` to work around a rootless-Podman/SELinux mount-labeling gap that isn't a factor under real Docker). Against a freshly-onboarded container:

- **`ci/test-python-integration.sh`'s actual suite, full run: 93/93 passed**, including both new tests and their not-found-error siblings.
- Caught and fixed a real bug in my *own test* along the way: `get_url(hass)` resolves to whatever address HA considers itself externally reachable at (here, the container's internal network IP, e.g. `10.0.2.100`) — not necessarily the address a given client reaches HA over. My first assertion (`signed_url.startswith(HA_URL)`) was too strict and failed before ever reaching the download step. Fixed the test to check the URL's structure (scheme/path/`authSig` query) and then fetch through the address the test suite is actually configured to reach HA on, keeping the signed path/query — the part actually under test — untouched.
- Manually confirmed the literal end-to-end claim this PR makes, byte-for-byte, before fixing the test: `curl`ing the exact signed URL the service returned, with **zero** `Authorization` header, returned the uploaded PDF's exact 45 bytes back.

Also still clean from before (unchanged by the above): `mypy`, `ruff check`/`format --check`, `python -m py_compile`, and `pytest tests/ --ignore=tests/integration` (692 passed, 1 skipped) / `test_translations_parity.py` (80/80) in the same environment CI's unit-test job uses.


custom_components/home_keeper/manuals.py
```diff
@@ -288,6 +302,54 @@ def _part_with_file(asset: dict[str, Any] | None, part_id: str) -> dict | None:
...
+    session's does).
+    """
+    coord = _coordinator(hass)
+    document = _file_document(
+        coord.store.get_asset(asset_id) if coord else None, document_id
+    )
+    if document is None:
+        return None
+    path = document_path(asset_id, document_id)
+    return async_sign_path(hass, path, DOCUMENT_URL_TTL)
```
prestomation: I don't think omitting `refresh_token_id` gets you the content user — looking at `async_sign_path`, that's only reached via `use_content_user=True` or as the last fallback, after it tries the current websocket connection and then the current HTTP request. So the websocket commands are probably signing exactly as before, and the service picks its identity from whatever context it happens to be called in (which differs across the in-process, remote, and standalone ha-mcp deployments).

What do you think about passing `use_content_user=True` explicitly? It makes the behavior the same everywhere, and it's what this docstring already describes.

---

custom_components/home_keeper/manuals.py
```diff
@@ -60,15 +62,27 @@
...
 # a long time, and deleting a live one is far worse than keeping a stray for a day.
 _TEMP_MAX_AGE_S = 24 * 60 * 60
+# How long a signed document/part-file URL stays valid. The dashboard card pre-signs
+# file documents and embeds the URL as a plain <a href> (so a tap opens natively —
+# the iOS app's WKWebView blocks an async window.open), so the URL must outlive a
+# reasonably idle dashboard, not just a click; the card re-signs well before this on
+# refresh. Also the TTL for URLs minted by the sign_document_url/sign_part_file_url
+# *services* (issue #161), for a non-browser caller — e.g. an MCP-connected agent —
+# to fetch the file after receiving the service response.
+DOCUMENT_URL_TTL = timedelta(hours=1)
```
prestomation, line 70: The 1 hour here exists for the dashboard, which embeds the URL in a static `<a href>` that has to survive an idle page. An agent fetching bytes right after the service response doesn't need anywhere near that, and since the URL needs no auth header it ends up sitting in model-provider logs. Any objection to a separate ~15 minute TTL for the service path?

---

custom_components/home_keeper/__init__.py
```diff
@@ -1002,6 +1019,47 @@ async def handle_update_asset_document(call: ServiceCall) -> None:
...
+            hass, call.data["asset_id"], call.data["document_id"]
+        )
+        if signed is None:
+            raise ServiceValidationError(
+                translation_domain=DOMAIN,
+                translation_key="unknown_document",
+                translation_placeholders={"document_id": call.data["document_id"]},
+            )
+        return {
+            "url": f"{get_url(hass)}{signed}",
```
prestomation: Per [Getting the instance URL](https://developers.home-assistant.io/docs/instance_url/), "By default, we prefer internal URLs over external ones" — so I'd expect this to hand back an internal URL that Claude.ai's servers can't reach in your setup. Did the URL actually work when you tried it from the agent, or are you seeing different behavior than I'd expect here?

---

custom_components/home_keeper/__init__.py
```diff
@@ -1002,6 +1019,47 @@ async def handle_update_asset_document(call: ServiceCall) -> None:
...
+            "expires_in": int(manuals.DOCUMENT_URL_TTL.total_seconds()),
```
prestomation: `get_url` also raises `NoURLAvailableError` if nothing is configured, which would propagate unhandled. We target Platinum, so user-facing exceptions need a localized message — could we catch it and raise with a translation key like the `unknown_document` one just above?

---

custom_components/home_keeper/manuals.py
mdarocha: Confirmed, you're right, and this was a real bug, not just a hypothetical. `async_sign_path` only reaches `hass.data[STORAGE_KEY]` (the content user) as the LAST fallback, after checking `websocket_api.current_connection.get()` and then `current_request.get()` — omitting `refresh_token_id` doesn't make it identity-independent, it just makes it pick up whatever ambient context happens to be live. Fixed in 5647a48: both signing helpers now pass `use_content_user=True` explicitly. Side effect worth calling out: this also changes the *existing* websocket commands to genuinely sign as the content user now too — previously they'd have hit the `current_connection` branch and kept using the browser session's own token, which is what my original PR description claimed was happening but, per your catch, wasn't actually implemented that way.

custom_components/home_keeper/manuals.py
mdarocha: Agreed, and done in 5647a48: added a separate `SERVICE_DOCUMENT_URL_TTL = timedelta(minutes=15)`, and an optional `ttl` param on both signing helpers (default stays `DOCUMENT_URL_TTL`/1h, so the websocket commands are unaffected). The two services now pass the shorter one. Verified against a live container: the service response's `expires_in` now asserts exactly `900`, not just `> 0`.

custom_components/home_keeper/__init__.py
mdarocha: Good catch — I hadn't actually tested from an external MCP client (Claude.ai etc.), only from the same Docker network as the test HA instance, where `get_url()`'s internal-network IP happened to be reachable either way. That masked exactly the gap you're describing. Fixed in 5647a48: added `_instance_base_url(hass)`, which calls `get_url(hass, prefer_external=True)`. This only nudges the preference — an instance with no external URL configured keeps behaving exactly as before (internal URL, unreachable off-network, same as today), it's specifically for instances that *do* have one configured, which is presumably most setups intending to let a cloud-hosted agent reach them at all (since the MCP webhook itself needs the same external reachability).

custom_components/home_keeper/__init__.py
mdarocha: Done in 5647a48: `_instance_base_url` (added for the sibling comment on this thread) catches `NoURLAvailableError` and raises `ServiceValidationError` with a new `no_instance_url` translation key, added to `strings.json` and all 16 locale translation files (English-first, matching the existing `exceptions.*` convention — key/placeholder parity is enforced by `test_translations_parity.py`, full translation happens incrementally per that same convention).

mdarocha:
Pushed `5647a48` addressing all 4 review comments:

1. **Identity resolution bug (real, not hypothetical)** — confirmed `async_sign_path` only reaches the content user as a last-resort fallback after checking the live websocket connection and then the live HTTP request, so omitting `refresh_token_id` wasn't actually identity-independent. Now passes `use_content_user=True` explicitly in both signing helpers.
2. **Separate service TTL** — added `SERVICE_DOCUMENT_URL_TTL` (15 min) via an optional `ttl` param, default unchanged for the websocket/dashboard path (1h).
3. **`prefer_external=True`** — added `_instance_base_url(hass)`; degrades to today's behavior when no external URL is configured.
4. **`NoURLAvailableError` handling** — caught and re-raised as `ServiceValidationError` with a new `no_instance_url` translation key (all 16 locales, English-first per the existing `exceptions.*` convention).

Replied to and resolved all 4 threads. Also fixed the CHANGELOG em-dash/vale CI failure from the prior push.

Re-ran the full Docker integration suite locally (podman) against a fresh container after these changes: 93/93 passed, including the service TTL now asserting exactly 900s (not just `> 0`) to actually prove fix #2 took effect.

This push's CI is sitting at `action_required` (fork-PR workflow approval gate — confirmed via API, I only have `pull` permission on this repo). Whenever you get a chance.

prestomation:
Thanks! Have you successfully tested this with your claude connector setup using ha-mcp? Happy to merge this when we confirm that

mdarocha:
Did the live testing — installed this branch on my own running Home Assistant instance via HACS (added the fork as a custom repository, built a release zip the same way `release.yml` does: `npm ci` + `ci/build-panel.sh` then `ci/build-zip.sh`, published it as a prerelease on my fork so HACS had something to install, then restarted).

Verified:
- Integration loaded cleanly post-restart: `home_keeper` config entry state `loaded`, zero new errors in the log, zero repair issues.
- Both `sign_document_url` and `sign_part_file_url` show up registered in the service list.
- Called `sign_document_url` for real against an actual uploaded manual on an existing appliance. Response came back with an **external** URL (not an internal network address) and `expires_in: 900` — confirming both the `prefer_external=True` and the shorter 15-minute service TTL took effect live, not just in the Docker suite.
- Fetched that exact signed URL with a plain `curl`, **zero `Authorization` header** — got back `200`, the correct `application/pdf` content type, and the exact byte count matching the stored file (a genuine, valid multi-page PDF, not an error page). This is the core fix (`use_content_user=True`) proven end-to-end in a real production instance: the URL was signed by ha-mcp's own service-call context (not a browser session) and still worked when fetched externally.
- `sign_part_file_url` — no part on my instance currently has an attached file to exercise the same live round-trip, but it's the identical shared implementation (`manuals.async_sign_part_file_url`) and is covered by the Docker integration suite's dedicated test, which passed.

No regressions observed elsewhere in Home Keeper (existing assets/tasks/documents all still load and behave normally).

prestomation:
The point of this is for tools like ha-mcp to be able to discover and use. It would made me feel better if I had your confirmation you had success with your actual usecase

mdarocha:
> The point of this is for tools like ha-mcp to be able to discover and use. It would made me feel better if I had your confirmation you had success with your actual usecase 

Yeah, the previous comment was from a test I ran with my agent - it worked the exact way I wanted it to (used ha-mcp to call the service, then downloaded the file)

prestomation:
Just curious, are you using opus 5?

mdarocha:
> Just curious, are you using opus 5?

Nope, right now I'm testing Sonnet 5 for most workflows, trying to conserve the tokens :)

PR #187, opened 2026-08-10
Usage tasks come due on hours *or* time, whichever comes first

prestomation:
## Why

[`BambamNZ/maintenance-tracker`](https://github.com/BambamNZ/maintenance-tracker) is a third-party integration built to do something Home Keeper could not say: **"every 300 hours *or* every 6 months, whichever comes first."** Its author's worked example is a Bambu Lab printer nozzle metered off a cumulative hours sensor — squarely our problem space, which makes it a useful requirements document rather than just a competitor.

Reading it turned up three real gaps. Two are closed here; the third is the companion PR in [`ha-home-keeper-bambu-lab`](https://github.com/prestomation/ha-home-keeper-bambu-lab).

The full teardown — parity matrix, what's still missing, and the use cases each remaining primitive would unlock — is in the new **`docs/USAGE_MAINTENANCE_GAP_ANALYSIS.md`**.

### 1. A time backstop on usage tasks

A usage binding takes an optional `also_every: {interval, unit}` and a `combinator`:

```yaml
sensor:
  entity_id: sensor.x1c_total_usage_hours
  mode: usage
  target: 300
  unit: h
  also_every: { interval: 6, unit: months }
  combinator: any     # "any" = whichever comes first (default) · "all" = both must be met
```

`all` is the "no earlier than" case — a standby generator you want run *at least* monthly *and* serviced after 100 engine hours.

Two design decisions worth reviewing closely:

- **The clock anchors to `last_completed`, falling back to the task's `created`** — deliberately *not* to the meter baseline. A meter reset (replaced controller, rolled-over counter) is bookkeeping, not a service, and must not push the calendar half out.
- **The watcher used to `continue` on a missing reading.** That would have silenced the backstop in exactly the situation it exists for: a machine idle or unplugged long enough to escape its calendar service. A usage task carrying `also_every` is now evaluated with `reading=None` — the meter half can't be met, and the baseline is left alone.

### 2. Progress you can read and automate against

`sensor.unit` labels the target (the panel prefills it from the bound entity, so "300" reads as "300 h"), the task page draws a meter bar with a "180 h to go" line, and a **device-attached** task's next-due sensor now carries `usage_consumed` / `usage_remaining` / `usage_percent` / `usage_target` / `usage_unit` / `backstop_due`.

Also fixed in passing: the dashboard card's recurrence filter was missing `one-off` and `sensor` entirely.

### 3. `home_keeper.set_task_meter`

Re-anchors a usage baseline **without** recording a completion — for work done before Home Keeper was watching, or a meter that was replaced or zeroed. Omit `baseline` to anchor to the live reading. Fires `task_updated`, unlike the watcher's own silent baseline bookkeeping.

### Correction to `IDEAS.md`

It claimed no cumulative counter exists for Bambu Lab printers. `sensor.<printer>_total_usage_hours` does — `TOTAL_INCREASING`, `DURATION`/hours, diagnostic, no per-model `exists_fn`. It counts *printer usage* hours rather than strictly print hours, which is close enough for a service interval and is what Bambu's own duty-cycle guidance is written against.

## Screenshots

A usage task in flight — the meter bar, the "to go" line, and the backstop note:

Task detail for a usage task: recurrence reads Every 300 h of use, or every 6 months; a progress bar at 40% with 180 h to go

The create form with the backstop fields and the live hint:

Sensor task form showing Target, Unit label, Or every, Backstop unit and Combine with fields

The walkthrough tour was extended to step through building a usage-or-time task and to linger on the filled meter bar, so the regenerated CI comment should show both.

## Testing

Run locally, all green: **717** pure unit tests (incl. a new backstop matrix in `test_sensor_tasks.py` and binding validation in `test_models_sensor.py`) · **321** vitest · **97** Docker integration tests, including new cases for time-only arming with the bound entity unavailable, the `all` combinator refusing to arm on the meter alone, `set_task_meter`, and the usage attributes · **44** Playwright specs · `ruff check` + `ruff format` + `tsc --noEmit` clean.

**`mypy` could not be verified locally.** This environment's package mirror caps `homeassistant` at 2024.3.3, which predates HA shipping `py.typed`, so every import resolves as `import-untyped` and the run is 121 errors of noise. Filtering those leaves only two pre-existing findings in files this PR doesn't touch. **CI's `lint.yml` is the first real mypy signal here — worth checking that job before merge.**

## Notes for review

- Version bumped to **`0.12.0b1`** (the post-0.11.0 beta bump hadn't happened yet, so the existing `Unreleased` content rolls into this section).
- Forward compatibility: `normalize_sensor` builds its result from known keys only, so an older Home Keeper silently drops `also_every`/`unit` and the task still works as a plain meter. That's what lets the Bambu glue send them unconditionally instead of version-gating — documented in `INTEGRATING.md`.
- The screenshot diff is large because every capture re-renders (the version footer changed and the seed gained a demo usage task). The two stacked toast overlays in `15-panel-monitored-section.png` are a pre-existing capture artifact, present at HEAD too.

https://claude.ai/code/session_01RRnPQqVuzmZhMDSkTV2R9R

---
_Generated by [Claude Code](https://claude.ai/code/session_01RRnPQqVuzmZhMDSkTV2R9R)_


prestomation:
/q review Be critical and skeptical, and lead with the most serious problems — but don't withhold the minor ones either. This adds a *time backstop* to usage (meter) sensor tasks, so please concentrate there:

**Correctness.** `sensor_tasks.backstop_due()` anchors to `last_completed`, falling back to the task's `created`, and calls `recurrence.add_interval`. Attack the date arithmetic: DST transitions on a `days`/`weeks` interval, month-end clamping (Jan 31 + 1 month), a `last_completed` in a different timezone from `now`, and the inclusive `now >= due_at` boundary. In `evaluate_usage`, check the interaction between the meter-reset debounce and the new time branch — can a single evaluation both re-baseline and arm, or drop a pending `reset_candidate` it should have kept? Is the `combinator: "all"` semantics right when a task has never been completed (anchor = `created`)?

**The unavailable-reading path.** `sensor_watcher.async_evaluate` now calls `_evaluate_usage(reading=None)` for any usage task carrying `also_every`. Is there a path where a `None` reading corrupts or clears a stored baseline, or where a task with no baseline yet and no reading gets stuck permanently un-armable?

**Migration / back-compat.** Existing usage tasks have no `also_every` and possibly no `created`. Does anything here regress them? `models.build_task` now applies a seeded `last_completed` to sensor tasks (it used to ignore it) — can that arm a task that should be born dormant, or corrupt `completions[]`?

**Maintainability.** `sensor_tasks` now imports `recurrence` (both pure, no cycle) and reaches for `recurrence._parse`, matching `transitions.py`'s existing precedent — is that the right call or should `_parse` be promoted? Is `also_every` the right home for this versus top-level `interval`/`unit`?

**Performance.** The watcher's bound-entity set and the per-tick evaluation loop — any new N+1 or redundant work now that unavailable-reading tasks are no longer skipped?

**HA best practices.** The new `set_task_meter` service (schema, error surfacing, `ServiceValidationError` translation keys), firing `task_updated` from `store.set_sensor_baseline(silent=False)`, and putting `usage_*` figures as attributes on the existing next-due sensor rather than adding new entities.

---

prestomation:
/q review Please be critical and skeptical, lead with the most serious problems, and don't withhold minor ones. New in 8908426: the task form now renders the recurrence rule as one plain sentence directly above the save button, live.

**The reuse decision.** `formRecurrenceSummary` builds the preview by running the live edit state through `buildTaskPayload` and handing the result to `recurrenceSummary` — deliberately the same two functions that produce the saved payload and the card's caption, so the preview can't drift from reality. Attack that: is running a *partial* form through the real payload builder safe on every path? `buildTaskPayload` coerces (`Math.max(1, Number(interval) || 1)`, `haDateTimeToIso(task.due) || new Date().toISOString()`, `cardLinksFromTokens`) — can any of that throw, allocate, or produce a summary that misrepresents what will actually be saved? The triggered case already bit us (the payload omits `recurrence_type`, so the preview read "every day" until we re-attached it) — are there other fields the payload legitimately drops or rewrites where the preview would then lie?

**Correctness of the copy itself.** Is "Every 100 h of use, or every 6 months" an accurate reading of `combinator: any` given the backstop anchors on `last_completed` falling back to `created`? And is the `all` variant's "and every 6 months" right, or does it under-sell "both must be met"? A summary that is subtly wrong is worse than no summary.

**Render/update path.** The strip is refreshed by `_updateFormHints` from the form's `value-changed` handler, in place, to avoid stealing focus from the input being typed in. I widened that branch from sensor-only to every recurrence type — check for a case where the schema *does* change but we now take the in-place branch (or vice versa) and the strip goes stale, and whether reading `closest('.hk-form-summary')` off the value span is fragile. Any XSS surface: the label goes through `escapeHTML` into `innerHTML`, the value is set via `textContent` — is that split right?

**Performance.** This now runs `buildTaskPayload` + `recurrenceSummary` on every keystroke in the form. Real cost or noise?

**i18n.** `form.summary.label` was added to all 16 locales. The summary composes strings that were written to be read *after* a noun ("every 3 months after completion") under a label — does that read as broken grammar in any of them, and is the sentence-case mismatch between the sensor variants ("Every…") and the clock variants ("every…") a bug?

**Test quality.** `sensor-form.test.js` asserts `formRecurrenceSummary(form) === recurrenceSummary(buildTaskPayload(form))` — is that a tautology that would pass even if both were wrong, and what would you assert instead?

**Harness fix.** `expandGroup()` now guards the `<details>` toggle in the screenshot capture (step 14b opened the Monitored group and step 15 blindly clicked it shut). Is a guarded toggle the right fix, or should the harness reset group state between steps?

---

prestomation:
/q review

(Re-posting — the previous request got a "which comment should I reply to?" response rather than a review.)

Be critical and skeptical, lead with the most serious problems, and don't withhold the minor ones. The new work in `8908426`..`f6df1c3` is a **live plain-language summary of the recurrence rule in the task form**, rendered directly above the save button. Concentrate there.

**The reuse decision, which is the load-bearing one.** `forms.formRecurrenceSummary` builds the preview by running the live edit state through `buildTaskPayload` and handing the result to `utils.recurrenceSummary` — deliberately the same two functions that produce the saved payload and the card's caption, so the preview cannot drift from what the task becomes. Attack that: is feeding a *partially filled* form to the real payload builder safe on every path? It coerces (`Math.max(1, Number(interval) || 1)`, `haDateTimeToIso(task.due) || new Date().toISOString()`, `cardLinksFromTokens`) — can any of that throw, mutate the caller's object, or yield a sentence that misstates what will actually be saved? One case already bit us: the payload deliberately omits `recurrence_type` for triggered tasks, so the preview read *"every day"* until we re-attached it. Are there other fields the payload legitimately drops or rewrites where the preview would then lie?

**Is the copy itself true?** Does "Every 100 h of use, or every 6 months" correctly describe `combinator: any` given the backstop anchors on `last_completed` falling back to `created`? Does "and every 6 months" fairly describe `all`? A summary that is subtly wrong is worse than none.

**The update path.** The strip refreshes via `_updateFormHints` from the form's `value-changed` handler, in place, to avoid stealing focus from the field being typed in. I widened that branch from sensor-only to *every* recurrence type — look for a case where the schema does change but we now take the in-place branch (or the reverse) and the strip goes stale. Is `closest('.hk-form-summary')` off the value span a fragile way to find the container? On escaping: the label goes through `escapeHTML` into `innerHTML`, the value via `textContent` — is that split correct, and is there any path where user text reaches `innerHTML` unescaped?

**Cost.** This runs `buildTaskPayload` + `recurrenceSummary` on every keystroke in the form. Real or noise?

**i18n.** `form.summary.label` was added to all 16 locales. The summary composes strings written to be read *after a noun* ("every 3 months after completion") beneath a label — does that read as broken grammar anywhere? And is the sentence-case mismatch between the sensor variants ("Every…") and the clock variants ("every…") a bug worth fixing?

**Test quality.** `sensor-form.test.js` asserts `formRecurrenceSummary(form) === recurrenceSummary(buildTaskPayload(form))`. Is that a tautology that would pass even if both sides were wrong, and what would you assert instead?

**Harness fix.** `expandGroup()` now guards the `<details>` toggle in the screenshot capture, because step 14b expanded the Monitored group and step 15 blindly clicked it shut, capturing a collapsed group. Is a guarded toggle right, or should the harness reset group state between steps?

---

prestomation:
/q review

Be critical and skeptical, lead with the most serious problems, don't withhold minor ones. New in `fb86719`, two form changes on top of the summary strip:

**1. The time backstop moved behind a switch** (`sensor_backstop_on`), replacing the old "interval of 0 means off" convention. The switch is authoritative, and one predicate — `backstopEnabled()` — answers "is the backstop on?" for the schema, the payload and the hint, reading either the flat form flag or a loaded task's `sensor.also_every`.

Attack the round-trip and the precedence. Flipping the switch off only *hides* `sensor_also_every`; it doesn't clear it, so a stale interval survives in the edit state — is `backstopEnabled()` genuinely consulted on every path that could resurrect it? Is inferring the switch from `task.sensor?.also_every` right for a task loaded mid-edit, or is there a sequence (load a backstop task → switch off → change something else → the schema rebuilds from the task) where the inference overrides the user's explicit off? Is `flag !== undefined && flag !== null` the right test, or does some path put `''`/`0` in there and flip the meaning?

**2. Seeding on enable.** Switching it on writes `DEFAULT_BACKSTOP_INTERVAL` (6) into `sensor_also_every` when it's blank or zero, from the panel's `value-changed` handler, so the revealed fields describe a real rule. Is mutating edit state from that handler sound, or can it fight the form's own value propagation? Does the guard `!(Number(...) > 0)` behave for `NaN`, `''`, `'6'`?

**3. The two info panels became one box.** `_updateFormHints` now takes an optional container so first paint and subsequent keystrokes go through identical code. Check the lifecycle: is passing the not-yet-attached element sound, does the lookup fall back correctly after a re-render, and can the box end up visible-but-empty or hidden-with-content?

**Migration.** Tasks created before this change carry `also_every` and no flag; tasks that are pure meters carry neither. Both must open correctly and save unchanged if untouched. Is there any stored shape that opens wrong?

**Harness.** `setBackstop()` drives the switch by asserting the resulting number-selector count rather than reading the widget, and clicks `ha-switch` rather than the wrapper (a centred click on the wrapper hits dead space). Is counting selectors too coupled to the schema's shape?

**Tests.** I updated the existing backstop fixtures to set the new flag, which is a contract change — are the added cases (switch off with a stale interval, inference from a loaded task, seeded interval) the right ones, and what's still uncovered?

---

Issue #183, opened 2026-08-08
Issues after upgrading Home Assistant to 2026.8


mdarocha:
2026.8 introduced splitting up devices by integrations, no longer merging them together.

This is mostly fine (but means that home-keeper entities no longer show up in another integration's device page, which maybe should be documented), but introduces several small issues:

1. GUIDs show up for devices instead of names, and a re-selection is required for the primary device.

This causes a bit of confusion regarding devices that have a "real" device from an integration, and a home-keeper "virtual" device.

It seems like it'd be a lot cleaner to get rid of the virtual/real split and always create a "virtual" home-keeper backing device for each entry, with the ability to set other integrations as related using existing functionalities

2. Some tasks that are auto-created from problem entities are separated when grouping by device

When the "main" device is set to the home-keeper virtual one, automated tasks from other integrations show up as separate entries when grouping by devices

<img width="858" height="285" alt="Image" src="https://github.com/user-attachments/assets/e917f40f-d5d3-4cf4-b218-6b9e5b03f344" />

3. Battery keeper issues

Battery-keeper created tasks have the before mentioned issue when it points to a non-existing GUID device, but there's no easy way to fix it.

<img width="858" height="285" alt="Image" src="https://github.com/user-attachments/assets/e503c9d3-4a5a-4e28-acd6-758589ad1b2e" />

prestomation:
Thanks for letting me know! I usually don't update until.1 or .2. I'll plan to get a beta release out so we can try it out.

prestomation:
Well I didn't end up releasing a beta 🫠, but please let me know if you have feedback on 0.13

mdarocha:
It seems like to most visible issues were resolved, so I can close this

PR #189, opened 2026-08-11
Link task entities to their device instead of forking a duplicate (#183)

prestomation:
Fixes the HA 2026.8 device-attachment break, repairs installs that already upgraded, and adds the test apparatus that should have caught it. Bumps to `0.13.0b1`; `0.13.0.dev189` is installable via HACS now.

## The fix (option A)

HA 2026.8 made device identifiers unique per config entry, so Home Keeper's identifier copy stopped merging onto an existing device and instead **forked a duplicate** — one owned by us, with no `name`, which is why #183 saw raw GUIDs.

Per-task entities now set `entity.device_entry` and leave `device_info` unset, so they land on the target device without Home Keeper claiming it. The self-owned `DeviceInfo` fallback stays for a task whose `device_id` doesn't resolve.

**Accepted cost, chosen deliberately:** HA sources device triggers and per-device diagnostics from `device.config_entries`, so neither is offered on a device we merely link to. Most usage goes through the panel or a Lovelace card, and the same automations remain buildable from the task's entities or the `home_keeper_*` event. Appliances Home Keeper owns keep both. README, `EVENTS.md` and the migration guide all say so.

Options B (always own a device) and C (hybrid) are written up with costs in `docs/DEVICE_REGISTRY_2026_8_PLAN.md`.

## Every upgrade path now ends up correct

Three orders, each its own container path in `tests/upgrade/test_upgrade_order.py`, plus a dedicated repair path:

| order | outcome |
|---|---|
| Home Keeper first, then HA | clean — nothing to repair |
| HA first, Home Keeper after | repaired at setup |
| both together | repaired at setup |

Two mechanisms, because prevention only reaches one of them:

**`devices.async_detach_legacy_merged_devices`** drops our config entry from a device an earlier release merged onto. Run while still on HA 2026.7, it leaves nothing of ours joined to those devices, so HA's split has nothing to divide. Limited to devices with another config entry besides ours — removing the last entry deletes the device and would strand the entities on it.

**`devices.async_heal_split_device_ids`** repairs what prevention can't reach. The mechanism is worth knowing, because the obvious approach fails silently: **a stale id doesn't look stale.** `dr.async_get()` synthesizes a read-only composite for a pre-migration id, so "does this still resolve?" answers yes and a naive heal skips every task — which is exactly what my first attempt did. What HA refuses is *linking an entity* to it:

> Ignoring request to link entity from integration home_keeper to device …: the device id refers to a composite device which was split into one device per config entry

The supported query is `async_get_devices_for_composite_device_id`; the successor to adopt is the split that isn't ours, with the composite's former `primary_config_entry` breaking ties.

**`store.async_merge_split_duplicates`** cleans up the contributor tasks the broken references caused. Same insight in reverse: the two copies point at different halves of the same original, so they look unrelated until device ids are canonicalized back through `composite_device_id`. The survivor keeps the history but adopts the **newer** task's `device_id` — the newer one was created by the contributor *after* the split, so it carries that contributor's own answer for where the task belongs; adopting anything else gets it duplicated again on the next reconcile.

**Sharpest edge, flagged for review:** the merge deletes with `force=True`, bypassing deletion protection. That protection exists to stop a *user* deleting something their integration will recreate; here the integration created the duplicate and cannot clean it up itself. Guarded by identical `source` payload after canonicalization, and by never removing a task carrying completions.

## Three harness bugs, each of which produced a confident wrong answer

Recording these because every one looked fine and wasn't:

1. The suite originally ran *this* Home Keeper in **both** phases — a journey no user takes. It made the fix look like it also cured the glue duplication.
2. The working-tree build was cached in a gitignored dir `ci/fetch-glues.sh` never cleared, so edits never reached the container and the suite reported on **stale code**. That is where "order doesn't matter" came from.
3. `_run_path` created its own temp dir, so a failure inside it left no logs and a running container. Fixing that is the only reason the composite-device traceback was findable.

## Test apparatus

- **`tests/integration/test_device_attach.py`** — the foreign-device contract. The battery-notes stub now owns a real device; nothing in the container previously did, which is why the break went unnoticed while CI already ran 2026.8.
- **`tests/upgrade/`** — real multi-phase upgrades against one config dir, generalized to a sequence of `(ha_tag, hk_build)` boots so orders can be compared, with the previous release fetched at a pinned tag for phase 1.
- **`ci/fetch-glues.sh`** — three real glues plus both Home Keeper builds at pinned SHAs. Upstreams (`battery_notes`, `bambu_lab`) are stubbed; the real ones need hardware or vendor cloud.
- **`.github/workflows/ha-beta.yml`** — nightly integration + e2e + upgrade + mypy against `beta`, filing one reused `ha-beta-regression` issue. Gates nothing.

**All eight `xfail(strict=True)` markers are gone.** Every one flipped to XPASS when the repair landed, which is the ratchet working as designed.

## Docs

**"Upgrading to Home Assistant 2026.8"** ([preview](https://prestomation.github.io/ha-home-keeper/pr-preview/pr-189/docs/guide/migration-2026-8)) — what an already-upgraded install sees, that the repair is automatic, and the device-trigger change. Linked from the CHANGELOG.

## Verification

```
tests/upgrade                   24 passed,  0 xfailed
tests/integration               99 passed,  0 xfailed
tests/unit                     714 passed,  1 skipped
frontend (vitest)              333 passed
ruff · vale                    clean
```


prestomation:
/q review Please be critical and skeptical — this PR is entirely test infrastructure, so a test that passes for the wrong reason is the main risk. Surface the most serious issues first, and don't withhold minor ones.

**Correctness of the xfail ratchet.** Five tests are `xfail(strict=True)`. For each, is it xfailing for the reason its docstring claims, or could it xfail for an unrelated reason (fixture never seeded, device lookup returning None, a service call silently failing) and still look like a faithful record of #183? Are any assertions weak enough that a *partial* fix would flip them to XPASS while the real bug persists? Specifically check `test_device_attach.py::test_task_entities_land_on_the_foreign_device` and `tests/upgrade/test_upgrade.py::test_stored_device_ids_still_resolve`.

**Two-phase upgrade harness (`tests/upgrade/conftest.py`).** This is the riskiest code. Is `docker compose down` without `-v` genuinely sufficient to preserve the bind-mounted config dir between the two boots, and is there any path where phase 2 silently starts from a fresh config (which would make every assertion vacuous)? Is `_wait_for_running`'s onboard-or-login branching correct on the second boot? Are the fixed `SETTLE_SECONDS = 20` sleeps a race — could a slow runner snapshot a half-built registry and produce a misleading before/after diff? Is the `finally` teardown leak-proof if a boot fails midway, and does the port choice (8124) actually avoid colliding with the integration container?

**Fixture fidelity.** `tests/upgrade/stubs/{battery_notes,bambu_lab}` stand in for real integrations, and the whole result rests on them reproducing the predicates the real glues match on (`entity.platform`, `device_class`, `unique_id` suffixes, state attributes). Are they faithful enough that conclusions drawn about the real glues actually transfer? The `battery_notes` stub deliberately copies a foreign device's identifiers to recreate the pre-split merge — is that reproducing the real integration's behaviour or an artefact I've invented that inflates the finding?

**Maintainability.** `ci/fetch-glues.sh` pins five... actually three repos to commit SHAs with no automated bump path — is that the right call versus branch refs, and how will anyone notice the pins have gone stale? Is `ha_registry.py` the right home for the lifted websocket helpers, and did the refactor of `test_device_cleanup.py` / `test_entity_cleanup.py` change their behaviour at all?

**CI (`ha-beta.yml`).** Is the find-or-create logic for the `ha-beta-regression` issue race-free and correctly scoped to `schedule` runs only? Does `if: failure()` on the `report` job actually fire when any of the four needed jobs fails, including cancellation/timeout? Any concern about the nightly's cost or about `--pre homeassistant` breaking the mypy job for reasons unrelated to Home Keeper?

**Anything I've missed** about whether this suite would actually have caught #183 before release — that's its entire justification, so please push back if it wouldn't have.

---

prestomation:
Thanks — the review caught two things that genuinely mattered. Pushed in `1528dee`.

**First, a CI failure that was mine, not yours to find:** `ci/test-python-unit.sh` runs `pytest tests/ --ignore=tests/integration`, which now also collected `tests/upgrade` — a Docker suite needing staged fixtures. It passed locally only because Docker was up and the glues were staged. Both unit lanes now ignore it; the unit run is back to ~1s.

## Fixed

**#6 — phase 2 continuity.** The best catch in the review. A silent fresh-config start would have made every assertion in the suite vacuous while looking perfectly healthy. The fixture now fails loudly unless a **task id** survives across the two boots — ids are generated per creation, so a shared one proves continuity in a way names can't (a fresh seed would recreate identical names).

**#4 — `source.<ns>.device_id`.** Correct and the sharpest point. Added `test_source_namespace_device_ids_still_resolve` as a *separate* xfail rather than folding it in, precisely so that healing only the task field flips one green and leaves the other red. The two together are the real acceptance criterion for the auto-heal.

**#1 — settle race.** Replaced the fixed sleeps with polling until device/entity/task counts stop changing. Also 27s faster.

**#10 — xfail granularity.** Split into two tests, though the stated mechanism isn't right: `strict=True` only reports XPASS when *all* assertions pass, so a later failing assertion would have kept it xfail rather than falsely reporting success. The underlying point stands — a partial fix should be legible — so it's split anyway, and now reads as one XPASS plus one remaining xfail instead of an undifferentiated failure.

**#3 — fidelity.** The highest-risk item, so I verified it against the real source rather than reasoning about it. The stub is faithful to **pre-2026.8** Battery Notes, which is what phase 1 needs — that's the world users upgrade *from*, and the merged device it produces is what HA then splits. Battery Notes confirms this was its old shape in its own `common.py`: *"On older versions devices sharing an identifier or connection were merged into a single device."*

But your instinct was right that something was off: **current Battery Notes has already migrated.** 3.0.0-dev links via `async_entity_id_to_device_id` and no longer copies identifiers at all. So results here support claims about upgrading from a pre-2026.8 install — which is exactly what #183 reports — and must **not** be read as claims about Battery Notes today. Documented in the stub's docstring. Worth noting for the model decision: upstream has moved to entity-level linking and Home Keeper hasn't.

Also fixed a pre-existing fragility this surfaced: the integration conftest treated only `403` as "already onboarded", but HA returns `404` once the endpoint is gone. `tests/e2e/global-setup.ts` already tolerated both.

## Not changing, with reasons

**Hardcoded test credentials (CWE-798).** These provision a throwaway HA container the test creates and destroys itself; the "secret" is the password the test just set, on a container nobody else can reach. Moving it behind an env var with the same value as the default adds indirection without changing exposure. `tests/integration/conftest.py` and `tests/e2e/global-setup.ts` already do it this way, and diverging would be worse than the status quo.

**Repository URL validation in `fetch-glues.sh` (CWE-494).** The URLs are hardcoded defaults; the env override exists so a developer can test an unreleased glue branch. An attacker who controls this script's environment already has code execution on the machine — the validation would gate nothing they couldn't bypass. Integrity is handled the way that actually helps here: every repo is pinned to a **commit SHA**, so the fetched content is fixed regardless of what the remote does. Separately, the suggested patch's regex is `^`, which matches every string and so validates nothing.

**JSON error handling in `ws_command`.** If HA's websocket emits malformed JSON, a `JSONDecodeError` with the raw frame in the traceback is more useful than a `RuntimeError` that discards it. The assertion on the next line already handles the realistic failure (a well-formed error response).

**#9 — `ws_command` duplication.** The two take different things — the integration helper takes an authenticated `requests.Session`, the upgrade one takes a bare token, against a different port and lifecycle. Sharing them means a parameter-shuffling wrapper for ~10 lines. Left as is.

**#5 / #2 / #8 — issue-creation TOCTOU, port collision, pin staleness.** All fair, all low-severity. One nightly cron means two concurrent runs are near-impossible; `HA_PORT` already exists for the port; the SHA pins carry a comment explaining they're deliberate. Not worth code in this PR.

## On "would this have caught #183"

Agreed it would, with one sharpening: not because CI would have been on a newer HA — CI already ran `stable`, so it was on 2026.8 the whole time and stayed green. It's the *assertion* that was missing, which is why the PR adds both the test and the nightly rather than treating the version strategy as the fix.

---

prestomation:
/q review Two commits since your last pass added the actual behaviour change. Please be critical and skeptical, most serious issues first, and don't withhold minor ones.

**`devices.async_detach_legacy_merged_devices` is the riskiest new code.** It calls `async_update_device(remove_config_entry_id=...)` on devices an earlier release merged onto. Is the guard (`skip when we're the only config entry`) actually sufficient to prevent orphaning entities, or are there paths where the last entry is removed anyway — a race with `async_reconcile_assets`, a device whose other entry is being unloaded concurrently, or an entry removed between the iteration and the update? Is running it *before* `async_prune_orphaned_devices` in `async_setup_entry` the correct order, and is it idempotent across the reloads Home Keeper triggers on every task mutation? Could it detach from a device we legitimately want to stay on — an existing-kind asset, or a virtual asset device whose identifier check I've got wrong?

**Entity-level linking.** `entity.device_entry` is set in `HomeKeeperTaskEntity.__init__` with `device_info` left unset. Is that the right lifecycle hook — does a task re-attached to a *different* device actually move, given HA caches device association at add time and Home Keeper reloads the entry via `entity_set_key`? Any case where both `device_info` and `device_entry` end up set, or both `None`, and what happens then?

**Upgrade-order conclusions.** `tests/upgrade/test_upgrade_order.py` claims "update Home Keeper first is clean". Is that conclusion sound, or an artefact of the fixture? Specifically: does `OUR_FAULT` (`devices_we_own_but_did_not_create`) actually capture the damage users care about, or does asserting only on it hide real breakage the other two measures would have caught? Two earlier versions of this harness produced confident wrong answers (running the new code in both phases; a stale cached build) — please look for a third.

**Fixture fidelity.** Phase 1 now runs the previous release fetched at a pinned tag. Is swapping the mounted `custom_components/home_keeper` directory between boots a faithful stand-in for a HACS update, or does it miss something a real update does (storage migrations, `async_migrate_entry`, version bookkeeping)?

**Anything else** about whether the migration guide's advice is actually correct for a real user, since that is now published guidance rather than an internal note.

---

prestomation:
Thanks — the confirmations on the entity lifecycle, setup ordering and virtual-device guard are useful, since those were the three I couldn't check any other way. Pushed `9ff4708`.

## Taken, implemented differently

**#3 — the `OUR_FAULT` hole.** You're right: asserting only on the measure Home Keeper is answerable for means the other two could quietly worsen and the test would still pass.

I didn't use a hardcoded baseline, though. A magic number needs updating whenever the fixture changes, and it says nothing about the claim actually being made. Instead the orders are now compared **against each other**:

```python
for measure in hk_first:
    ours = len(hk_first[measure])
    for label, other in (("HA first", ha_first), ("both together", together)):
        assert ours <= len(other[measure]), ...
```

That encodes the published advice directly — "update Home Keeper first" is only honest if that order is no worse than the alternatives on *every* measure — and it self-maintains as the fixture evolves.

## Recorded rather than changed

**#1 — the TOCTOU.** Agreed on both the analysis and that there's no better API; the window is a few synchronous statements during our own setup. But it was only implicit before, so the docstring now says the check and the update aren't atomic and why the guard is the best available. A reader shouldn't have to rediscover that.

**#4 — fixture fidelity.** Also agreed, and worth being explicit about since it's a limit on what a green run means: swapping the mounted directory reproduces a HACS update of the *code*, not anything keyed on the version having changed (`.storage` schema bumps, `async_migrate_entry`, manifest bookkeeping). That caveat is now in the harness docstring.

## One correction

> HA prioritizes `device_info` (entity owns device)

Worth being precise, since it's load-bearing for this design: `entity_platform` only consults `entity.device_entry` when `device_info` is `None` — `device = entity.device_entry` sits in the `else` branch. So it isn't a priority ordering so much as `device_info` short-circuiting the check entirely. Same practical outcome here (exactly one is ever set, by construction), but if both were somehow set the `device_entry` would be silently overwritten rather than merely deprioritised, which is a more confusing failure than it sounds.

`11 passed, 8 xfailed` on the upgrade suite with the new guard, unit `714 passed`.

---

prestomation:
/q review Since your last pass this PR gained an automatic repair for installs that already upgraded, which is the riskiest thing in it — it mutates stored tasks and deletes some. Please be critical and skeptical, most serious first, and don't withhold minor ones.

**`store.async_merge_split_duplicates` deletes tasks with `force=True`, bypassing deletion protection.** That is the sharpest edge here. Is the guard set sufficient — identical `source` payload after canonicalizing device ids through `composite_device_id`, plus never deleting a task with completions? Can you construct a case where two *legitimately distinct* contributed tasks canonicalize to the same key and one is destroyed? Specifically: a contributor whose `source` payload is only `{"device_id": …}` (Battery Notes is exactly this) with two genuinely different batteries that were split from the same composite. Also, is deleting the survivor's peers while iterating `groups` safe given `delete_task` mutates `self._tasks`?

**Survivor selection.** The survivor keeps history but adopts the *newest* task's `device_id`. Is `max(by_age, key=lambda t: (len(completions), -by_age.index(t)))` correct — note `by_age.index(t)` compares dicts by `==`, not identity, so two tasks with identical content would resolve to the same index. Can that mis-select? What if *both* tasks have completions — we keep one and refuse to delete the other, leaving a duplicate behind; is that the right call or should it merge the completion lists?

**`devices.async_heal_split_device_ids`.** `_split_successor` picks the split that isn't ours, breaking ties on the composite's `primary_config_entry`, else sorted-by-id. Is that stable across restarts, and can it pick a device the user would consider wrong — e.g. the upstream's device rather than the one whose entity the contributor actually reads? Is running it before `async_config_entry_first_refresh` the right point, and is it genuinely idempotent once healed?

**Ordering and failure modes.** The heal runs at setup before platforms; the detach and prune run after. If the heal raises, `async_setup_entry` fails and the integration doesn't load at all — that already happened once in testing (a `TaskValidationError` escaping the merge). Should this whole repair be defensive so a bad edge case degrades instead of bricking setup?

**Anything I've missed** about data safety. This writes to `.storage/home_keeper` on every start where it finds something to fix, and users cannot undo it.

---

prestomation:
/q review Please be critical and skeptical, and put the most serious issues first — but don't withhold the minor ones.

Since your last pass this PR gained an automatic repair for installs that already upgraded to HA 2026.8, plus a fix for a fixture regression I introduced. Focus on:

**Correctness of the repair** (`devices.async_heal_split_device_ids`, `devices._split_successor`, `store.async_repoint_device_ids`, `store.async_merge_split_duplicates`):
- `_split_successor` picks one device out of a composite's splits — tie-break is the composite's `primary_config_entry`, then sorted-by-id. Is that stable and is the fallback defensible, or can it pick the wrong half?
- `async_merge_split_duplicates` groups on `f"{namespace}|{json.dumps(normalized, sort_keys=True)}|{task['name']}"` and keeps the survivor with the most completions, oldest-wins on a tie. Can that ever delete a task a user cared about? It refuses to delete anything with completions and uses `delete_task(..., force=True)` to get past `locked_fields` — is the force too broad?
- `async_repoint_device_ids` bypasses `models.merge_update` deliberately (that helper enforces `locked_fields`, which every glue sets on `device_id`). Is bypassing it the right call, and does it rewrite every place a device id is stored — `task["device_id"]` and `task["source"][ns]["device_id"]` — or is there a third?
- The whole repair is wrapped in `try/except Exception` in `__init__.py` so it can never block setup. Is swallowing broadly right here, or does it hide a state worth failing on?
- Idempotency: it runs on every setup. Second run on already-healed data must be a no-op.

**Ordering and races** — `async_detach_legacy_merged_devices` checks that another config entry remains before removing ours, and that check-then-update is not atomic. Is the window real in practice?

**Test honesty.** The upgrade harness in `tests/upgrade/` gave me three confidently wrong answers this session (running the working tree in both phases; a stale gitignored cache; a temp dir that ate its own logs on failure). Please attack it: does `ci/fetch-glues.sh` genuinely stage the *previous* release for phase 1 and the working tree for phase 2, and can any assertion in `test_upgrade_order.py` / `test_upgrade_repair.py` pass for a reason other than the one its docstring claims?

**HA best practices** for entity-level device linking (`entity.device_entry` with `device_info` unset) and for the device-registry API use.

---

prestomation:
Triaged in `7607542`. Taking three, declining two, and one of your questions had already been answered in the code.

**Taken**

1. **The survivor ranking.** You didn't flag this one, but re-reading `async_merge_split_duplicates` while checking your grouping question, `by_age.index(task)` was still in there. `list.index` compares with `==`, so two value-equal tasks resolve to the same position. Distinct task ids make it unreachable today, which is exactly the sort of invariant that stops holding quietly. Now ranks by `enumerate` position, which also drops the quadratic scan.
2. **The `_split_successor` fallback.** Comment added at the branch: a three-way-or-more split with no `primary_config_entry` is an arbitrary pick, sorted only so repeated setups don't flip-flop the task between devices. The two-way split this repair actually deals with never reaches that line.
3. **The "is there a third place?" question — the important one.** Checked all three glues rather than reasoning about it: bambu-lab and battery-notes both keep `device_id` at the top level of their `source` payload, and Pawsistant stores no device id at all (it keys on `dog_id` + `schedule_id`), so the two healed places are complete for every known contributor. But that's an accident of the current three, not a guarantee. `docs/INTEGRATING.md` now states it as a contract: put your device id at the top level of your `source` payload under `device_id`, because a device id anywhere else survives the rewrite stale and you'll duplicate your task on the next reconcile.

**Declined**

4. **Narrowing the `except Exception`.** This is deliberate, and narrowing it would invert the point. The guard exists because an earlier version raised out of `async_setup_entry` and left the integration unloadable — a repair refusing to run must never be why Home Keeper won't start. Catching only `KeyError` and `AttributeError` lets everything else brick setup, which is the exact failure mode being guarded. On "at least log the full traceback": it already does. It's `_LOGGER.exception`, not `_LOGGER.error`, and the message asks the user to report it with the traceback (`__init__.py:460-464`).
5. **Documenting the detach race.** Already documented, in the docstring of the function itself:

   > That check and the update below are not atomic: nothing stops another integration removing its own entry in between, which would make ours the last one after all. Home Assistant's registry offers no remove-if-not-last, and the window is a few synchronous statements during our setup, so this is the best the API allows.

**Answered**

6. **"Verify conftest.py actually runs different builds per phase."** Fair thing to insist on, given the harness's history. It does, and explicitly: each step is an `(ha_tag, hk_source)` pair, and `_use_home_keeper` replaces the mounted `custom_components/home_keeper` between boots while nothing is running (`tests/upgrade/conftest.py:285-302, 406-416`). Phase one is `PREVIOUS_HK` (the last release, fetched at a pinned SHA), phase two is `WORKING_TREE_HK`. The `ha_first` path is a third boot, so the fix arrives on the already-damaged world rather than being present while it's created.

**One correction.** You wrote that the grouping key "doesn't include task name, which is correct" — it never did include the task name, so there's nothing there to be right or wrong about. It's `f"{namespace}|{json.dumps(normalized, sort_keys=True, default=str)}"` over the canonicalized `source` payload only, exactly as you quoted from line 804. Worth flagging because your comment reads as though a change had been made and verified.

---

prestomation:
/q review Two convention changes in `e3a66cc`, plus whatever you want to revisit from the previous push. Be critical and skeptical; most serious first, and don't withhold the minor ones.

- `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md` now say **never comment on a GitHub issue** — analysis and status go in the PR, and `Fixes #N` closing on merge is the only signal an issue needs. Check the two files actually agree with each other and that the rule doesn't contradict anything else in either file (PR comments, including `/q review` and review-thread replies, are explicitly still required — make sure that carve-out is unambiguous).
- `IDEAS.md` records that `companions_catalog.py` lists only `battery_notes`, so the Bambu Lab glue is never *suggested* even though push registration still surfaces it once installed. Is that characterization right about how `companions.py` / `companions_catalog.build_companion_list` behave?

Also still open from the last round if you want another look: the survivor ranking in `store.async_merge_split_duplicates` now uses `enumerate` rather than `list.index`, and `docs/INTEGRATING.md` states the top-level `source.<ns>.device_id` contract the repair depends on.

---

prestomation:
/q review One workflow change in `9685a3e`. Be critical and skeptical, most serious first, and don't withhold minor points.

`walkthrough-preview.yml`'s sticky-comment step moved from `if: always()` to `if: ${{ !cancelled() }}`. The job cancels its own in-progress runs via `concurrency: cancel-in-progress`, so a second push a few minutes later killed the first run mid-capture; that cancelled run still reached the comment step, read a non-success `steps.capture.outcome`, and overwrote a working gif embed with a "capture failed" note for a run nothing was wrong with.

What I want scrutinized:

- **Does `!cancelled()` actually cover the case?** Step-level cancellation vs job-level, and whether a step cancelled by the job timeout (`timeout-minutes: 20`) behaves the same way. If a timeout should still post the failure note, say so — I'd consider that a real capture failure worth surfacing.
- **Does this open a hole where a genuine failure now goes unreported?** The step it guards is the only thing that writes the comment, so I want to be sure a real Playwright failure still reaches it. `continue-on-error: true` on the capture step means the job succeeds, so `!cancelled()` should be true — confirm that reasoning.
- **Staleness.** With the failure note suppressed on cancellation, the comment keeps showing the *previous* push's gif until the superseding run finishes. Is a stale-but-valid embed for a couple of minutes the right trade against a false failure claim, or should it say something in between?
- **Interaction with the `closed` tombstone branch** and with the `Publish / remove preview` step's own `if:`, which I did not change.

---

prestomation:
No code change from this round — the two substantive answers confirm the reasoning, but three claims need correcting for anyone reading this thread later.

**The `closed` tombstone is in the concurrency group.** The group key is `walkthrough-preview-${{ github.event.number }}`, with nothing distinguishing event types, so a `closed` run absolutely can cancel an in-flight `synchronize` run for the same PR (and be cancelled by one). "Closed events aren't part of the concurrency group" is wrong. It happens not to matter: the `closed` run is the one that survives being *started* last, and it does no capture, so it reaches the comment step in a couple of seconds.

**The timeout claim I'm not taking on faith, and it doesn't matter either way.** A 20-minute job timeout terminates the job, so the comment step wouldn't have run under `always()` either — this change can't have regressed that path. If a timed-out capture ever does need a note, the fix is a separate step, not loosening this condition to `success() || failure()`, which walks straight back into the false-failure bug.

**The "successful capture with no posted embed" edge case doesn't leave a hole.** A push during the publish/comment steps cancels run A before it comments, yes — but run B then posts *its own* result to the same sticky comment. The comment always ends up reflecting the last run to finish, which is the correct one. There's no state where a capture succeeds and the comment is permanently missing, so the suggested concurrency rework has nothing to fix. Declining it.

Agreed on the two that count: `continue-on-error: true` keeps the job green on a genuine Playwright failure, so `!cancelled()` is true and the failure note still posts; and stale-but-valid for a couple of minutes beats a false claim that contradicts a live embed. Not adding a "capture in progress" variant — that's a third state to maintain for a two-minute window.

---

PR #192, opened 2026-08-11
Mutation pass over the pure Python core (79.4% → 84.2%)

prestomation:
## What

A full mutation pass over the Python surface — the follow-up to #190, which added the gate but left the existing debt in place. `--all` was failing at **79.42%**, just under the 80% threshold; it now passes at **84.21%**.

**5,136 mutants. 4,079 detected → 4,325.** 246 killed, 748 unit tests passing (up from 717). Tests only; no production code changed.

## What the survivors were telling us

The 1,057 survivors weren't scattered. Four systematic gaps accounted for most of them, each the same shape: a test that *runs* code without *asserting* on what it produced.

### 1. Validation messages were never checked — 60 sites, ~245 mutants

Every `pytest.raises(TaskValidationError)` in the unit tier asserted only that *something* was raised. mutmut could blank a message to `None`, upper-case it, or wrap it in junk, and nothing failed. Those strings are contract: the service layer re-raises them to the user as `ServiceValidationError`, and the panel shows them verbatim.

All 60 now go through a shared helper:

```python
def raises_exactly(exception, message):
    return pytest.raises(exception, match=f"^{re.escape(message)}$")
```

Both the anchoring and the escaping earn their keep. `pytest.raises(match=...)` is a **search**, so an unanchored pattern passes happily on `"XXname must not be emptyXX"` — I hit exactly that in the template's equivalent pass. And messages like `"invalid date for 'Warranty': 'not-a-date' (expected YYYY-MM-DD)"` contain regex metacharacters that would otherwise be interpreted rather than matched.

The 60 expected messages were captured mechanically rather than hand-written, by temporarily wrapping `pytest.raises` in a throwaway plugin to record what each site actually raised — so they document current behaviour rather than my guess at it.

### 2. `merge_update`'s carry-forward was never exercised one field at a time — ~70 mutants

`models.merge_update` and `assets.merge_update` both build a candidate from `updates.get(field, existing.get(field))` across every editable field, then re-normalise. So **every field a caller omits has to survive untouched** — the panel's edit form sends a partial payload, and a service caller may send a single key.

Merging a full payload, which is all the suite did, structurally cannot tell the carry-forward apart from the update. Both are now parameterised over the field list, with a guard test that a newly added editable field must appear in the fixtures rather than silently going untested. Plus the degenerate empty-payload case, and the identity/history fields (`id`, `created`, `completions`, `identifiers`) that survive only because the merge starts from `dict(existing)`.

That work surfaced a real coupling worth naming: switching `completion_detail` away from `required` **clears** `completion_required_fields`. That's correct — a mandatory-field list is meaningless when capture is off — so the test data documents it as an expected knock-on rather than the assertion being weakened to tolerate it.

### 3. The inventory row shape was only spot-checked — ~22 mutants

That dict *is* the export contract: the CSV writer and the panel index it by key, and the inventory export is the one artefact a user hands to a third party (an insurer). Asserting a handful of fields let a renamed or dropped key through. It now asserts the whole row, that absent text coalesces to `""` rather than a literal `"None"` in a CSV cell, and that money rounds to cents — with values chosen so 2dp differs from **both** 0dp and 3dp, which the previous assertions couldn't distinguish (`round(9.999, 2) == round(9.999) == 10`).

## What remains, and why I stopped here

**811 mutants are still undetected**, concentrated in `assets` (296), `models` (168), `notifications` (92) and `recurrence` (81). I stopped deliberately: after these four sweeps the remainder are individual gaps rather than a fifth pattern, so they want a considered follow-up pass rather than another mechanical sweep, and a 4,000-line diff of one-off tests would not be reviewable.

The gate is unaffected either way — it's diff-scoped, so a PR is judged on the functions it touches, not on this backlog.

## Verification

```console
$ bash ci/test-mutation-python.sh --all
**Mutation score: 84.21%** (4325 detected / 5136 scored) — threshold 80% — ✅ pass

$ python -m pytest tests/unit -q          # 748 passed, 1 skipped (was 717)
$ ruff check custom_components tests ci   # clean
$ ruff format --check custom_components tests ci   # 111 files already formatted
```

## Notes

- Developer-facing, tests only: no CHANGELOG entry, no beta bump, no UI touched.
- Stacked on the same `main` as #191 (the gate fix); no conflict between them.

---
_Generated by [Claude Code](https://claude.ai/code/session_011FSVAV9NUEmLS79voYAEja)_

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,6 +23,12 @@ include = ["home_keeper*"]
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 asyncio_mode = "auto"
+# Make `from asserts import raises_exactly` (tests/unit/asserts.py) resolve
+# explicitly. Without it the import works only by accident — pytest's default
+# import mode puts each test file's own directory on sys.path, so it would break
+# the moment a unit test moved into a subdirectory. This adds no *source* path,
+# so it doesn't affect how mutmut derives a mutant's module name.
+pythonpath = ["tests/unit"]
 markers = [
     "real_ha: marks tests that use real HA modules (not mocked)",
 ]
--- a/tests/unit/asserts.py
+++ b/tests/unit/asserts.py
@@ -1,0 +1,29 @@
+"""Shared assertion helpers for the pure unit tier.
+
+Kept out of ``conftest.py`` so it stays a plain importable module: the unit
+tests import it as ``from asserts import raises_exactly`` (pytest puts each test
+file's directory on ``sys.path``).
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Any
+
+import pytest
+
+
+def raises_exactly(exception: type[BaseException], message: str) -> Any:
+    """``pytest.raises`` that pins the **whole** error message.
+
+    Home Keeper's validation messages are contract, not decoration: the service
+    layer re-raises them to the user as ``ServiceValidationError`` and the panel
+    shows them verbatim, so a test that asserts only "something was raised"
+    lets the wrong message — or an empty one — ship unnoticed.
+
+    Anchored and escaped on purpose. ``pytest.raises(match=...)`` is a
+    *search* over a regex, so an unanchored pattern passes on a message with
+    anything bolted onto either end, and a message containing ``(`` or ``.``
+    would otherwise be read as a pattern rather than as text.
+    """
+    return pytest.raises(exception, match=f"^{re.escape(message)}$")
--- a/tests/unit/test_assets.py
+++ b/tests/unit/test_assets.py
@@ -8,6 +8,7 @@ from datetime import datetime, timedelta, timezone
 
 import hk_assets as a
 import pytest
+from asserts import raises_exactly
 
 TZ = timezone(timedelta(hours=-4))
 NOW = datetime(2026, 6, 13, 10, tzinfo=TZ)
@@ -100,12 +101,12 @@ def test_asset_device_identifier_is_prefixed():
 
 
 def test_build_virtual_asset_requires_name():
-    with pytest.raises(a.AssetValidationError):
+    with raises_exactly(a.AssetValidationError, "missing required field: 'name'"):
         a.build_asset({"manufacturer": "X"}, now=NOW)
 
 
 def test_build_existing_asset_requires_device_id():
-    with pytest.raises(a.AssetValidationError):
+    with raises_exactly(a.AssetValidationError, "missing required field: 'device_id'"):
         a.build_asset({"kind": "existing"}, now=NOW)
 
 
@@ -128,7 +129,7 @@ def test_build_existing_asset_keeps_device_id_no_identifier():
 
 
 def test_build_asset_rejects_bad_kind():
-    with pytest.raises(a.AssetValidationError):
+    with raises_exactly(a.AssetValidationError, "invalid kind: 'imaginary'"):
         a.build_asset({"name": "x", "kind": "imaginary"}, now=NOW)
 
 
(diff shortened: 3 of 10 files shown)


prestomation:
/q review Please be critical and skeptical, most-serious-first, and don't withhold minor findings. Tests only, so the question is whether these are *good* tests or mutant-shaped ones.

**`raises_exactly` is the change with the widest blast radius** — 60 call sites now pin the full text of a validation message. Is that the right trade for messages that reach users through `ServiceValidationError`, or does it make every copy edit a test failure? If the latter, what should the model expose instead — the messages are currently the only machine-readable thing `TaskValidationError` carries, and moving to codes means touching `strings.json` exception keys too. Also: is `tests/unit/asserts.py` importable the way I've assumed (bare `from asserts import …`, relying on pytest putting the test dir on `sys.path`), or will that break under a different import mode or if a name collides?

**The `merge_update` carry-forward tests.** `FULLY_SET_TASK` / `FULLY_SET_ASSET` plus a parameterised field list, with a guard test asserting the fixtures cover `CARRIED_FIELDS`. Does the guard actually fail when someone adds an editable field, or can a field be added to `merge_update`'s candidate dict without tripping it? Is asserting `updated == task` for an empty payload too strong — is there any field that legitimately moves on a no-op merge (a timestamp, a recompute)? And is the `completion_detail` → `completion_required_fields` coupling I encoded in the test data actually correct behaviour, or is it a bug I've just cemented?

**The inventory row test asserts the entire row dict.** Over-fitted, or right for an export contract? It will fail on any added column — is that the failure you want?

**Rounding.** `test_spares_value_rounds_to_cents` picks values where 2dp differs from 0dp and 3dp. Does it accidentally depend on float representation or Python's round-half-to-even in a way that could differ across versions?

**What I did *not* do.** 811 mutants remain, mostly in `assets`, `models`, `notifications`, `recurrence`. Is stopping here defensible, or is there a cluster in those four I've misjudged as "individual gaps" that's actually another systematic pattern worth sweeping now?

**Anything now at a high score but still genuinely untested** — where the number is flattering the suite rather than reflecting it.

---

prestomation:
Thanks — one actionable finding, taken in 2fd0668.

**The `asserts` import was brittle, and you're right about why.** `from asserts import raises_exactly` worked only because pytest's default import mode puts each test file's *own* directory on `sys.path`, which today happens to be `tests/unit`. Move a unit test into a subdirectory and that subdirectory goes on the path instead, and the import breaks with a message that points nowhere useful.

Fixed by stating the dependency rather than relying on the accident:

```toml
[tool.pytest.ini_options]
pythonpath = ["tests/unit"]
```

I didn't use `from .asserts import` — `tests/unit/` has no `__init__.py` and adding one changes how pytest roots the whole tier — nor `from tests.unit.asserts import`, which needs the same. `pythonpath` is the narrow fix.

Worth confirming explicitly, since AGENTS.md warns that a `pythonpath` in pytest config is a classic cause of mutmut mis-keying every mutant: this adds no **source** path, only a test-helper one, so mutmut still derives mutant keys from the real module paths. Verified rather than assumed — re-ran a filtered mutation after the change:

```console
$ mutmut run "custom_components.home_keeper.inventory.x_build_inventory*"
🎉 custom_components.home_keeper.inventory.x_build_inventory__mutmut_99
```

Keys unchanged, mutants still killed. 748 unit tests pass, `ruff check custom_components tests ci` clean.

**On error codes** — agreed it's the better long-term shape and agreed it's not this PR. It would mean adding a code to `TaskValidationError`, threading it through the service layer, and adding matching `strings.json` exception keys for localisation parity. Happy to open that separately if you want it.

**On the guard test's limits** — a fair catch, and I'd rather be precise about what it does and doesn't do. `test_merge_update_cases_cover_the_carried_fields` asserts the *fixtures* are self-consistent: it fails if `MERGE_UPDATE_CASES` names a field outside `CARRIED_FIELDS`, or if `FULLY_SET_TASK` does. It cannot see a field added to `merge_update`'s candidate dict and nowhere else — nothing in the test tier can, short of reflecting over the production source, which I'd rather not do. So it's a fixture-drift guard, not a completeness guard, and the comment now says so rather than implying more.

Everything else — the empty-payload strictness, the `completion_detail` coupling, the whole-row inventory assertion, and the rounding values being exactly representable — thanks for confirming; no changes.

---

PR #193, opened 2026-08-11
Mutation pass over the frontend (58.3% → 74.9%)

prestomation:
## What

A mutation pass over the TypeScript surface — the frontend counterpart to #192, and the follow-up to #190 which added the gate but left the existing debt in place.

**1,821 mutants. 1,062 detected → 1,363. 58.32% → 74.85%**, over two rounds. Of the 759 originally undetected, **285 had no coverage at all**: whole exported functions no test ever called. That's down to 86.

| File | Before | After |
| --- | ---: | ---: |
| `forms.ts` | 32.94% | **66.14%** |
| `documents.ts` | 59.83% | **81.20%** |
| `utils.ts` | 79.73% | **84.38%** |
| `card-filter.ts` | 75.69% | 77.19% |
| `markdown.ts` | 80.15% | 80.15% |
| `limits.ts` | 100% | 100% |
| **All files** | **58.32%** | **74.85%** |

**Frontend tests: 333 → 445.** Tests only; no production code changed.

**This is still below the 80% gate**, and I'd rather say so plainly than dress it up — see "What remains" below. The gate is diff-scoped, so it blocks nothing; `--all` is a manual `workflow_dispatch` run.

## What the survivors were telling us

**`forms.ts` is the panel's entire edit surface** (32.94%), and only its task payload was meaningfully tested — `completion.test.js` and `sensor-form.test.js` import three of its ~30 exports.

- **The selector factories**, asserted as whole objects. A selector *is* a contract with `ha-form`: the wrong shape silently renders the wrong control, or none. That includes details that look incidental — `sort: false` on dropdowns matters because the options are ordered deliberately (statuses run worst-first) and letting `ha-form` sort them alphabetically reorders them.
- **`taskSchema` per recurrence type**, asserted as whole field lists: floating gets "every N units", fixed swaps the unit for a calendar frequency plus an anchor, one-off gets a single due date and no cadence or "last done" seed, and triggered gets *only* its descriptive fields — offering it a schedule would let a user arm a dormant task whose state the monitoring integration owns. Locked fields are asserted **omitted**, not shown disabled.
- **The datetime bridge** between HA's local `"YYYY-MM-DD HH:mm:ss"` and the ISO we persist, including the single-digit case that is the only reason `padStart` is there — without it a January date renders `2026-1-2 3:4:5`.
- **`backstopEnabled`**, which keeps the schema, payload and hint from disagreeing about whether a usage task's time backstop applies — the split that shows a switched-off backstop still in force.
- **The notification round-trip's coercions**: a null profile renders as `""` for the select and comes back as `null` rather than the string; a blank name falls back instead of saving empty; snooze hours reject `0` and `NaN`, both of which would otherwise mean "snooze forever" or "snooze never".

**`card-filter.ts`: `profileMatches` decides which tasks a notification profile sends**, and had no direct test. Its windows have to agree with the backend or a digest reports a different set than the panel previewed. [Verified against the Python side](https://github.com/prestomation/ha-home-keeper/pull/193#issuecomment-5251401469): `overdue or (now < due <= now + window)` collapses to `due <= now + 3 days`, matching the frontend's inclusive edge exactly.

**`documents.ts`: the link-vs-file branch lives in exactly one place** so the panel and card can't drift (uploaded files were first missed on the card). Each kind keys off a *different* field, so a link carrying only a `filename` renders an un-openable affordance.

**`utils.ts`: the URL guards.** `isHttpUrl` / `safeHref` / `isSafeImageUrl` stand between a caller-supplied string and an `href`/`src` — `home_keeper.complete_task` takes a photo URL from outside the panel — so they're asserted against `javascript:`, `data:`, `vbscript:` and protocol-relative `//host` payloads, and on **anchoring**: a scheme appearing anywhere in the string must not pass.

`randomId` is tested on its fallback paths, where the interesting bug lives: `crypto.randomUUID` exists only in a secure context, so over a plain-HTTP LAN address it is `undefined` and calling it throws, silently breaking uploads. The hand-built v4 has to set the RFC 4122 version and variant bits itself, so those are asserted directly rather than via a lax hex regex.

## What remains

**458 mutants are still undetected**, 231 of them in `forms.ts` — and unlike the no-coverage bucket these are *covered but unasserted*: `taskSchema`, `buildTaskPayload` and `sensorHintText` run under existing tests that don't check enough of what they produce. Reaching 80% overall needs roughly 94 more kills, almost all there.

I stopped after two rounds because the remainder are individual assertions rather than a third pattern, and each verification run is ~50 minutes over 1,821 mutants, so the iterate-and-measure loop is slow. A third round is worth doing; it isn't worth bundling into an already-large diff.

`markdown.ts` (80.15%) I left alone: it's already at the gate, and its logic is the lazy-registration dance for `<ha-markdown>` — DOM-heavy, better served by the e2e tier. Its user text is `escapeHTML`'d on both the rendered and plain-text branches before reaching `innerHTML`, so the deferral to HA's DOMPurify-in-a-Worker sanitizer is a second layer, not the only one.

## Verification

```console
$ bash ci/test-mutation-frontend.sh --all
All files  |  74.85 |   78.56 |     1360 |         3 |        372 |       86 |        0 |

$ npx vitest run
Test Files  18 passed (18)
     Tests  445 passed (445)
```

Datetime tests confirmed timezone-independent by running them under `Asia/Kathmandu` (+05:45), `Pacific/Chatham`, `Australia/Lord_Howe` (30-minute DST shift), `Pacific/Apia` and `America/New_York` — all green.

## Notes

- Developer-facing, tests only: no CHANGELOG entry, no beta bump, no UI touched.
- Independent of #192; both branch from the same `main`.
- One thing I noticed and deliberately did **not** fix here: the 3-day due-soon window is declared twice — `transitions.DUE_SOON_WINDOW` in Python and `DUE_SOON_DAYS` in `card-filter.ts` — with nothing enforcing they stay equal. Real drift risk, but not a tests-only change.

### DIFF
--- a/custom_components/home_keeper/frontend/test/card-filter-profile.test.js
+++ b/custom_components/home_keeper/frontend/test/card-filter-profile.test.js
@@ -1,0 +1,136 @@
+import { describe, expect, it } from 'vitest';
+import { DUE_SOON_DAYS, profileMatches } from '../src/card-filter.ts';
+
+// `profileMatches` decides which tasks a notification profile sends. It has to
+// agree with the backend's windows exactly, or a digest reports a different set
+// than the panel previews. `card-filter.test.js` covers the card's own filters;
+// this covers the profile predicate, which nothing exercised directly.
+
+const DAY_MS = 86_400_000;
+const NOW = Date.parse('2026-06-13T10:00:00Z');
+const at = (offsetDays) => new Date(NOW + offsetDays * DAY_MS).toISOString();
+
+const task = (over = {}) => ({
+  id: 't1',
+  name: 'Filter',
+  enabled: true,
+  next_due: at(-1),
+  ...over,
+});
+
+describe('profileMatches gating', () => {
+  it('excludes disabled tasks', () => {
+    expect(profileMatches(task(), {}, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ enabled: false }), {}, {}, {}, NOW)).toBe(false);
+  });
+
+  it('treats an absent `enabled` as enabled', () => {
+    // Only an explicit `false` disables; `undefined` is an older record.
+    const t = task();
+    delete t.enabled;
+    expect(profileMatches(t, {}, {}, {}, NOW)).toBe(true);
+  });
+
+  it('excludes undated tasks', () => {
+    expect(profileMatches(task({ next_due: null }), {}, {}, {}, NOW)).toBe(false);
+    expect(profileMatches(task({ next_due: '' }), {}, {}, {}, NOW)).toBe(false);
+  });
+
+  it('excludes problem-sensor tasks', () => {
+    // These are driven by a binary sensor, not a schedule; notifying on them
+    // would duplicate the sensor's own alert.
+    expect(profileMatches(task({ source: { problem_sensor: 'binary_sensor.x' } }), {}, {}, {}, NOW)).toBe(
+      false,
+    );
+    // A source of another shape must not be swept up with them.
+    expect(profileMatches(task({ source: { part: { asset_id: 'a1' } } }), {}, {}, {}, NOW)).toBe(
+      true,
+    );
+    expect(profileMatches(task({ source: null }), {}, {}, {}, NOW)).toBe(true);
+  });
+});
+
+describe('profileMatches status windows', () => {
+  it('defaults to overdue when no status is set', () => {
+    // The default has to be `overdue`, not "everything": an empty profile that
+    // matched every dated task would notify on the whole list.
+    expect(profileMatches(task({ next_due: at(-1) }), {}, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ next_due: at(1) }), {}, {}, {}, NOW)).toBe(false);
+    expect(profileMatches(task({ next_due: at(1) }), { status: '' }, {}, {}, NOW)).toBe(false);
+  });
+
+  it('overdue means due at or before now, inclusive', () => {
+    const f = { status: 'overdue' };
+    expect(profileMatches(task({ next_due: at(-1) }), f, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ next_due: new Date(NOW).toISOString() }), f, {}, {}, NOW)).toBe(
+      true,
+    );
+    expect(profileMatches(task({ next_due: at(0.001) }), f, {}, {}, NOW)).toBe(false);
+  });
+
+  it('due_soon spans overdue through the window edge, inclusive', () => {
+    const f = { status: 'due_soon' };
+    expect(profileMatches(task({ next_due: at(-5) }), f, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ next_due: at(DUE_SOON_DAYS) }), f, {}, {}, NOW)).toBe(true);
+    // One millisecond past the window is out — this is what pins the boundary
+    // as `>` rather than `>=`.
+    const justPast = new Date(NOW + DUE_SOON_DAYS * DAY_MS + 1).toISOString();
+    expect(profileMatches(task({ next_due: justPast }), f, {}, {}, NOW)).toBe(false);
+  });
+
+  it('all accepts any dated, enabled task', () => {
+    const f = { status: 'all' };
+    expect(profileMatches(task({ next_due: at(-30) }), f, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ next_due: at(365) }), f, {}, {}, NOW)).toBe(true);
+    expect(profileMatches(task({ next_due: null }), f, {}, {}, NOW)).toBe(false);
+  });
+});
+
+describe('profileMatches area and device filters', () => {
+  const devices = { d1: { area_id: 'kitchen' }, d2: { area_id: 'garage' } };
+
+  it('an empty list means "no constraint", not "match nothing"', () => {
+    const t = task({ area_id: 'kitchen', device_id: 'd1' });
+    expect(profileMatches(t, { areas: [], devices: [], labels: [] }, devices, {}, NOW)).toBe(true);
+    // Absent behaves the same as empty.
+    expect(profileMatches(t, {}, devices, {}, NOW)).toBe(true);
+  });
+
+  it('filters by the task area', () => {
+    const t = task({ area_id: 'kitchen' });
+    expect(profileMatches(t, { areas: ['kitchen'] }, devices, {}, NOW)).toBe(true);
+    expect(profileMatches(t, { areas: ['garage'] }, devices, {}, NOW)).toBe(false);
+    expect(profileMatches(t, { areas: ['garage', 'kitchen'] }, devices, {}, NOW)).toBe(true);
+  });
+
+  it('falls back to the device area when the task has none', () => {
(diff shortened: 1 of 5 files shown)


prestomation:
/q review Please be critical and skeptical, most-serious-first, and don't withhold minor findings. Tests only, so the question is whether these are *good* tests or mutant-shaped ones.

**Asserting whole selector objects.** `selText()` etc. are asserted with `toEqual({text: {}})` rather than checking a key is present. I argued a selector is a contract with `ha-form` and the exact shape matters (`{device: {}}` vs `{device: {multiple: false}}` render differently). Is that right, or have I just frozen an implementation detail that will break on every `ha-form` version bump? Same question for `sort: false` on dropdowns.

**Timezone dependence.** The `isoToHaDateTime` tests build their expected values from `new Date(2026, 5, 13, 9, 8, 7)` — local parts — so they should hold in any TZ. Do they? Is there a DST boundary or a non-integer-offset zone (Kathmandu, Chatham) where the round-trip assertion `haDateTimeToIso(isoToHaDateTime(x)) === x` fails? CI runs UTC but contributors don't.

**`profileMatches` boundary tests.** I assert `due_soon` is inclusive at exactly `now + DUE_SOON_DAYS * DAY_MS` and exclusive one millisecond past. Is that actually the backend's rule, or have I pinned the frontend to something the Python side disagrees with? A digest that reports a different set than the panel previewed is the bug this is supposed to prevent, and a test agreeing with the *wrong* side makes it permanent.

**Global stubbing in `randomId` tests.** `vi.stubGlobal('crypto', undefined)` with `vi.unstubAllGlobals()` in an `afterEach` scoped to that `describe`. Does that leak if a test throws, or across files in the same worker? And is `expect(id[14]).toBe('4')` / `'89ab'.toContain(id[19])` genuinely testing the RFC 4122 bit-setting, or would it pass on a wrong implementation?

**Over-fitting.** `documentLabel({kind:'link'})` returning `''`, and the assertions that a link does *not* fall back to `filename`. Real contract, or am I pinning accidental behaviour?

**What I left.** `markdown.ts` at 80.15% untouched, and `panel.ts`/`card.ts`/`api.ts` are off the mutable surface entirely. Is `markdown.ts` a mistake to skip given it renders user-supplied text into the DOM?

---

prestomation:
Thanks — that's a useful review. Two of the questions I asked deserved evidence rather than reasoning, so I went and got it. No code changes needed.

## The `due_soon` boundary does match the backend — verified

You were right to push on this rather than take my comment's word for it; a test agreeing with the *wrong* side would make the bug permanent. The backend:

```python
# recurrence.py
def is_due_soon(task, window, *, now):
    return now < next_due <= now + window        # inclusive upper edge

# profiles.py — STATUS_DUE_SOON
if status == STATUS_DUE_SOON and not (overdue or recurrence.is_due_soon(task, window, now=now)):
```

`overdue OR (now < due <= now + window)` collapses to `due <= now + window`. The frontend is `if (status === 'due_soon' && due > now + DUE_SOON_DAYS * DAY_MS) return false` — passes on `due <= now + 3 days`. Same set, same inclusive edge.

The windows are also the same 3 days from independent constants — `transitions.DUE_SOON_WINDOW = timedelta(days=3)` and `card-filter.ts`'s `DUE_SOON_DAYS = 3`. Worth noting those are *two* declarations that could drift; nothing enforces they stay equal. Not something to fix in a tests-only PR, but a real thing, and I've left it out of scope rather than quietly patching it here.

## Timezone independence — measured, not assumed

Your reasoning was right, but "both conversions use the same `Date` APIs so they're self-consistent" is the kind of claim that's easy to believe and wrong at a DST boundary, so I ran the datetime tests under the awkward zones:

```console
UTC                    Tests  36 passed (36)
Asia/Kathmandu         Tests  36 passed (36)   # +05:45
Pacific/Chatham        Tests  36 passed (36)   # +12:45 / +13:45 DST
Australia/Lord_Howe    Tests  36 passed (36)   # 30-minute DST shift
Pacific/Apia           Tests  36 passed (36)
America/New_York       Tests  36 passed (36)
```

## On `markdown.ts` — your description checks out

I verified before repeating it, since it's a security claim. `renderMarkdown` emits `<ha-markdown data-md="${escapeHTML(value)}">` and the plain-text fallback is `escapeHTML(value)` too, so user text is escaped on **both** branches before it reaches `innerHTML` — the deferral to HA's DOMPurify-in-a-Worker sanitizer is a second layer, not the only one. The interesting logic there is the lazy-registration dance for `<ha-markdown>` (absent from HA's eager bundle), which is DOM-heavy and better served by the e2e tier than by mutation testing. Leaving it at 80.15%.

## On the selector objects

Taking your agreement, with one caveat I want on the record: if a future `ha-form` changes its defaults, these tests fail as a *notification*, not as a correctness signal — the right response would be to update them, not to treat it as a regression. That's the trade I made deliberately, since the alternative (asserting only that a key is present) can't see `sort: false` going missing, which silently reorders every dropdown.

The `--all` score for this branch is still measuring; I'll post it here when it lands.

---

prestomation:
Final score, as promised. **58.32% → 74.85%** (1,363 of 1,821 detected, +301 kills). PR body updated with the real numbers.

```
All files       |  74.85 |   78.56 |     1360 |         3 |        372 |       86 |        0 |
 card-filter.ts |  77.19 |   83.24 |      308 |         0 |         62 |       29 |        0 |
 documents.ts   |  81.20 |   93.14 |       95 |         0 |          7 |       15 |        0 |
 forms.ts       |  66.14 |   68.57 |      504 |         0 |        231 |       27 |        0 |
 limits.ts      | 100.00 |  100.00 |        2 |         0 |          0 |        0 |        0 |
 markdown.ts    |  80.15 |   81.40 |      103 |         2 |         24 |        2 |        0 |
 utils.ts       |  84.38 |   87.25 |      307 |         1 |         45 |       12 |        0 |
```

**It's below the 80% gate and I'm not going to pretend otherwise.** `--all` exits 1. It blocks nothing — the PR check is diff-scoped and `--all` only runs on manual `workflow_dispatch` — but the honest summary is "substantially better, not yet at the bar."

Two rounds got it here. The first covered the *uncovered* — whole exported functions no test ever called, which was 285 mutants across `forms.ts`, `documents.ts`, `utils.ts` and `card-filter.ts`. That landed 67.44%, which I want to note because I briefly read an early progress line (54 survivors at ~600 mutants) as a strong signal and it wasn't — the survivor rate climbed steeply once the run reached `forms.ts`. The second round went after `taskSchema` and the notification round-trip specifically, which took `forms.ts` from 48.43% to 66.14% and the whole surface to 74.85%.

**What's left is one shape, not four.** 231 of the 458 remaining are in `forms.ts`, and they're *covered but unasserted* — `taskSchema`, `buildTaskPayload` and `sensorHintText` all execute under existing tests that don't check enough of what comes back. Roughly 94 more kills clears 80%, almost all there. I stopped because those are individual assertions rather than a third systematic pattern, and each verification is ~50 minutes over 1,821 mutants, so bundling another round into this diff means a much larger change reviewed against a much slower loop.

Worth flagging one thing writing these caught, since it's the kind of error the tests exist to prevent: my first `taskSchema` assertions failed because I assumed the cadence fields were top-level. They're nested inside an unnamed `grid` container, so `map(f => f.name)` silently missed `interval`, `unit` and `freq` — exactly the fields under test. Four tests went red before I dumped the real structure. The helper recurses now.

---

PR #197, opened 2026-08-12
fix: heal a task off the asset that shares its dead device (#194 follow-up)

prestomation:
Follow-up to a review of #194. No panel UI changes, so the screenshots and walkthrough gates don't apply.

## The bug

#194 added a snapshot fallback so a composite device Home Assistant has already garbage-collected can still be resolved — but wired it into the asset path only. Tasks keep no identifiers/connections snapshot, so `_split_successor` still returned `None` for them, and `asset_mapping` was never fed to `async_repoint_device_ids`.

Result: a task and an asset on the same dead composite came apart. The real snapshot committed with that PR shows the case exactly — asset "T6 Pro Thermostat" and task "Replace Air Filter (Appliance)" both sit on `620eef30…`:

| | before #194 | after #194 | after this PR |
|---|---|---|---|
| Asset | dead composite | healed | healed |
| Task on the same device | dead composite | **still dead** | healed |

One appliance split across two device pages is worse than the consistent breakage it replaced.

## Fix

Tasks and assets now resolve into **one shared mapping**, applied to both stores. A device only an asset can identify heals the task sitting on it. The `mapping` / `asset_mapping` two-step and its reuse special-case both disappear.

Also from the review:

- **Merge `_resolve_successor_from_snapshot` into `_resolve_by_snapshot`** — near-identical functions 300 lines apart in the same file. The foreign preference becomes an optional `prefer_not_entry` argument. Its `len(foreign) > 1` branch goes: `sorted(xs)[0]` on a one-element list *is* that element.
- **`store.async_repoint_asset_device_ids`**, mirroring the task repoint — one write for the batch instead of `set_asset_device_id`'s one full store serialization per asset during setup, and one pass over `list_assets()` instead of two.
- **`ASSET_KIND_EXISTING`** instead of a bare `"existing"` literal (the constant exists; the sibling `ASSET_KIND_VIRTUAL` was already imported here).
- **`set_asset_device_id` docstring** corrected — it serves existing-kind assets too, not only provisioned virtual ones.

## Testing

**`tests/unit/test_snapshot_heal.py` did not test the shipped code.** It replicated `_resolve_successor_from_snapshot` verbatim into the test file and tested the copy, so the production function had zero coverage and any later edit to it would stay green. Its stated reason ("can't import devices.py on this host") doesn't hold for this repo — four files in `tests/unit/` already import `homeassistant`. Two of its six tests also could not fail: the mock registry returned a single candidate, making the foreign preference unobservable, and `..._picks_sorted` never asserted *which* device it picked.

Replaced with **`tests/unit/test_device_heal.py`** (17 tests), which loads the real `devices.py` under `hk.*` — same stub-and-load pattern as `test_calendar.py` / `test_coordinator_purge.py` — and injects the registry by patching the loaded module's `dr` binding. Every assertion was verified against a mutated source first:

| mutation | tests failed |
|---|---|
| task repoint dropped (this PR's bug) | 3 |
| foreign preference removed | 3 |
| sorted tie-break → first match | 1 |
| no-op guard removed | 1 |
| snapshot fallback removed (#194's bug 2) | 3 |
| existing-kind filter dropped | 1 |

**Upgrade-tier assertions added.** AGENTS.md: *"Anything resting on an HA framework contract — device registry … — needs an integration-level assertion."* The task-only version of that assertion is what let the asset half ship broken. `tests/upgrade/test_upgrade_repair.py` now asserts every asset points at a live device, that the probe asset lands back on the real water heater, and that a task and asset on one device stay together — with a task seeded onto the asset's device so the shared-mapping path runs against a real registry. `snapshot()` fetches assets alongside tasks.

**Removed `tests/upgrade/fixtures/hk_store_snapshot_2026-08-12.json`** (1995 lines). It's a real `.storage/home_keeper` dump referenced by no test, carrying MAC addresses, Z-Wave node ids, appliance makes/models, purchase and warranty dates, and a Google Drive sharing URL with an intact `resourcekey` capability token. Deleting it doesn't remove it from history — **the Drive link is worth rotating.**

Conventions recorded in `.amazonq/rules/testing-and-workflow.md`: test the shipped function not a copy; check a new test can fail; never commit a real `.storage` dump as a fixture.

## Checks run

- `pytest tests/unit` — 765 passed, 1 skipped
- `ruff check` + `ruff format --check` — clean
- mypy: Home Assistant won't install in this environment (a transitive dep, PyRIC, fails to build), so the integration-wide run is left to `lint.yml`. `mypy --follow-imports=skip` on the two changed modules is clean.
- `tests/upgrade` needs Docker plus `ci/fetch-glues.sh` staging and did not run here — the new assertions are unexecuted locally and CI is the first real run.

Bug-fix only, so no beta bump; `CHANGELOG.md` gets an `## [Unreleased]` entry (0.13.0b2 is already released).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Quz1R9sgc4KR2EDZXsWTTa)_


prestomation:
/q review Be critical and skeptical — this PR fixes bugs a previous review missed, so assume it has its own. Surface the most serious issues first, but don't withhold minor ones.

**Correctness of the shared mapping** in `devices.async_heal_split_device_ids`. Tasks and assets now resolve into one `mapping` applied to both `async_repoint_device_ids` and the new `async_repoint_asset_device_ids`. Can an entry discovered from an asset's snapshot ever be *wrong* for a task on the same device id? Consider: a stale snapshot resolving to a different device than the one the task meant; a virtual asset's device id colliding with a mapping key; two existing-kind assets on the same dead composite with divergent snapshots (last writer wins in the dict — is the skip-if-already-mapped ordering defensible, or should it prefer the richer snapshot?); and whether the `successor.id != device_id` guard is enough to keep the pass idempotent across restarts.

**The snapshot fallback firing on healthy ids.** `async_get_devices_for_composite_device_id` returns `[]` for an ordinary, never-split id as well as for a collected composite, so `_resolve_by_snapshot` now runs on every existing-kind asset at every setup. Is a silent repoint the right behaviour when the snapshot resolves somewhere other than the asset's current live device? `_reconcile_existing` only does snapshot recovery when the device is *missing*.

**`_resolve_by_snapshot`'s two modes.** It serves both reconciliation (`prefer_not_entry=None`, first match) and successor resolution (foreign preference + sorted tie-break). Did merging the two functions change reconciliation's behaviour in any case that matters — the original returned on the first identifier hit and skipped the connections lookup entirely?

**HA best practices / registry contract.** `async_get_device(identifiers={...})` post-2026.8: can two devices legitimately answer for the same identifier tuple, and does the foreign preference actually reach a multi-candidate case in a real registry, or is it dead code outside the multi-identifier scenario the tests construct?

**Performance.** `async_repoint_asset_device_ids` does one save for the batch; is one pass over `list_assets()` plus one over `get_tasks()` at every setup acceptable, and does the shared mapping re-attempt resolution for ids the task loop already failed on?

**Test quality.** `tests/unit/test_device_heal.py` loads the real `devices.py` with stubbed HA symbols — is that load pattern sound, and can `FakeRegistry`/`FakeStore` diverge from the real contracts in a way that would make a test pass while production breaks? I'd rather hear it now than find it in another review.

---

prestomation:
Triaged. Three fixed in `f158e84`, four pushed back on.

### Fixed

**Snapshot fallback on healthy ids — agreed, this was a real defect.** An empty `async_get_devices_for_composite_device_id` answer conflates "collected composite" with "ordinary id, never split", so the fallback ran against every existing-kind asset at every setup. A drifted snapshot would then silently repoint a live asset — and with the shared mapping, the tasks on it too, which is worse than the bug this PR set out to fix.

Implemented the recommended guard: fall back only when `registry.async_get(device_id)` is `None`. That separates the two cases exactly, because a composite is synthesized from the devices still pointing at it — once they're all re-homed there's nothing to synthesize and it resolves to nothing, while a live device answers for itself. It also restores the contract `_reconcile_existing` already had (snapshot recovery only when the device is missing).

Worth flagging that this rests on `async_get` returning `None` for a collected composite. That's what makes the composite/ordinary distinction possible at all, and it's consistent with how the duplicate-merge loop below reads `composite_device_id` off devices in `registry.devices`, but it's an upstream behaviour I inferred rather than executed — `tests/upgrade` is where it gets confirmed against a real registry.

**First-hit short circuit restored.** Correct that merging the two resolvers added a full identifier sweep plus a connections lookup to the reconcile path. Now returns at the first hit when there's no preference, so the behaviour is byte-for-byte what `_reconcile_existing` had before.

**Unresolved-id cache added.** 100 tasks on one dead device now cost one lookup. The cache is deliberately scoped to the snapshot-less pass — assets must still retry those ids, since a snapshot is exactly the extra information that resolves them, and skipping the retry would undo this PR's fix. `test_heal_retries_an_unresolved_task_device_once_an_asset_supplies_a_snapshot` pins that; mutating the guard to `if device_id in unresolved` (dropping the `snapshot is None`) turns three tests red.

### Not changing

**Virtual asset device_id colliding with a dead composite key.** Not reachable, and not a UUID-collision question. Mapping keys come only from task device_ids and existing-kind asset device_ids. For a virtual asset's device to become a key, a task would have to point at it *and* resolve to a successor — but `_split_successor` on a device we own outright returns `[]` from the composite lookup, and the task path passes no snapshot, so it returns `None`. Guarding an unreachable path would add a branch no test could justify.

**Divergent snapshots on one dead composite.** It's first-writer-wins, not last — `_record` returns early on `device_id in mapping`. Two existing-kind assets sharing a `device_id` decorate the same device, so `_reconcile_existing` refreshes both snapshots from that one device and they converge; divergence means the store is already corrupt. Order is `list_assets()` order, which is dict insertion order and stable across restarts, so the choice is deterministic. Preferring a "richer" snapshot would add a ranking heuristic to serve a state that shouldn't exist.

**`successor.id != device_id` insufficient for idempotence.** The premise is that resolution can be non-deterministic, but the sorted tie-break exists precisely to remove that — same registry, same answer, every run. And after the first heal the id *is* the successor, so the second run hits `successor.id == device_id` and writes nothing. `test_heal_is_a_no_op_once_everything_is_healed` asserts zero writes on a settled store, and dropping the guard turns it red.

**`FakeRegistry` has no side-effect tracking.** `async_heal_split_device_ids` performs no registry mutations — it reads the registry and writes the store. There's no side effect to track; device removal happens in `async_prune_orphaned_devices`, which is a different function and out of this diff. Agreed on the broader point that the upgrade tier is the real validation, which is why this PR adds asset assertions there.

Also noting the coverage report: `devices.py` went 0% → 36%. That's the diff between the tests this replaces (which re-implemented the function and tested the copy) and tests that run the shipped code.

---

prestomation:
/q review Re-review `f158e84`, which acts on your last round. Be critical and skeptical — this commit adds a *guard* to a repair path, and a guard that's too tight silently disables the fix it protects, which is the worst failure mode here.

**Correctness of the new gate.** `_split_successor` now falls back to the snapshot only when `registry.async_get(device_id) is None`. That rests on a collected composite resolving to nothing while a live device answers for itself. Is that right for HA 2026.8+? Specifically: can `async_get` still synthesize a composite after every half has been re-homed (which would make this gate silently disable the GC'd-composite repair — #194's headline fix), and is there any state where a *dead* id still resolves to something? If the assumption is wrong I'd rather revert the gate than ship a fix that quietly stops firing.

**The unresolved-id cache.** It's scoped to the snapshot-less pass so assets still retry those ids. Check that scoping is airtight: is there any ordering of tasks and assets where an id lands in `unresolved` and never gets its snapshot-backed retry?

**The restored first-hit short circuit** in `_resolve_by_snapshot` — does returning early for `prefer_not_entry is None` differ in any case from the pre-merge `_reconcile_existing` behaviour, including snapshots with no identifiers but populated connections?

**Interaction with `_reconcile_existing`.** Both now consult the same resolver, and the heal runs before reconcile at setup. Is there a sequence across two restarts where they disagree and oscillate an asset between devices?

Most serious first, and don't withhold minor ones.

---

prestomation:
One correction on the re-review, and a gap it didn't catch.

**The `async_get` gate is still unverified.** The re-review answers question 1 by citing "the implementation and tests (particularly `test_heal_will_not_repoint_a_device_that_is_still_alive`)". That's circular — `FakeRegistry.async_get` is `self.devices.get(device_id)`, a model I wrote to encode the assumption under test. It can't be evidence for it. So the position is unchanged from the PR body: a collected composite resolving to `None` is inferred, not executed, and `tests/upgrade` against a real registry is the check that counts. Flagging it rather than letting "no issues found" stand in for confirmation.

**The coverage report caught something the review didn't.** 54 of 55 new statements covered, with the multi-split branch of `_split_successor` as the miss — no test had more than one foreign split, so neither the `primary_config_entry` preference nor the sorted fallback behind it ran. Covered in `7286131`; removing either turns a test red. 23 unit tests now, 771 in the suite.

On the cache-persistence note: `unresolved` is a local inside `async_heal_split_device_ids`, so it lives exactly one pass and can't outlive it. Nothing to change unless the iteration is ever parallelized, which isn't planned.

Still worth a human eye, in rough order:

1. The `async_get` assumption above — the one thing here that rests on upstream behaviour I couldn't run.
2. `store.async_repoint_asset_device_ids` has no unit coverage (store.py sits at 0% overall — it imports HA). `FakeStore` mirrors its logic rather than running it, which is the same shape of gap this PR set out to fix; the difference is that `tests/upgrade` does execute the real method. Calling it out rather than leaving it implied.
3. The `tests/upgrade` additions have never run locally — Docker plus `ci/fetch-glues.sh` staging — so CI is their first execution.

---

PR #200, opened 2026-08-12
fix(ci): restore real signal to the HA beta nightly and the mypy gate

### LINKED ISSUE #199: Home Keeper fails against the Home Assistant beta
Nightly Home Assistant `beta` run failed: https://github.com/prestomation/ha-home-keeper/actions/runs/31577647335

prestomation:
Fixes #199

## There is no Home Assistant regression

The nightly filed #199 having tested nothing. Home Keeper is fine against the beta — verified locally:

- `bash ci/test-python-integration.sh` against `HA_TAG=beta` (currently 2026.8.1): **99 passed**
- `mypy custom_components/home_keeper` against HA 2026.8.1: **Success, no issues in 39 source files**

Three CI bugs produced the false alarm. The second had been quietly degrading a PR gate for months.

## 1. The version print that killed the nightly

`homeassistant.__version__` does not exist and never has — `homeassistant/__init__.py` is a one-line docstring, and the version lives in `homeassistant.const`. So the "Report the version actually tested" step failed with an `AttributeError` on the nightly's **very first scheduled run** (`run_number: 1`).

That step sits *before* the real work, so it took down:

- the `integration` job, before `ci/test-python-integration.sh` ran
- the `typing` job, before `mypy` ran

`e2e` and `upgrade` have no such step and passed — which is why only two jobs are red in the run.

Fixed by reading `homeassistant.const.__version__`, and by marking the informational steps `continue-on-error: true` so a diagnostic can never again abort the suite it precedes.

## 2. Both mypy jobs were checking a six-month-old Home Assistant

HA 2026.3 raised its floor to Python `>=3.14.2`. Both mypy jobs pinned Python 3.13, so `pip install homeassistant` **did not fail** — it backtracked to `2026.2.3`, the last release supporting 3.13. The install log in the failing run shows exactly that.

The consequence is worse than the nightly outage:

- `ha-beta.yml` → `typing`, the job whose entire purpose is early warning about upstream API changes, was type-checking an HA **older than the one users run**.
- `lint.yml` → `mypy`, a gate on **every PR**, was doing the same.

Bug 1 masked this: a working version print would have shown `HA 2026.2.3` in the log.

Fixed by moving both jobs to Python 3.14 and adding `ci/check-ha-version.py`, which compares the resolved version against PyPI and fails loudly on a stale resolve rather than passing quietly. `--pre` selects the pre-release channel for the nightly. A PyPI outage reports the version and skips the check — a network blip must not redden CI.

## 3. mypy could not parse Home Assistant at all

`[tool.mypy] python_version` was `3.13`. HA 2026.8's source uses PEP 758 parenthesis-free `except A, B:`, which mypy cannot parse when targeting an older version:

```
homeassistant/config_entries.py:903: error: Except expressions without parentheses are
only supported in Python 3.14 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

"errors prevented further checking" means **none of our code was checked**. So even with fix 2 alone, the gate would have stayed hollow. Targeting 3.14 makes it pass cleanly with 39 files actually checked.

## Verification

| Check | Result |
| --- | --- |
| Integration suite vs `HA_TAG=beta` (2026.8.1) | 99 passed |
| `mypy` vs HA 2026.8.1, `python_version = 3.14` | clean, 39 files |
| Unit tests | 779 passed, 1 skipped |
| Guard vs current HA / stale 2026.2.3 / PyPI outage | exit 0 / exit 1 / exit 0 |
| Both workflow YAMLs parse | ok |

## Notes

- The new guard's release selection is pure and unit-tested (`tests/unit/test_ci_check_ha_version.py`, 8 cases covering prerelease channels, numeric-vs-lexicographic ordering, yanked and fileless releases). A version check that is itself wrong fails exactly as silently as the bug it guards against.
- All three lessons are recorded as conventions in `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md`, per the project rule that a convention is not real until it is written into the rules.
- No `CHANGELOG.md` entry and no beta bump: this is developer-only CI/config work with no user-facing change. No panel UI touched, so the screenshot and walkthrough gates do not apply.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TCiNZxVtUSoXzH791Vz5Z4)_

### DIFF
--- a/.amazonq/rules/testing-and-workflow.md
+++ b/.amazonq/rules/testing-and-workflow.md
@@ -186,6 +186,18 @@ gate); CI publishes it to the job summary.
 - Run it locally before pushing: `pip install mypy homeassistant && mypy
   custom_components/home_keeper`. The pure modules (`models.py`, `recurrence.py`,
   `events.py`) stay HA-free and type-check standalone.
+- **Run it on a Python at or above Home Assistant's own floor** (>=3.14.2 since HA
+  2026.3). On an older interpreter `pip install homeassistant` does not fail — it
+  backtracks to the last HA that supported it, so mypy checks a months-old API and
+  passes. Both mypy jobs therefore run `python ci/check-ha-version.py`, which
+  compares the installed version against PyPI and fails on a stale resolve (#199).
+- **`[tool.mypy] python_version` tracks HA's floor, not the integration's.** HA's
+  source uses syntax from its own minimum Python (2026.8 uses PEP 758
+  parenthesis-free `except A, B:`); targeting anything older makes mypy bail on a
+  syntax error inside HA, reporting nothing about our code.
+- **Read the HA version from `homeassistant.const.__version__`.**
+  `homeassistant.__version__` does not exist — `homeassistant/__init__.py` is a
+  one-line docstring — so a step reading it dies with an `AttributeError`.
 
 ## Quality scale
 - Home Keeper targets **Platinum** (`manifest.json` `quality_scale`), with the
--- a/.github/workflows/ha-beta.yml
+++ b/.github/workflows/ha-beta.yml
@@ -66,9 +66,13 @@ jobs:
           (cd tests/integration && docker compose logs)
           exit 1
 
+      # Informational only: `continue-on-error` so a broken diagnostic can never
+      # again abort the suite it precedes, which is exactly how #199 hid a whole
+      # nightly's worth of signal behind a one-line version print.
       - name: Report the version actually tested
+        continue-on-error: true
         working-directory: tests/integration
-        run: docker compose exec -T homeassistant python -c "import homeassistant;print('HA',homeassistant.__version__)"
+        run: docker compose exec -T homeassistant python -c "from homeassistant.const import __version__; print('HA', __version__)"
 
       - name: Run integration tests
         run: bash ci/test-python-integration.sh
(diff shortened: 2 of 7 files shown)


prestomation:
/q review Be critical and skeptical — surface the most serious issues first, and don't withhold minor ones. This PR claims the #199 nightly failure was three CI bugs and not a Home Keeper regression, so please attack that claim as well as the code.

**Correctness of the diagnosis.** Is it actually true that `homeassistant.__version__` never existed and `homeassistant.const.__version__` is the right accessor across the HA versions this repo tests (including the frozen pre-split `2026.7` tag used by `tests/upgrade`)? Is the pip-backtracking story right — could HA 2026.2.3 have been resolved for some other reason (a transitive pin, the runner's pip version) that my fix does not address? Am I wrong that `python_version = "3.13"` made mypy check nothing, or does mypy still check our own files after a syntax error in a followed import?

**`ci/check-ha-version.py`.** Hunt for ways this guard passes when it should fail, which would recreate exactly the silent-degradation bug it exists to prevent. Specifically: the comparison `installed < latest` against PyPI's newest — is there a legitimate case where a correct install is *older* than PyPI's newest and this goes red for no reason (a release published mid-job, a yanked-then-restored version, HA publishing a patch for an older branch)? Is the exception tuple around `urlopen` broad enough that a real failure is swallowed as "network blip", or narrow enough that an unexpected error crashes the step? Should the stable channel's "newest" really be the global max, given HA sometimes patches older minors?

**The `continue-on-error` decision.** I made the version-report steps non-fatal in `ha-beta.yml` but kept the typing job's check fatal. Is that split defensible, or does `continue-on-error` on the integration job now hide a genuinely broken container (e.g. HA failing to import at all) that we would want to catch?

**Python 3.14 bump.** `lint.yml` and the nightly's typing job now request `python-version: "3.14"`. Will GitHub's runner image resolve that to >=3.14.2 as HA requires, or could it land on an earlier 3.14.x and reintroduce the backtrack? Should this be pinned more precisely? Also: `pyproject.toml` still declares `requires-python = ">=3.12"` and ruff `target-version = "py312"` while HA itself now needs 3.14.2 — is leaving those alone right, or is it an inconsistency worth resolving here?

**Maintainability / HA best practices.** Does the new script belong in `ci/` given the repo has no other tested CI helper, and is `tests/unit/test_ci_check_ha_version.py` loading a hyphenated filename via `importlib` a pattern worth keeping or a smell? Are the conventions I added to `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md` accurate and non-duplicative with what's already there?

---

PR #203, opened 2026-08-16
feat: tell companions which completion an undo removed

prestomation:
## Why

Completing a task travels to a contributing integration. Undoing that completion did not.

A user completed a Pawsistant-managed pet-care task in Home Keeper by mistake, undid the completion here, and the mirrored "gave medicine" entry stayed behind in Pawsistant. The reason is in the payload: `store.delete_completion` fired `home_keeper_task_uncompleted` as a bare task spine, so a listener could tell that *a* completion went away but not *which one*, and had nothing to match its own record against.

Its two siblings already do this correctly — `update_completion` and `move_completion` both merge `extra={"ts": ...}`. Only the undo path was missing it.

This is the Home Keeper half. The Pawsistant half (subscribing to the event, plus the reverse direction) lands in a follow-up PR against `prestomation/Pawsistant` and depends on this being on `main` first, since its cross-integration suite installs Home Keeper from `git+…@main`.

## What changed

- **`delete_completion` fires the removed `ts`** and takes the same optional `origin` marker `complete_task` already accepts, echoed on the event so a caller can ignore the echo of an undo it initiated. Service schema, `services.yaml`, and `strings.json` + all 16 locales.
- **Undoing a `ts` that isn't in the history no longer fires.** `recurrence.remove_completion` has always been a documented no-op for an unknown timestamp, but the event fired regardless. Now that the payload *names* a timestamp, announcing it would tell a listener to drop a record for a completion that never existed.
- **`move_completion` explicitly keeps `origin: None`** on both events it fires. Moving is a user edit from the panel; an integration re-times its own mirror with `delete_completion` + `complete_task`, both of which already carry the marker. No new service surface for it.
- **`testing.py`** gains a `delete_completion` handler and `fire_user_uncompletion`, so glue authors can drive the undo path against the published fake.
- **Docs.** `EVENTS.md` documents the new payload fields. `INTEGRATING.md` gains an "Undoing a completion" section under loop prevention, and drops a stale claim in §5 that no event fires when a task is deleted in Home Keeper — `home_keeper_task_deleted` has fired on every deletion path for a while.
- Version bumped to `0.14.0b1` with a matching CHANGELOG section.

## Compatibility

Additive. `origin` is optional on `delete_completion`; existing callers and the panel's websocket path are unchanged. The only behaviour change is the silent no-op case, which previously fired an event carrying no actionable information.

## Testing

- `pytest tests/unit` — 780 passed, 1 skipped. New: `test_uncompleted_spine_carries_ts_and_origin`.
- New integration tests in `tests/integration/test_events.py`: the undo payload's `ts` round-trips exactly against the completion's `completed_at`, `origin` and `source` ride along, and an unknown `ts` fires nothing.
- `ruff check` / `ruff format --check` clean; `test_translations_parity.py` clean (80 passed) across all 16 locales.
- `vale` clean on the diff-scoped lines it gates.
- **mypy was not run locally.** This environment tops out at Python 3.13 and Home Assistant requires ≥ 3.14.2, so a local install would silently backtrack to a stale HA — exactly the failure mode `ci/check-ha-version.py` exists to catch (#199). Leaving it to CI. The change is type-simple: one keyword-only `origin: str | None = None` parameter and a dict literal.
- Docker integration tests not run locally; CI covers them.

https://claude.ai/code/session_01JpmQHyKb3YijDbVENaMER6

---
_Generated by [Claude Code](https://claude.ai/code/session_01JpmQHyKb3YijDbVENaMER6)_


prestomation:
/q review Please be critical and skeptical. This changes a public cross-integration event contract, so I want the most serious problems first, and no minor ones withheld.

**Correctness — the areas I am least sure about:**
- `store.delete_completion` now short-circuits when `ts` is not in `completions[]`, returning the *existing* task without saving or firing. Is returning the un-copied `existing` dict (rather than a copy) a mutation hazard for any caller? Does any current caller rely on `delete_completion` always persisting or always firing?
- The membership check compares `ts` by exact string equality, matching what `recurrence.remove_completion` does internally. Is there any path where a stored `ts` and a caller-supplied `ts` differ only by timezone-offset serialisation (`+00:00` vs `Z`, microseconds present or absent) so that the check and the removal disagree, or so a legitimate undo is silently dropped? DST and offset-less input especially.
- `move_completion` now passes an explicit `origin: None` in the uncompleted event's `extra` where previously the key was absent entirely. Does any consumer distinguish "key missing" from "key present and None"?
- The uncompleted event now carries `ts`, but `move_completion` and `delete_completion` reach it by different routes. Are the two payloads genuinely the same shape?

**Maintainability / API design:**
- I deliberately did *not* add `origin` to `move_completion`'s service schema, on the reasoning that a move is a user edit and an integration re-times its mirror via `delete_completion` + `complete_task`. Is that the right call, or is it an inconsistency that will bite someone?
- `testing.py`'s `_uncomplete` duplicates the store's no-op guard. That duplication is deliberate (the fake must mirror behaviour) but it is a drift risk — is there a better way to share it?

**HA best practices:** service schema and `services.yaml` / `strings.json` parity, and whether the optional `origin` field is modelled the way HA expects.

**Also please check:** whether anything else in the codebase fires `home_keeper_task_uncompleted` and now has an inconsistent payload, and whether the docs changes in `INTEGRATING.md` describe behaviour that actually matches the code.

---

prestomation:
Thanks — triaged. One fix pushed (8660b6e), and two corrections.

**Fixed: un-copied `existing` on the no-op branch.** Fair catch, and the reason I'm taking it isn't quite the one given. `get_task` and `list_tasks` already hand out live dicts from `self._tasks`, so this wasn't a *new* escape route into store internals. But within `delete_completion` the two branches disagreed: the removal path returns a copy and the short-circuit returned the live dict. That inconsistency is the actual defect — it makes the no-op the one branch where a caller could mutate the store by accident. Now `dict(existing)`, with a comment saying why.

**Correction — `move_completion` is registered as a service.** The review says *"The service schema omission prevents an integration from calling `move_completion` at all (it's not registered as a service), so there's no way for them to correct timestamps."* It is registered (`__init__.py`, `MOVE_COMPLETION_SCHEMA`, documented in `services.yaml`), and has been. What this PR declines to add is the optional `origin` field on it, which is a much smaller gap: an integration can call `move_completion` today and will simply see the resulting events with `origin: None`, exactly as it would for a user edit from the panel.

That said, the underlying design question is fair, so to be concrete about the trade-off. The atomicity concern cuts the other way from what's suggested: `move_completion` is *internally* modelled as "undo, then redo" (`store.py`, "Modeled as 'undo, then redo at the new time'"), and it fires the same two events either way, so `delete_completion` + `complete_task` is not a degradation — it's the same sequence with a marker attached. The metadata point is real but narrow, and re-timing a mirrored record is not something a companion should be doing often. Adding `origin` there also means a third round of translated service strings across 16 locales for a field with no known consumer. I've documented the decision in `INTEGRATING.md` rather than widening the surface; happy to add it if you'd rather have the symmetry.

**Correction — `services.yaml` and `strings.json` are in the diff.** The review says it couldn't verify parity because they weren't included. Both are in the changeset, along with all 16 `translations/*.json`; `tests/unit/test_translations_parity.py` passes (80 tests), which is the gate that enforces key-structure parity and catches untranslated leaks.

**Noted, keeping as-is:** the explicit `origin: None` in `move_completion`'s `extra`. It is redundant in the sense that `task_event_data` would omit the key entirely — that's precisely why it's there, so both fire sites of `home_keeper_task_uncompleted` produce the same key set rather than one payload having the key and the other not. Item 4 of the review confirms the shapes now match, which is the property I wanted.

---

PR #205, opened 2026-08-17
feat(panel): let a task be assigned to an area

### LINKED ISSUE #204: Assign task to area
Maybe I’m overlooking something, but I haven’t found a way to assign a task to an area, although the UI suggests that it’s possible. How to do that, if the task is not associated with a device?

prestomation:
Fixes #204

## The gap

`area_id` has always been a first-class task field. It's normalized in `models.py:377`, preserved by `merge_update`, exposed by `add_task` / `update_task` (`services.yaml:105`, `:260`), validated by the websocket API (`websocket_api.py:_area_ok`), and read by every consumer:

- `panel.ts:_taskAreaId` → **Group by → Area**
- `card-filter.ts:taskAreaId` → the card's area filter / sort / group
- `profiles.py:134` and `notifier.py:91-97` → profile matching and the notifications built on it
- `events.py:34` → the event payload

Tasks already get their own area through three paths: the services, asset reconcile (`reconcile.py:179`, a part-replacement task inherits its appliance's area), and problem-sensor sync — which lists `area_id` in `_LOCKED_FIELDS` precisely because unsynced tasks are supposed to offer it.

The panel's task form was the one place that never did. Resolution is own-area-first with the attached device's area as a fallback, so a task with a device looked fine; a task **without** one inherited nothing, sat under **Unassigned**, and no control existed to move it. That's what #204 hit.

## The change

- **`taskSchema` gains an Area picker**, next to the device picker on every task kind (including `triggered`), guarded by `locked.has('area_id')` so a synced problem-sensor task keeps its source-owned area read-only.
- **`taskFormData`** loads the saved area; **`buildTaskPayload`** always round-trips it, sending an explicit `null` when cleared so `merge_update` drops the old area instead of silently keeping it.
- **The task detail page grows an area chip** beside the device chip, showing the *effective* area (own, else device's) so the page explains which Group-by-Area section the task landed in. Absent when the task is unplaced, matching how the device chip stays absent.
- **`help.area_id`** added to all 16 locales (`field.area_id` was already translated everywhere), matching the existing per-field helper pattern.

## Screenshots

The Area picker in the task form, holding the room a device-less task was placed in:

The task edit form for 'Take medicine', showing an Area field set to Living Room below the device picker, with helper text explaining it falls back to the attached device's area

The detail page naming the area next to the status chip:

Task detail page for 'Take medicine' with an Overdue chip and a Living Room area chip beside it

Grouped by Area — the task now sorts into its room instead of Unassigned:

The Tasks tab grouped by Area, with a Living Room section containing one task above an Unassigned section containing the rest

The empty picker on a new task, with its helper text:

The new-task form showing an empty Area field below the device picker

Screenshot churn is deliberately limited: every task-form shot picked up the new field, and the three `42*` shots are new. Shots that changed only by date drift were reverted to keep the diff reviewable.

## Testing

- `ci/test-frontend.sh` — 457 passing (18 files). New: 9 cases covering the picker across every task kind, the locked-area cases (floating and triggered), form load, payload round-trip and the explicit-null clear; 4 panel cases for the chip (own area, device fallback, own-over-device precedence, unplaced).
- `ci/test-mutation-frontend.sh` — **100%** (12/12) on the changed `forms.ts` surface. The first run scored 75%: the locked-case assertions went through a `names()` helper that drops nameless entries, so a locked branch emitting junk instead of `[]` read as a pass. Fixed by counting raw schema entries and adding the triggered-locked case, in a separate commit.
- `ci/test-mutation-python.sh` — no mutable Python changed.
- `ci/e2e-up.sh` — 44 e2e passing against the panel change; the screenshot harness gained a step for the new surface.
- `ruff check` / `ruff format --check` over `custom_components tests ci` — clean. `tsc --noEmit` — clean.
- Vale (`ai-tells`), diff-scoped to added lines in `README.md` / `CHANGELOG.md` / `locales/en.json` — 0 hits.
- `mypy` not run locally: this sandbox is on Python 3.11, below HA's 3.14.2 floor, so `pip install homeassistant` would silently backtrack to a stale HA and check an API nobody runs (#199). The only Python change in this PR is the `PANEL_VERSION` string.

## Release

Bumped to `0.14.0b2` (`manifest.json` + `const.py`) with a matching CHANGELOG section — `0.14.0b1` is already tagged, so this opens the next beta. Needs the `preview-release` label.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TP87ayHZKi2GJAdMWGZ9Yi)_


prestomation:
/q review Be critical and skeptical. This adds an Area picker to the panel's task form (`forms.ts` `taskSchema`/`taskFormData`/`buildTaskPayload`) plus an area chip on the task detail page (`panel.ts`). Surface the most serious issues first, and don't withhold minor ones.

Please scrutinize specifically:

**Correctness / edge cases**
- `buildTaskPayload` now sets `payload.area_id = task.area_id || null` unconditionally for every task kind, including `triggered` (whose payload is deliberately restricted to descriptive fields). Is sending `area_id` on a triggered task right, or can it disturb a dormant task's state?
- The `|| null` coercion: is an explicit `null` the correct clear signal for `merge_update`, and does it interact safely with `normalize_fields`' `data.get("area_id") or None`? Any path where a cleared area silently persists, or where `null` is written when the user never touched the field?
- Locked fields: the picker is hidden when `area_id` is locked, but `buildTaskPayload` still emits `area_id` from the edit state. `merge_update` strips locked keys — is relying on that server-side strip correct, or should the client omit it? Same question for the pre-existing `device_id` behaviour.
- A dangling `area_id` (area deleted in HA after assignment): `areaName()` falls back to printing the raw id, so the chip and the picker would show a UUID. `devices.py` self-heals this for assets — should tasks?
- The detail chip uses the *effective* area (own, else the device's). Is that the right choice versus showing only the task's own area, given the card row (`card.ts:809`) reads `task.area_id` directly and falls back to the device *name*? Is that pre-existing inconsistency now more visible?

**Maintainability**
- `_areaChip` in `panel.ts` versus the existing `_deviceChip`/`_managedChip` — right level of duplication?
- The area field is now declared in two branches of `taskSchema` (triggered and the main list), mirroring `device_id`. Worth factoring out?

**HA best practices**
- `selArea()` single-select and the `mdi:texture-box` chip icon: are these the conventional HA choices? Should the chip link to the area's config page the way `_deviceChip` links to the device?

**Performance** — the chip resolves through `this._hass?.areas` per render; any concern on large task lists (the chip is detail-page only today)?

**Test quality** — the mutation run initially scored 75% because locked-case assertions went through a `names()` helper that drops nameless entries. Fixed in `1b1a4f8`. Are there other assertions in the new tests that pass without actually constraining behaviour?

---

prestomation:
Correction to the review request above: the test-hardening commit is `09366b0b`, not `1b1a4f8` (which doesn't exist in this branch).

---

prestomation:
/q review Pushed `3ec36b3d`, fixing the `preview` (docs-site) failure: the new README section links to `#dashboard-task-card`, and `sync-docs.mjs` splits the README into one guide page per `##` section, so that cross-section anchor dangled and Docusaurus failed the build. Added it to `ANCHOR_ROUTES`. Verified with a local `npm run build` in `website/`.

Be critical and skeptical on this follow-up, and re-examine the questions in my earlier request that the first pass didn't reach — that review only covered the `panel.ts` area chip and didn't touch `forms.ts`, which is where the behavioural risk is. Most serious issues first, minor ones included.

Specifically:
- `buildTaskPayload` sets `payload.area_id = task.area_id || null` for **every** task kind including `triggered` (whose payload is otherwise deliberately restricted to descriptive fields). Correct, or can it disturb a dormant task?
- The `|| null` clear signal against `merge_update` + `normalize_fields`' `data.get("area_id") or None` — any path where a cleared area silently persists, or a `null` is written when the user never touched the field?
- Locked fields: the picker is hidden when `area_id` is locked but the payload still carries it, relying on `merge_update` stripping locked keys server-side. Right call, or should the client omit it?
- Dangling `area_id` (area deleted in HA after assignment): `areaName()` falls back to the raw id, so chip and picker would show a UUID. `devices.py` self-heals this for assets — should tasks?
- Are there other `ANCHOR_ROUTES`-style cross-section links in the README that are currently dangling and only avoided by luck?

---

PR #206, opened 2026-08-17
Make the privilege model hold: administration is admin-only, usage is not

prestomation:
A security review found six issues. The headline one: **the two places Home Keeper tried to draw a privilege line didn't actually hold**, because each was gated in the websocket API and left open in the matching service. All six are addressed, plus two more found while fixing them.

The shape of the fix, decided in the review discussion: **managing your home is an admin job, using it is not** — the same line Home Assistant draws around its own Settings, and the one `docs/DESIGN.md` already claimed Home Keeper drew.

## What changed

**1. The service-side bypass (the actual vulnerability).** `ws_set_options` and `ws_export_inventory` carry `@websocket_api.require_admin`, but both operations were also registered as plain services with no check on `call.context.user_id`. Any signed-in non-admin got the identical result — serial numbers, purchase costs and inventory value totals included — by calling the service over the same websocket connection. Service handlers now go through `_verify_admin`, which raises HA's `Unauthorized`. A call carrying no user (automation, internal) stays trusted, matching core.

**2. The panel is `require_admin=True`.** Appliance CRUD is gated on both surfaces alongside it — those commands each echo the full asset back in their result, so leaving them open would have handed out exactly what the export gate exists to protect, and creating an appliance writes the device registry, which core reserves for admins. Non-admins keep the to-do list, the calendar, the device-page entities and the dashboard card, and can still complete, snooze, skip and add tasks.

**3. Appliance reads are projected, not gated.** The dashboard card resolves task card-links against appliance data, so `get_assets` / `list_assets` stay open; a non-admin gets `assets.card_projection` — documents, `link`-type metadata, and a part's id/name/url. Costs, serials, warranty dates and free-text custom fields are withheld. It's a whitelist, so a field added later is private until someone publishes it deliberately.

**4. Notify targets are enforced where they're stored,** not only in the picker. `mobile_app_*` and `persistent_notification` are accepted; anything else is dropped with a warning, or fails the call when passed explicitly as `target:`. Otherwise a saved notification could relay caller-controlled text through whatever SMTP or chat integration an admin configured. `persistent_notification` is allowlisted because it never leaves the instance — and it's a real use, which the localization integration test relies on.

**5. Only built assets are published.** Rollup emits to `frontend/dist/`; `panel.py` mounts that. HA serves static paths *before* authentication, so the old mount exposed `src/`, `test/`, `node_modules/` and `package*.json` to anyone who could reach the port. HACS installs were unaffected (the release archive already excluded those); git-clone installs were not. Served URLs are unchanged, so cached Lovelace resources and the `?v=` cache token still work.

**6. A signed URL can't become an upload primitive.** Verified against the HA the suite actually runs (2026.8.2): `homeassistant/components/http/auth.py` only considers a signed request when `request.method in ["GET", "HEAD"]`, so a signed POST never authenticates. The upload views additionally refuse any request not authenticated as a real (non-system) user, as a belt for the day that changes.

**7. Consistent link guards.** `safeFileHref` — http(s) **or** site-relative, so server-minted signed URLs survive where `safeHref` would blank them — replaces bare `escapeHTML` at the document, part-file and card-chip anchors; part product URLs use `safeHref`, and pick up the missing `noreferrer`. Escaping alone doesn't defuse a `javascript:` URI, and chips arrive from other integrations through `add_task`.

## Screenshots

The changed surfaces are the document, part-file and product-link anchors on the appliance page. Both still render and still carry their hrefs (the file chip's signed, site-relative URL is exactly what `safeHref` would have blanked).

Appliance page showing the manuals and documents list with link and file documents, and a part carrying an attached receipt

Appliance page after uploading a part file, showing the part's attached receipt card with its open-in-new link

## Testing

- **`tests/integration/test_admin_gates.py`** (new) creates a real non-admin user through HA's auth API and drives both surfaces with both tokens: the admin-only services refuse it, appliance mutations refuse it, `get_assets`/`list_assets` come back projected, the panel is absent from its `get_panels`, and it can still complete a task. The gates rest on framework contracts a mocked unit test cannot see — which is how the service bypass survived in the first place.
- **`test_signed_url_cannot_be_used_to_upload`** mints a signed URL, confirms the GET still works, and asserts the POST is refused.
- Unit tests for `card_projection` (including that it doesn't mutate the store's dicts) and the target allowlist; frontend tests for `safeFileHref`.
- Full suites run locally against HA 2026.8.2: 793 unit, 461 frontend, 116 integration, 44 e2e. The e2e run matters here — `documents.spec.ts` asserts the signed hrefs are present before any click, which is the direct check that the new guard didn't render every file link inert.
- Mutation gate: Python 95.24%, TypeScript 100%.

**mypy was not run locally.** This environment's newest Python is 3.13, below HA's 3.14.2 floor, so any locally-installed HA is the stale resolve `ci/check-ha-version.py` exists to catch. Leaving that to CI.

## Also in here

- **`docs/SECURITY.md`** (new, wired into the Developer Guide and Vale): what's admin-only, what's open, what a non-admin reads instead, and what this explicitly doesn't defend against. Linked from `README.md` and `docs/DESIGN.md`; the convention is recorded in `.amazonq/rules/architecture-and-code.md` and `AGENTS.md` so the "gate both halves" rule survives the next feature.
- **A fix to the mutation job itself.** mutmut runs the suite inside a `mutants/` copy holding only the source paths and tests, but `tests/unit/test_ci_check_ha_version.py` loads `ci/check-ha-version.py` by path, so every run died at stat collection. It went unnoticed because the two PRs since that test landed changed no mutable Python, so the job reported "nothing to test" and went green. `also_copy = ["ci"]`.

## Notes for review

- **This is a behaviour change for households with non-admin users.** They lose the panel. That was the deliberate choice over the alternatives (gating operations while keeping the panel visible, or documenting the flat model as-is).
- **No version bump.** These are fixes, so the notes are under `## [Unreleased]`; `0.14.0b2` is already published.
- One deviation worth naming: the non-admin projection keeps `link`-type metadata entries. The card resolves "show on card" links against them (`card.ts:754`), so dropping all custom fields would silently break that feature. Text and date entries — where warranty dates and prices live — are dropped.

https://claude.ai/code/session_016nfZVNCtB71n9KYD1Yt5T7

---
_Generated by [Claude Code](https://claude.ai/code/session_016nfZVNCtB71n9KYD1Yt5T7)_


prestomation:
/q review Be critical and skeptical — this PR changes a privilege boundary, so a missed hole is worse than a false positive. Lead with the most serious findings, and don't withhold minor ones.

**Completeness of the gate.** Is there any remaining path by which a non-admin reads full appliance data or performs an administrative action? Check every websocket command in `websocket_api.py` and every service in `__init__.py` against the intent: appliance CRUD + `set_options` + `export_inventory` are admin-only; task usage, `sign_document_url`/`sign_part_file_url` and the projected reads stay open. Two specific worries: (a) the ungated sign-URL commands mint a bearer URL for any document on any appliance — is that acceptable given the card needs it, or does it leak file *existence*? (b) do any events, entity attributes, or the `todo`/`calendar`/sensor entities expose cost/serial data that the `card_projection` withholds, making the projection moot?

**The projection itself** (`assets.card_projection`). Is it genuinely non-mutating of the store's dicts (it shallow-copies documents and metadata entries by reference — can a caller mutate a document in place)? Does it drop anything the dashboard card actually needs, in any code path other than `_resolveDocuments`/`_resolvePartLink`? Is including `link`-type metadata the right call?

**`_verify_admin` correctness.** Is treating `context.user_id is None` as trusted right for every path a service call can arrive by — REST, websocket `call_service`, script/automation, `hass.services.async_call` from another integration? Could a non-admin cause a service to be invoked with a null-user context and thereby bypass the check? Is `Unauthorized` the correct exception, and does it surface sanely through both the REST and websocket layers?

**Notify allowlist edge cases** (`notifications.split_targets`, `notifier.async_run_notify`). Prefix matching, casing, whitespace, non-string members, and the interaction between the silent-drop path (stored notifications) and the raise path (explicit `target:`). Does an empty result after filtering produce a sensible error rather than a confusing "no targets" one?

**The `frontend/dist/` move.** Anything still referencing the old bundle location — CI, the release zip, the preview-release workflow, HACS metadata, the upgrade-test staging in `ci/fetch-glues.sh`? Any upgrade-path problem for an existing install whose `frontend/` still holds a stale `home-keeper-panel.js`?

**HA best practice.** Is `require_admin=True` on the panel plus per-command gating the idiomatic way core does this, or is there a framework mechanism I've reimplemented?

---

prestomation:
Pushed `100ffd8`: fixed the ruff and vale failures, and took the one substantive point from the Cue review.

**ruff** — two E501s in tests I added after my last local lint run. My miss, not a version difference.

**vale** — the new prose in `CHANGELOG.md`, `README.md`, `docs/DESIGN.md`, `docs/SECURITY.md` and `services.yaml` rewritten to clear the ai-tells style (tricolons, em dashes, `accordingly`/`additionally`/`therefore`, `genuinely`, and the "What this does not protect against" explainer heading, now "Outside the boundary"). Verified locally by mirroring only this PR's added lines into a scratch tree and running vale against them, since the gate is diff-scoped.

**On the sign-URL existence disclosure** — agreed, and now written down rather than assumed. `docs/SECURITY.md` says it plainly: minting a signed URL isn't admin-gated because the card needs one to open a document on a task anyone can complete, so a non-admin who *guesses* an appliance/document id pair learns whether it exists. Guessing is the operative word — the narrowed appliance view only lists documents already surfaced on a card, so there's nothing to enumerate from.

The rest of the review I read as confirmation rather than action: no bypass paths found, `_verify_admin`'s null-user handling correct, the projection's whitelist shape and copying behaviour sound, notify edge cases covered, `dist/` references complete.

One correction to the review's reading, since it affects what a future reader trusts: it says the sign-URL commands are safe partly because "the signed URLs are GET-only (verified in upload views)". The upload views aren't what makes that true — Home Assistant's auth middleware is, and it only considers a signature when `request.method` is `GET` or `HEAD` (`homeassistant/components/http/auth.py`, confirmed against 2026.8.2). The check I added to the upload views is a second layer for the day that changes, not the primary control. `test_signed_url_cannot_be_used_to_upload` pins the real behaviour.

---

PR #207, opened 2026-08-18
feat(sensor tasks): drive a task from a binary sensor with a new State mode

prestomation:
## The gap

Sensor-based tasks only understood **numbers**. `sensor_tasks.parse_reading()` returns `None` for anything non-numeric and the watcher skips a task with no reading, so a binary sensor could never arm one.

The panel's entity picker is unfiltered, though (`selEntity({})` in `forms.ts`), so you could select `binary_sensor.robot_water_tank_low`, save the task, and have it **silently never fire**. A dead end that looked like a supported path.

Plenty of hardware never publishes a number:

- a robot vacuum reporting **water tank low**
- a device reporting **`battery_almost_empty`** instead of a battery percentage
- a leak detector, a *filter needs replacing* flag

The existing **problem-sensor sync** doesn't reach these either: it only mirrors `device_class: problem`, all-or-nothing, as tasks you can't complete by hand. `battery_almost_empty` is `device_class: battery`, so the sync never sees it.

## What this adds

A third sensor mode, **`state`**, that compares the entity's state *string*.

- **Arms on the transition into the state**, not on the state being true. A task appears once per event rather than repeating while the sensor stays on, and after you complete it it re-arms only on a fresh crossing. "Fill the water tank" shows up once per empty tank, not every five minutes until you get to it.
- **A restart doesn't fabricate work.** `async_baseline` records an already-matching sensor as met-without-a-crossing, so a task you already dealt with doesn't come back.
- **Not binary-only.** The comparison is textual, so `vacuum.rosie` = `docked` and `sensor.washer_status` = `finished` work identically. The form offers **On**/**Off** for a `binary_sensor` and free text otherwise.
- The existing **hold** applies, for *"the door has been open for 10 minutes"*.

Plus **`clear_on_recover`** (opt-in, off by default) on both `state` and `threshold`: an armed task clears itself once the condition goes away, applied by the watcher as a real `complete_task` tagged `ORIGIN_SENSOR_RECOVER` so the history still shows the work and an automation can tell it from someone pressing Done.

## Design note: one edge, two modes

`threshold` and `state` differ **only** in how "is the condition true right now" is computed. Rather than a second copy of the rising-edge machinery (hold timer, consuming a crossing so a steady-true sensor never re-arms), that logic moves into a shared `sensor_tasks._evaluate_edge` that both delegate to. `clear_on_recover` is then one branch in that helper, which is why it lands for threshold mode at no extra cost — the `clear_on_recover` seam `docs/SENSOR_TASKS_PLAN.md` §8 already named.

**A missing reading is not a recovery.** `unavailable`/`unknown` holds the carried edge state and decides nothing. Reading a dropout as "the condition went away" would silently complete every `clear_on_recover` task the first time its device fell off the Zigbee mesh. There are unit *and* integration tests pinning that specifically.

## Screenshots

State mode bound to a binary sensor — the value control is an On/Off picker, and the rule reads back above the button:

Creating a state-mode sensor task bound to binary_sensor HK demo water tank low, with Trigger state set to On, a Hold field, a Clear when back to normal toggle, and the summary reading When it changes to on

Threshold mode, which picks up the same **Clear when back to normal** toggle:

The threshold-mode sensor task form showing comparison, threshold value, hold, and the new Clear when back to normal toggle

## Testing

Everything below was run locally, not left to CI.

| Gate | Result |
|---|---|
| `pytest tests/unit` | 839 passed |
| `vitest` (frontend) | 482 passed |
| `tests/integration/test_sensor_watcher.py` (real HA container) | 13 passed |
| `tests/e2e` full suite | 44 passed |
| Screenshot capture harness | passed, both PNGs visually inspected |
| Walkthrough capture harness | passed with the new tour segment |
| `ruff check` / `ruff format` | clean |
| Vale (`ai-tells`, added lines only) | 0 hits |
| Mutation — Python | **90.19%** (threshold 80%) |
| Mutation — frontend | **88.50%** (threshold 80%) |

Six new **integration** tests drive a real template `binary_sensor` (following an `input_boolean`, added to the e2e HA config) so the off → on state-change event is genuine. They cover arming on the flip, not re-arming while it stays on, re-arming after a fresh crossing, a non-binary entity matching its state, `clear_on_recover` completing the task, and a config-entry reload with the sensor already on **not** resurrecting it.

`mypy` could not be run locally — Home Assistant no longer installs on this environment's Python 3.11 (PyRIC wheel build failure; HA now floors at 3.14). `mypy --ignore-missing-imports` over the three changed modules is clean, and CI runs the real gate.

## Also in this change

- `0.14.0b2` → **`0.14.0b3`** (`manifest.json` + `PANEL_VERSION`), with `## [Unreleased]` renamed to `## [0.14.0b3]` per AGENTS.md.
- New README section **"When a device just tells you"**, leading with the vacuum / `battery_almost_empty` cases and spelling out how this differs from problem-sensor sync.
- ~10 new i18n keys translated across **all 16 locales** (the parity test gates both key coverage and untranslated leaks).
- `docs/EVENTS.md` documents the new `home_keeper_sensor_recover` origin; `docs/SENSOR_TASKS_PLAN.md` records the mode and retires the `clear_on_recover` deferral.
- Two panel spots assumed "not threshold means usage" and would have rendered a `0 of 0` meter bar for a state task; both now allowlist usage explicitly.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RwjfHJrzt5gAJ2JrCTSXSC)_


prestomation:
/q review Please be critical and skeptical, and rank the most serious issues first — but don't withhold the minor ones either. This adds a third sensor-task mode (`state`) that arms a task from a binary sensor's `on`/`off`, plus an opt-in `clear_on_recover`. Specific things I want scrutinized:

**Correctness / edge cases.** The rising-edge machinery was extracted from `evaluate_threshold` into a shared `sensor_tasks._evaluate_edge` that both `threshold` and `state` now call — please check that refactor is genuinely behaviour-preserving for existing threshold tasks, especially the hold timer and the "consume the crossing so a steady-true sensor never re-arms" logic. Then attack the unavailable/dropout path: `evaluate_state` returns early on a `None` state to avoid treating a device falling off the network as a recovery. Is that early return complete, or can a task still be armed or cleared spuriously through some interleaving (dropout mid-hold, dropout while armed, flapping availability, an `attribute` binding whose attribute disappears while the entity stays available)? Also: is the startup baseline in `SensorTaskWatcher.async_baseline` airtight for `state`, and does creating a task while the condition is *already* true behave sensibly?

**`clear_on_recover` semantics.** It's applied as a real `store.complete_task(..., origin=ORIGIN_SENSOR_RECOVER)`. Is recording a completion the right call versus clearing without history, and does it interact badly with anything downstream — consumable draw-down, wear-part stamping, `completion_detail: required` tasks (does a required-fields task even accept an automatic completion?), or a task whose completion has side effects the user didn't authorize?

**Validation.** `models.normalize_sensor` became a three-way branch with hoisted `_normalize_for_seconds` / `_reject_fields` helpers, and the edge-driven modes now reject `target`/`baseline` in addition to `also_every`/`combinator`/`unit`. Could that new rejection break an existing stored binding or a service call that used to work — particularly through `merge_update`, which carries `existing["sensor"]` forward when an update omits it?

**HA best practices.** `read_sensor_state` returning `str(raw)` for attribute values, the state-change subscription, and whether comparing raw state strings is the idiomatic way to do this in HA.

**Maintainability / performance.** Module boundaries between `sensor_tasks` (pure) and `sensor_watcher` (HA-aware), the `_apply_edge` helper, and whether the frontend's `isBinarySensorBinding` re-render trigger in `panel.ts` is the right seam or a hack.

---

PR #219, opened 2026-08-21
feat(profiles): exclude tasks by label, area or device

### LINKED ISSUE #214: Exclusions in profiles
I wanted to setup a profile to list all tasks that can be done without calling a maintance person/mechanic.

It seems like there's no way to filter **out** by label/device, only include. Can exclusions be added?

Image

prestomation:
Fixes #214

A Profile's filter could only ever narrow *down to* a set of labels, areas or devices. Expressing "everything I can do myself" meant labelling every task that *isn't* a call-out, because there was no way to say "all of these, minus those".

## What changed

The profile filter gains `exclude_labels`, `exclude_areas` and `exclude_devices`, surfaced in **Settings → Profiles** as three pickers under the existing include pickers.

- **Exclusions apply after the include lists and win over them.** A task that cleared every include list is still dropped if it carries an excluded label, sits in an excluded area, or hangs off an excluded device.
- **An empty exclude list excludes nothing**, so profiles saved before this release behave identically. `normalize_filter` defaults the keys on read (`options.current_options` re-normalizes every read), so there is no migration and no storage-version bump — the same additive pattern `labels`/`areas`/`devices` already use.
- **Exclusions read the same *effective* ids as the include lists.** Excluding a label also drops a task that carries it only via its device or its area, matching how the include pickers already resolve inheritance.

Because a Profile is the shared saved filter, this reaches every consumer at once: the notifications built on it, the **Profile** dropdown on the Tasks tab, and the dashboard card's **Filter by profile**. The card's own inline `labels`/`areas`/`devices` YAML options are deliberately left include-only.

## Screenshot

Settings → Profiles, with the three new pickers and an area exclusion set on the "Upstairs" profile:

The Settings → Profiles card showing Exclude labels, Exclude areas and Exclude devices pickers beneath the include pickers, with Living Room set as an excluded area

## Parity

The filter has two independent implementations that must agree — `profiles.matches_filter` (Python, drives notifications) and `card-filter.profileMatches` (TypeScript, drives the panel list and the card). Both were changed together, and `tests/fixtures/profile_filter_cases.json` gains 11 exclusion cases replayed by **both** suites, so the two can't drift.

While mirroring the matcher I replaced the `?? ''` sentinel comparisons with a small `listHas` helper. It reads better and, more usefully, turns "a task with no area/device isn't swept up by a non-empty exclude list" into a real, testable branch instead of relying on a placeholder that can't collide with a real id.

## Testing

- `tests/unit/test_profiles.py` — exclusion matching, exclusion-beats-include, empty/absent lists excluding nothing, unset area/device surviving an exclusion, `due_queue` applying exclusions. Also closes two pre-existing assertion gaps the mutation run surfaced: nothing verified which input keys `normalize_filter` reads, and nothing verified that a filter with no `status` defaults to `overdue` rather than "everything".
- `frontend/test/card-filter-profile.test.js` — the same cases on the TS side, plus inherited-label and inherited-area exclusion.
- `frontend/test/forms-helpers.test.js` — schema shape, flatten/rebuild round-trip, and that each exclude row offers the same multi-picker as its include twin.
- `tests/integration/test_notifications.py` — a new case proving the exclusion survives the real `set_options` → `normalize_filter` → storage → notifier round-trip, and that `list_profiles` reports it back. `normalize_filter` rebuilds the block from a fixed key set, so a key it forgot would be dropped silently rather than raising — worth an end-to-end assertion.

All green locally: 900 Python unit tests, 557 vitest, 46 e2e, `ruff`, `mypy` (with HA installed), and Vale on the added prose lines.

**Mutation:** 100% Python (145/145) and 100% TypeScript (41/41), both against the 80% gate. One genuinely equivalent mutant is annotated at the source with its reason — the `if (!id)` guard in `listHas` exists to narrow the type for `includes`, and falling through returns `false` for those inputs anyway.

## Also in this PR

- Version bumped to `0.16.0b1` (`manifest.json` + `PANEL_VERSION`) with a matching CHANGELOG section, so the feature ships to beta testers via HACS. `0.15.0` shipped yesterday, so this opens the next minor's beta line rather than a `0.15.0bN`.
- README's Profiles section documents the exclude pickers, the precedence, and the inheritance caveat.
- `screenshots.capture.ts` seeds an area exclusion so the captured shot shows a populated picker rather than three empty ones.

Only `docs/images/profiles-card.png` is re-committed — the capture run re-encodes all 51 images, but that's the one whose content actually changed.

The walkthrough tour is untouched: this extends an existing surface rather than adding a new one, and the tour has never stepped through Settings → Profiles.

---
_Generated by [Claude Code](https://claude.ai/code/session_014W8gxfxGTiq9CBcKNRXAYf)_


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones. This adds exclusion filters (`exclude_labels` / `exclude_areas` / `exclude_devices`) to a Profile's saved filter. Areas I'd like scrutinised:

**Correctness and precedence.** Exclusions are meant to apply after the include lists and win over them, with an empty or absent list excluding nothing. Look hard for the inversion that would silently empty every profile at once, and for the placeholder-collision class of bug where a task with no area or no device gets swept up by a non-empty exclude list. Is `listHas` in `card-filter.ts` right for `null`, `undefined` and `''`? Does the interaction with the status window (`overdue` / `due_soon` / `all`) hold — can an exclusion ever rescue a task the status check already dropped, or vice versa?

**Cross-language parity.** `profiles.matches_filter` (Python) and `card-filter.profileMatches` (TypeScript) are two independent implementations of one spec, pinned by `tests/fixtures/profile_filter_cases.json`. Do they actually agree on the new semantics, including the effective-id resolution — the backend enriches tasks in `notifier._effective_filter_tasks` before matching, while the frontend resolves inheritance inline via `taskLabelIds`/`taskAreaId`? Are the 11 new fixture cases covering the divergences that matter, or are there asymmetries they'd miss?

**Backwards compatibility.** The claim is that no migration is needed because `normalize_filter` defaults the new keys on read and `options.current_options` re-normalizes every read. Is that actually true for every path that reads or writes a profile — including `home_keeper.notify` with an ad-hoc profile, `list_profiles`, `home_keeper/get_profiles`, and a profile written by an older panel bundle against a newer backend (or the reverse, during a partial upgrade)?

**Maintainability.** `normalize_filter` rebuilds the block from a fixed key set, so it is the de-facto allowlist — is that failure mode (a forgotten key is dropped silently rather than raising) adequately guarded now? Is the include/exclude duplication in the form schema and the matchers worth factoring further, or is the parallel structure the clearer choice?

**Performance.** `profileMatches` now computes `taskLabelIds` unconditionally rather than only when a label filter is set. That's a Set allocation per task per render on the card and the admin list — is it worth restoring laziness, or is it noise at realistic task counts?

**HA best practices.** The three new fields use `selLabel(true)` / `selArea(true)` / `selDevice(true)` selectors and are labelled via `computeLabel`'s `t('notify.' + name)` fallback. Anything off about the selector usage, the i18n key namespace, or the fact that the status field is already labelled "Include" while these are labelled "Exclude ..."?

---

PR #222, opened 2026-08-21
fix(notifications): let "Mark done" work before a task is due (Fixes #216)

### LINKED ISSUE #216: Home Keeper is ignoring "Mark Done" for any task with a future date?
I made a couple of one-off test tasks to test notifications. The notification arrives on my phone saying "Due Soon" and gives me action buttons. When I hit "Mark as Done" nothing happens. Snooze works, but mark as done doesn't. 

It now appears that this is only the case if the event is in the _future_ (the test events i scheduled for 1-3 minutes ahead). If I wait to press the button until the event is overdue, it works. But if I press the button before the due date it is ignored forever and the task becomes overdue.

Edit: I can now report it does the same thing for recurring tasks that it does with one-off tasks. Ignores the notification action completely and eventually shows as overdue. I monitored "mobile_app_notification_action" while pressing Done and it did receive the event so it appears it is Home Keeper ignoring it. I'm seeing nothing in the logs.

prestomation:
Fixes #216.

## The bug

A notification built from a Profile set to **Due soon** (or **All**) legitimately queues tasks whose `next_due` is still in the future, and `build_notification` attaches the full button set to them — `_action_button` never consulted due-state. But the tap handler gated `complete` on `recurrence.is_overdue`:

```python
task = coord.store.get_task(task_id)
if task is None or not recurrence.is_overdue(task, now=dt_util.utcnow()):
    return
```

So the button did nothing at all: no completion, no log, and the task went on to become overdue as if it had been left alone. **Snooze** and **Skip** on the very same card worked, because neither consults due-state. That asymmetry is exactly what the reporter described.

## Why not just drop the guard

`complete_task` appends a completion *and* advances `next_due` a full interval, so a duplicated tap silently skips a cycle — a quarterly task jumps six months out. Two ways that happens in practice: a double-tap, and a notification whose `targets` list covers several phones (`_send_payload` fans out to each). The guard was protecting something real. It was just using the wrong predicate: "is this task overdue" as a proxy for "is this button still valid".

## The fix

Check freshness directly. Each button now carries the task's `next_due` at send time as a fifth action-string field:

```
home_keeper::{verb}::{task_id}::{notification_id}::{due_token}
```

A tap acts only if that token still matches the live task. Raw `next_due` rather than a digest: ISO-8601 never contains the `::` separator, the values compare as exact strings with no parsing or timezone normalization, and it reads plainly in a debug log or a hand-written test action string.

| between send and tap | token | outcome |
|---|---|---|
| nothing (still due soon / overdue) | matches | **completes** — the bug |
| completed elsewhere | differs | no-op |
| snoozed / skipped / rescheduled | differs | no-op |
| a second card for the same task was tapped first | differs | no-op |

The last row is the one a notification-scoped dedupe could not do. The token is keyed on the **task**, not on the card it rode in on, so two independent notifications covering one task encode the same token and invalidate each other.

### Simultaneous taps

The check and the mutation it guards are atomic, so two phones on one card — or two cards for one task — cannot both observe the pre-completion `next_due`. Nothing from `get_task()` through `complete_task()`'s `self._tasks[task_id] = updated` yields: awaiting a coroutine runs its body inline until *it* suspends, and `complete_task`'s first suspension is the later `await self._save()`. That invariant is now commented at both ends, since an `await` added above that assignment would quietly open the race.

### Backward compatibility

`decode_action` still accepts the four-field form, so buttons already delivered to phones keep working. Those fall back to the old overdue gate, which is what the pre-existing N1 stale-tap test now pins.

Also in here: a `dt_util.utcnow()`/`dt_util.now()` split in the handler, and a debug log on every drop path — the reporter saw nothing in the logs at all.

## Tests

Both new integration tests were confirmed **red before the fix and green after**, by restoring the old `is_overdue` gate and re-running against the container:

```
test_action_completes_a_task_that_is_not_yet_due FAILED
  AssertionError: an early 'Mark done' tap must complete the task (#216)
test_two_notifications_for_one_task_complete_it_once FAILED
  AssertionError: the first card's tap must complete the task
```

- `test_action_completes_a_task_that_is_not_yet_due` — a due-soon profile, a task snoozed two hours out, an early "Mark done" tap; then the same tap re-fired, asserting no second completion and no `next_due` advance.
- `test_two_notifications_for_one_task_complete_it_once` — two cards, both taps fired back to back with no wait, exactly one completion.
- Unit coverage for `due_token`, `is_current_action`, the legacy decode path, and the token on every button.

Verified locally:

| check | result |
|---|---|
| `pytest tests/unit` | 908 passed, 1 skipped |
| integration suite (Docker) | 131 passed |
| e2e (Playwright) | 46 passed |
| `ci/test-mutation-python.sh` | **100.00%** (100/100), threshold 80% |
| `vale CHANGELOG.md` | clean on added lines |

The mutation run surfaced a pre-existing gap it pulled into scope — nothing asserted the **skip** button's `action` key, so a mistyped key there would have shipped a dead button. Killed with an assertion over all four verbs' action strings.

## Release

Bumps to `0.16.0b2` (`manifest.json` + `PANEL_VERSION`) with a matching CHANGELOG section, since `0.16.0b1` is already published and the fix would otherwise have nowhere to ship. Please add the `preview-release` label so the reporter can install it via HACS before merge.

No `frontend/src/` change, so the screenshot gate doesn't apply and the walkthrough tour needs no edit.

## Note on local verification

`mypy` could not be run in the dev container — it is on Python 3.11 and Home Assistant needs 3.14, so `pip install homeassistant` fails to build. The changed files typecheck standalone under `--strict`, and CI's `lint.yml` mypy job has since passed on this head.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.16.0b2]
+
+### Fixed
+
+- **"Mark done" on a notification works before the task is due.** A notification built
+  from a Profile set to **Due soon** (or **All**) shows tasks that aren't overdue yet,
+  and its **Mark done** button did nothing at all when tapped. No completion, nothing in
+  the log, while **Snooze** and **Skip** on the same notification worked. The task then
+  went overdue as if it had been left alone. Tapping the button now completes the task
+  whenever it still reflects the task's current schedule. Taps that no longer do are
+  still ignored, so completing a task somewhere else (or on a second notification
+  covering the same task) can't push it out an extra cycle. (Fixes #216)
+
 ## [0.16.0b1] - 2026-08-21
 
 ### Added
--- a/custom_components/home_keeper/const.py
+++ b/custom_components/home_keeper/const.py
@@ -8,7 +8,7 @@ PLATFORMS = ["todo", "calendar", "button", "sensor", "binary_sensor", "number"]
 # Frontend panel.
 # PANEL_VERSION is the single source of truth that release.yml validates against
 # manifest.json's "version" (mirrors Pawsistant's CARD_VERSION check).
-PANEL_VERSION = "0.16.0b1"
+PANEL_VERSION = "0.16.0b2"
 PANEL_URL_PATH = "home-keeper"  # sidebar route -> /home-keeper
 PANEL_STATIC_URL = "/home_keeper_panel"  # static path that serves the JS bundle
 PANEL_JS_FILENAME = "home-keeper-panel.js"
--- a/custom_components/home_keeper/manifest.json
+++ b/custom_components/home_keeper/manifest.json
@@ -20,5 +20,5 @@
   "requirements": [
     "Babel>=2.12"
   ],
-  "version": "0.16.0b1"
+  "version": "0.16.0b2"
 }
--- a/custom_components/home_keeper/notifications.py
+++ b/custom_components/home_keeper/notifications.py
@@ -5,7 +5,8 @@ in ``profiles.py`` that decides *which* tasks) by ``profile_id`` and adds *how* 
 deliver them — mobile targets, the button set, snooze duration, style (walk/digest),
 and automatic triggers. This module owns only that delivery concern: notification
 normalization, the mobile-app **payload builders**, and the **action-string**
-encode/decode that routes a notification tap back to the right task and notification.
+encode/decode that routes a notification tap back to the right task and notification
+(and tells a fresh tap from a stale one — see :func:`is_current_action`).
 The filter/queue live in ``profiles.py``; HA-aware sending in ``notifier.py``.
 
 See ``docs/PROFILES_REFACTOR_PLAN.md`` / ``docs/ACTIONABLE_NOTIFICATIONS_PLAN.md``.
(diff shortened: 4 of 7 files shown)


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones either. This changes the mobile-notification action-string wire format and the "Mark done" gate, so scrutinise:

**Correctness of the freshness token.** The token is the task's raw `next_due` ISO string, compared for exact string equality. Is that comparison sound across a DST transition, a `hass.config.time_zone` change, or an HA restart that re-serializes stored datetimes? Could `next_due` round-trip through storage with a different-but-equivalent spelling (offset format, microseconds, `+00:00` vs `Z`) and turn a legitimate tap into a silent no-op? That failure mode would look exactly like the bug being fixed, so I'd rather over-check it. Is a normalized/parsed comparison, or a digest of the parsed instant, safer than raw string equality?

**Action-string parsing.** `decode_action` now accepts 4 *or* 5 fields and splits on `::`. Can any real `next_due` value contain `::`, or can a task id / notification id contain it, such that a 5-field string is mis-parsed? Is the empty-token rejection right, or should an empty token be treated as legacy? Are there companion-app constraints I'm missing — action identifier length caps, character restrictions, case normalization on iOS or Android — that a ~120-char string with `+`/`:`/`-` in it could trip?

**The atomicity claim.** I assert that `get_task()` → `is_current_action()` → `complete_task()`'s `self._tasks[task_id] = updated` contains no yield point, so two simultaneous taps can't both read the pre-completion `next_due`. Please verify that against the actual code rather than taking my word for it — particularly whether anything in `_reject_synced_problem` / `_reject_scan_required` / `recurrence.apply_completion` / `_reset_usage_baseline` / `_stamp_part_replacement` can await or otherwise suspend, and whether `hass.async_create_task` scheduling gives the ordering I'm assuming.

**Backward compatibility.** Notifications already on users' phones carry the 4-field form and fall back to the old `is_overdue` gate. Is that fallback reachable and correct? Does anything else in the repo (frontend, fixtures, docs, glue integrations) encode or parse these action strings and now disagree with the new format?

**Behaviour changes I may not have thought through.** Snooze and skip still act on a stale card while complete no longer does — is that inconsistency defensible or a latent bug? Is rejecting a "done" tap after the user snoozed the same card the right call? What about a task whose `next_due` is edited in the panel between send and tap?

**Maintainability / HA best practices.** Is `is_current_action`'s tri-state-ish `tokenless_ok` parameter the right shape, or is there a cleaner factoring? Are the debug logs at the right level given the reporter saw nothing at all in the logs — should a dropped tap be more visible than `debug`? Any HA conventions I'm violating in the event listener or service-call path?

---

PR #224, opened 2026-08-22
fix(todo): retire a completed one-off from the to-do list (Fixes #221)

### LINKED ISSUE #221: Bug: completed one-off tasks never leave the todo.* list’s needs_action status
Environment

	•	Home Keeper version: v0.14.0 (HACS custom repository prestomation/ha-home-keeper)
	•	Home Assistant Core: 2026.8.2

Summary

A task with recurrence_type: "one-off" correctly records its completion in Home Keeper’s own task store (last_completed is set), but the matching item on the integration’s todo.* entity never leaves needs_action. Other recurrence types don’t show this — a floating task completed the same way correctly advances to its next due date.

Steps to reproduce

	1.	Create a one-off task (home_keeper.add_task, recurrence_type: "one-off", no due given).
	2.	Confirm it shows on the linked todo.* entity as needs_action, due today.
	3.	Complete it (home_keeper.complete_task, just task_id).
	4.	Check the todo list again.

Expected: the item is gone from needs_action (or shows status: completed).

Actual: still needs_action, and its due field is now missing entirely.

Control test

A floating task, created and completed the same way: after completion it correctly advances to needs_action with tomorrow’s due date — normal behaviour for a recurring task. This rules out “any completion breaks the sync” — it’s specific to one-off tasks.

What did NOT fix it

	•	todo.update_item (status: completed) doesn’t change the displayed status, and adds an unwanted duplicate completion entry on the underlying task.
	•	Reloading the Home Keeper config entry (homeassistant.reload_config_entry) — rules out simple caching.
	•	The todo.* entity only supports update, not create/delete, so the stuck item can’t be removed via todo.remove_item either.

Working workaround

home_keeper.delete_task on the affected task removes it from the todo list immediately — destructive (loses completion history), not a real fix.

Suspected area

Whatever builds the todo.* entity’s item list from the task store doesn’t seem to handle a one-off task’s completion the way it handles a recurring task advancing — it leaves the stale item behind instead of dropping it or marking it completed.

prestomation:
Fixes #221

## What changed

`todo_items` skipped dormant tasks only for `triggered` and `sensor`:

```python
if rec_type in (REC_TRIGGERED, REC_SENSOR) and not due_iso:
    continue
```

But `recurrence.apply_completion` sets `next_due = None` for `one-off` too — its own comment says "all go dormant (every time surface drops them)". A completed do-once task therefore fell through and was emitted like any other task, and since the entity only ever emits `NEEDS_ACTION`, it came back unchecked on every refresh with `due=None`. Both of the reporter's symptoms (stuck `needs_action`, missing `due`) come from that one missing branch — the date the item showed *was* `next_due`.

Nothing could clear it, either: the entity declares no CREATE/DELETE, so `todo.remove_item` isn't available, and `todo.update_item(status: completed)` just re-completed the task and appended a duplicate entry to the history of work done exactly once. `delete_task` was the only way out, and it takes the completion record with it.

Every other surface already had this right — `calendar.py`, `binary_sensor.py`, `sensor.py`, `transitions.py`, `profiles.py`, and the panel's `_statusBucket`, which files exactly this shape under **Completed**. So did the docs: `README.md` ("drops out of every active surface") and `docs/INTEGRATING.md` ("invisible to the to-do list, the calendar…") both already describe the fixed behaviour. The to-do list was the lone holdout: its dormancy skip was written for triggered tasks and later extended to sensor ones, never to one-off.

**The fix** extends that skip. Deliberately not a bare `not due_iso` — a floating or fixed task always carries a `next_due`, so a missing one there is malformed data that should stay visible rather than silently vanish. The rule keys on dormancy, not on "has been completed", so a recurring task that just advanced still lists, and a *skipped* one-off (dormant, no `last_completed`) now leaves too.

**A completion arriving for an already-completed one-off is now ignored**, which is what stops the duplicate entry. It's defence in depth rather than the main fix — once the item is off the list, HA's own `_find_by_uid_or_summary` already rejects the reporter's call — but it still guards direct entity calls, intents, and a capture-then-complete race. The predicate stays local to `todo.py`: `recurrence.one_off_expired` answers a different question ("purgeable yet?") and takes a retention window to do it, and widening the mutmut allowlist for a two-line bug fix isn't worth it. `store.complete_task` is untouched, so the device button, the service, tag scans and notification actions keep working on a dormant one-off (e.g. re-completing after an undo).

### Tests

Red before, green after — verified by reverting the skip and re-running both tiers.

- **`tests/unit/test_todo.py`** (new — `todo.py` had no unit coverage at all). Loads the module under the synthetic `hk` package like `test_calendar.py` does, and pins the projection: completed and skipped one-offs absent, armed one-off present with its due date, a just-completed recurring task still listed, an undated floating task still listed, dormant/armed triggered and sensor regressions, plus the completion-routing and rename paths.
- **`tests/integration/test_lifecycle.py`** — the seeded completed one-off ("Renew car registration") is not on the list while the armed one ("Renew passport") is, and a full create → complete → gone flow that also asserts `todo.update_item` can no longer reach the item and the history stayed at exactly one completion. Asserts on `todo.get_items` rather than the entity's state count.

`test_calendar.py`'s HA-stub installer became gap-filling, matching the contract `test_coordinator_purge.py` and `test_device_heal.py` already document (and claim it follows). It early-returned when *any* `homeassistant` module existed, so whichever suite stubbed first won — `pytest tests/unit/test_device_heal.py tests/unit/test_calendar.py` already failed to collect on `main`, and a third such file made that likelier. A full `pytest tests/unit` run was never affected.

### Worth knowing

- **Buy reminders are `one-off` too.** `reconcile_buy_tasks` deliberately keeps a *completed* reminder in the store while the part is still low ("occupies the episode"). Those also lingered undated on the to-do list and now disappear once checked off — the intended level-triggered behaviour.
- **A skipped one-off** now leaves the to-do list, while the panel's `_statusBucket` still files it under the generic no-schedule bucket. That seems right (the panel is the admin surface where nothing silently disappears), but flagging it so it isn't read as a new bug.
- The committed integration fixture carries five leaked `E2E one-off …` records (pre-existing; `TEST_CREATED_MARKERS` has no `e2e` marker). Out of scope here, but they're why the seeded todo count drops by more than one.
- **`mypy` was not run locally**: this container is Python 3.11, below HA's floor, so `pip install homeassistant` backtracks to an ancient release that fails to build (the #199 trap). `lint.yml` covers it. `ruff check` / `ruff format --check` pass locally.

## Checklist

- [x] Tests run locally — `pytest tests/unit -v` (1068 passed) and the full Docker integration suite (132 passed). No frontend change, so `ci/test-frontend.sh` is unaffected.
- [x] `CHANGELOG.md` updated with `(Fixes #221)` under a new `## [0.16.0b3]` section
- [ ] Panel UI changed → n/a, backend-only (`todo.py`)
- [ ] New user-facing feature → n/a (bug fix); version still bumped to `0.16.0b3` in `manifest.json` + `const.py` so the fix reaches beta testers via HACS, `preview-release` not applied

Also updates `.amazonq/rules/architecture-and-code.md`: the rule said a dormant task falls out of every time surface "for free", which is the assumption that caused this bug. Surfaces that *compare* `next_due` get it free; a surface that **lists** tasks needs an explicit skip.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EqEpXgnmp7qh62kGdJsAAD)_

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -105,9 +105,13 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
   schedule — owned by another integration), and `sensor` (due-state derived from a
   bound numeric entity — see below). `one-off`, `triggered`, and `sensor` all carry
   **no cadence fields** (`normalize_fields` returns early for each); a completed
-  one-off and a dormant triggered/sensor task all have `next_due=None`, so they fall
-  out of every time surface (to-do/calendar/sensors/transitions) for free. Keep the
-  recurrence math in pure `recurrence.py` with an explicit `now`.
+  one-off and a dormant triggered/sensor task all have `next_due=None`, and every
+  time surface (to-do/calendar/sensors/transitions) must drop them. Only the surfaces
+  that *compare* `next_due` get that free (`is_overdue` is `False` for `None`);
+  a surface that **lists** tasks needs an explicit dormancy skip, so add one when you
+  build a new one — the to-do list shipped without one for `one-off`, and a finished
+  do-once task sat there undated forever (#221). Keep the recurrence math in pure
+  `recurrence.py` with an explicit `now`.
 - **Sensor-based tasks** (`REC_SENSOR`) bind a task to a numeric entity via a
   `task["sensor"]` block (`models.normalize_sensor`: `entity_id`, `mode`, and the
   mode's fields — `target` for `usage`, `comparison`+`value`+optional `for_seconds`
@@ -189,7 +193,8 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
 - Per-task device-page entities (`button`/`sensor`/`binary_sensor`) are created
   only for **enabled, device-attached** tasks (use
   `coordinator.device_attached_task_ids()`); `todo`/`calendar` likewise skip
-  disabled tasks.
+  disabled tasks, and both also skip **dormant** ones (`next_due is None` on a
+  `one-off`/`triggered`/`sensor`).
 - Attach to an existing device by reusing that device's `DeviceInfo`
   `identifiers`/`connections` (Battery-Notes-style identifier merge) — never
   create a duplicate device for someone else's hardware.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.16.0b3]
+
+### Fixed
+
+- **A do-once task you already finished no longer sits on the to-do list forever.**
+  Checking off a one-off task (renew the passport, register the car) retires it: it
+  goes dormant and leaves the calendar, the due-date sensors and the panel's active
+  list, landing in the panel's **Completed** section with its record intact. The native
+  `todo.home_keeper_tasks` list kept showing it anyway, unchecked and with its due date
+  gone, and nothing could clear it short of deleting the task and its history. A
+  finished one-off now drops off the to-do list as soon as it's done, and comes back at
+  its due date if you undo the completion. Checking one off a second time (through
+  `todo.update_item`, say) no longer records a second completion for work done once.
+  (Fixes #221)
+
 ## [0.16.0b2]
 
 ### Fixed
--- a/custom_components/home_keeper/const.py
+++ b/custom_components/home_keeper/const.py
@@ -8,7 +8,7 @@ PLATFORMS = ["todo", "calendar", "button", "sensor", "binary_sensor", "number"]
 # Frontend panel.
 # PANEL_VERSION is the single source of truth that release.yml validates against
 # manifest.json's "version" (mirrors Pawsistant's CARD_VERSION check).
-PANEL_VERSION = "0.16.0b2"
+PANEL_VERSION = "0.16.0b3"
 PANEL_URL_PATH = "home-keeper"  # sidebar route -> /home-keeper
 PANEL_STATIC_URL = "/home_keeper_panel"  # static path that serves the JS bundle
 PANEL_JS_FILENAME = "home-keeper-panel.js"
--- a/custom_components/home_keeper/manifest.json
+++ b/custom_components/home_keeper/manifest.json
@@ -20,5 +20,5 @@
   "requirements": [
     "Babel>=2.12"
   ],
-  "version": "0.16.0b2"
+  "version": "0.16.0b3"
 }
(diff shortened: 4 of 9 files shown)


prestomation:

This extends `todo.py`'s dormancy skip to `one-off` and adds a guard against duplicate completions. Topics I want scrutinized:

**Correctness**
- Is `rec_type in (REC_ONE_OFF, REC_TRIGGERED, REC_SENSOR) and not due_iso` the right predicate? Can a `one-off` legitimately be listable with no `next_due` — mid-migration, contributed by another integration via `add_task`, or a buy reminder created by `reconcile_buy_tasks`? Is there a path where a user loses sight of a task they still need to do?
- `_completed_one_off` requires `last_completed`, so a *skipped* one-off still routes into `complete_task`. Is that the right split, or should the guard key on dormancy alone?
- The `get_task` lookup moved above the COMPLETED/NEEDS_ACTION branch. Confirm the COMPLETED path's `task is None` behaviour is unchanged (it must still fall through to `complete_task` and raise).
- Undo/redo: after `delete_completion` re-arms a one-off, does it come back on the list and complete cleanly? Any ordering hazard between the guard and `async_settle_buy_tasks`?

**Maintainability**
- `_completed_one_off` duplicates a shape that also lives in `recurrence.one_off_expired` and the panel's `_statusBucket`. Three copies — is keeping it local right, or should it be one shared predicate despite the mutmut allowlist cost?

**Performance / HA best practices**
- `todo_items` is a hot property. Any concern with the added tuple membership test per task?
- Is dropping a completed item entirely the right HA to-do semantic, versus emitting `TodoItemStatus.COMPLETED`? Deliberate choice (the entity has no DELETE support, so completed items would accumulate unclearable), but push back if HA convention disagrees.

**Tests**
- `tests/unit/test_todo.py` is new and stubs a chunk of HA. Are the stubs faithful enough that a real-HA behaviour change would still be caught, or is anything tautological?

---

prestomation:
Thanks — the test gap you flagged was real, and the coverage report landed on the same spot. Fixed in `bc3dd7e`:

- **`test_completing_an_unknown_uid_still_reaches_the_store`** — pins that `task is None` on the COMPLETED path still falls through to `store.complete_task` and raises `KeyError`. Verified it's not tautological: swapping the guard to `task is None or _completed_one_off(task)` makes exactly this test fail and nothing else.
- **`FakeStore.complete_task`** now mirrors the real store's ordering — `KeyError` for an unknown id *before* recording anything — so "nothing was recorded" is a meaningful assertion rather than an artifact of the fake.
- **`test_item_without_a_uid_is_ignored`** — covers the `if not item.uid: return` early exit, the one new statement the coverage report showed uncovered (`todo.py:124`).

Taking your "if you add a 4th callsite, extract it" note as the standing rule for `_completed_one_off`.

/q review Second pass on `bc3dd7e` (test-only). Specifically: is `FakeStore` now faithful enough to `store.complete_task`'s contract that these tests would catch a real regression, or does making the fake raise `KeyError` itself just encode the assumption under test? Anything else in the fake that diverges from the real store in a way that could make a passing test misleading.

---

PR #225, opened 2026-08-22
test(e2e): assert the native surfaces, not just the panel

### LINKED ISSUE #221: Bug: completed one-off tasks never leave the todo.* list’s needs_action status

prestomation:
Follow-up to #224 (which fixed #221). No `Fixes` link — this is test and convention work, not a user-facing change.

## What changed

#221 shipped past a **passing** e2e suite. `one-off.spec.ts` already ran the reporter's exact flow — create a one-off, complete it — and asserted the panel filed it under Completed. The panel was never wrong. The to-do entity was, and nothing looked at it: the only assertion on the native cards was `smoke.spec.ts`'s "they render". The bug was even visible in a committed screenshot (`docs/images/4-usage-todo-and-calendar.png` shows stale to-do items sitting beside panel columns marking those same tasks Completed) for months — capturing a surface is not covering it.

Three gaps, addressed in turn.

### The native surfaces had no content assertions

`helpers.ts` grows `expectAbsentFromActiveSurfaces` / `expectOnTodoList`, and the completion flow in `one-off.spec.ts` now asserts the task is on the to-do list **before** the completion and gone from the to-do list, the calendar and the panel's active list after — plus that the entity's `needs_action` count dropped by exactly one, not zero. Asserting only the post-state would also pass for a task that was never listed at all, which is why both halves are there. A static counterpart asserts the same thing about the *seeded* completed one-off, so it holds without depending on the create flow.

The count assertion is polled deliberately: the entity's state trails its own item list by up to a coordinator tick (~10s observed here — items go first, the count catches up). Worth knowing, since a single read looks like a real failure.

### Only one cell of the dormancy matrix was tested

`dormancy.spec.ts` (new) walks `one-off`, `triggered` and `sensor` (all dormant) against `floating` (the control that re-arms instead), asserting every surface agrees. #221 was the `one-off` cell; the others were untested by construction.

`todo-card.spec.ts` (new) covers the card itself. `async_update_todo_item` had **no** browser coverage at all — checking an item off, renaming one, and the duplicate-completion guard were only ever exercised indirectly.

### The suite leaked its fixtures

Specs named tasks `E2E one-off ${Date.now()}` and never deleted them, so **eight leaked records reached the committed store** — four of them live instances of #221, checked into git. Now: `global-setup.ts` persists the access token it already mints, `helpers.ts` gets `createTask`/`deleteTask`, every spec that creates tears down in `afterEach`, and names are stable (a `Date.now()` suffix made each leak look like a new record rather than the same spec failing to clean up). The eight records are purged — a pure 204-line deletion, zero insertions — and `e2e` joins `TEST_CREATED_MARKERS` so the guard catches the next one however it's named.

## Verification

**Red-before-green, the whole point of this PR:** reverting the `todo.py` fix from #224 turns exactly **7** tests red — the seeded-fixture check, the create/complete flow, the dormancy `one-off` cell, both to-do-card one-off tests, and the new dashboard-contents test — while the `floating` control, the rename test and the recurring-completion test stay green. Before this PR, reverting that fix broke **nothing** in e2e.

With the fix restored:

- **57** e2e passed (was 46)
- **132** integration passed
- **1070** unit passed
- `ruff check` / `ruff format --check` clean
- The fixture came back from a full e2e run with **zero leftovers** — the teardown works.

## Checklist

- [x] Tests run locally — full e2e, integration and unit suites, plus the revert check above
- [ ] `CHANGELOG.md` — n/a, developer-only (tests, rules, AGENTS.md, fixture); `custom_components/` is untouched
- [ ] Panel UI changed → n/a
- [ ] New user-facing UI surface → n/a
- [ ] New user-facing feature → n/a, so no beta bump or `preview-release` label

Conventions are recorded in `.amazonq/rules/testing-and-workflow.md` (a panel assertion is not coverage for a native entity; assert disappearance, not just appearance; a screenshot is documentation, not verification; an e2e spec owns what it creates) and in AGENTS.md's screenshot gate.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EqEpXgnmp7qh62kGdJsAAD)_


prestomation:
/q review Please be critical and skeptical — most serious issues first, and don't withhold minor ones. This is test-infrastructure work, so the failure mode I most want caught is a test that looks like coverage but can't actually fail.

**Test validity (the big one)**
- Are the new assertions in `dormancy.spec.ts` and `todo-card.spec.ts` genuinely falsifiable, or do any of them pass for the wrong reason? I verified 7 tests go red when the `todo.py` fix is reverted, but that only exercises the `one-off` path — the `triggered`, `sensor` and `floating` cells were never red, so please scrutinise whether they'd catch a regression in *their* cell.
- `expectAbsentFromActiveSurfaces` asserts absence via `not.toContainText` on a card locator. Is that sound, or could it pass because the card hasn't rendered its items yet? I poll `todoSummaries()` (the entity) first to sequence it, but I'd like that challenged.
- `todo-card.spec.ts` drives `todo.update_item` rather than clicking the card's shadow-DOM checkbox. Is that a meaningful weakening of "browser coverage", or the right call?

**Correctness / flakiness**
- The `todoCount()` assertion polls because the entity's state trails its item list by a coordinator tick. Is polling the right fix, or is that lag itself worth reporting as a bug?
- Any ordering or shared-state hazards now that specs seed data over REST against a container running with `workers: 1` and a persistent store? The `afterEach` teardown uses `Promise.all` — is that safe against a mid-test failure?
- `helpers.ts` caches the token in module scope and reads `.auth/token.json` lazily. Any failure mode if global-setup is skipped or the file is stale?

**Fixture change**
- The purge of 8 leaked records is a pure deletion. Anything in the seed those records were load-bearing for (counts, ordering, screenshot composition)?

---

prestomation:
The `e2e` job failed on the first push — and it was my own new assertion, which is a fitting way for a PR about untrustworthy tests to start. Fixed in `38ebf8d`.

**Root cause.** `todoCount()` polled the entity's *state* (the needs_action count) and timed out at 30s expecting 10, getting 11. The entity has two views on different clocks: `todo_items` is computed on demand and reflects a completion immediately, while the state string is only written when the coordinator refreshes — and `SCAN_INTERVAL` is **5 minutes**, plus whatever debounced refresh a mutation happens to trigger. Locally a refresh landed ~10s after the completion, so 30s looked generous. Under CI load it didn't arrive. No timeout makes that assertion safe, because nothing bounds the next refresh.

So the assertion now counts `todoSummaries().length` — same intent (exactly one item left, not zero and not several), read from the same source the card renders from, deterministic. `todoCount()` is deleted rather than left unused, and the reasoning is documented on `todoSummaries()` so the next person doesn't reach for the state again.

**Answering your review directly:**

- **Sensor cell** — you were right that it only tested "born dormant". It now also completes the dormant task and asserts it stays dormant with its completion recorded, which exercises `apply_completion`'s sensor branch end-to-end. The armed → completed → dormant half genuinely isn't reachable from e2e: arming needs the watcher to observe the bound entity meeting its condition, and `trigger_task` is triggered-only. That half stays pinned in `test_recurrence_sensor.py`, and the test comment says so rather than implying coverage it doesn't have.
- **`todoCount()` lag "not a bug worth reporting"** — I'd gently push back on the framing, though I agree it isn't this PR's business. A 5-minute `SCAN_INTERVAL` means a user sitting on a dashboard may not see a completed task leave the to-do card until the next tick, unless the mutation's own `async_request_refresh` lands. That's a UX question worth a separate look, not a test artifact. Not changing it here.
- **`not.toContainText` timing** — fair caution. Each check runs after a fresh navigation (the card fetches on load) and after the entity poll has settled, so the stale-frontend window you describe is closed by construction rather than by timeout. Stress-tested below.
- **`Promise.all` teardown, token caching, fixture purge** — agreed on all three, no changes made.

**Verification:** full e2e 57 passed; the three timing-sensitive specs 42 passed under `--repeat-each=3 --retries=0` (retries off deliberately — retries would mask exactly this class of failure); unit 1070 passed; ruff clean; the fixture came back from every run with zero leftovers.

/q review Second pass on `38ebf8d`. Two things specifically: (1) is counting `todoSummaries().length` actually race-free, or have I just moved the race somewhere less visible — is there a path where the item list is settled for the *completed* task but still in flux for another? (2) the sensor cell now asserts a completion on an already-dormant task; does that assertion have any teeth, or would it pass regardless of what `apply_completion` does to a sensor task?

---

PR #226, opened 2026-08-22
feat(panel): add Shopping filter for buy tasks

### LINKED ISSUE #220: A few feature requests and questions

prestomation:
Fixes #220

## What changed

**Shopping filter (issue #220 item 1).** Adds a "Shopping" option to the panel's segmented task filter bar and the dashboard card's `filter` config, so users can isolate auto-created buy tasks at a glance instead of inspecting `source.buy` through the service API.

- `types.ts` — added `buy?` to the `Task.source` type (the backend already sends it; the frontend type just didn't declare it).
- `panel.ts` — extended `TaskFilter` with `'shopping'`, added the button in `_controls()`, added filter logic in `_tasksList()`.
- `card-filter.ts` — extended `CardFilter` with `'shopping'`, added `case 'shopping'` in `matchesFilter()`, exempted the shopping filter from `horizon_days` clipping (buy tasks may or may not be dated).
- `card-filter.test.js` — two new tests covering the shopping filter and its horizon bypass.
- All 16 locale files — added `filter.shopping` with translated labels.

**Document clear_on_recover stock behavior (issue #220 item 5).** The user asked whether a `clear_on_recover` auto-completion consumes stock from a linked consumable. It does — the code path is unified — but this was documented nowhere. Added cross-references in three places:

- `README.md` — "Clear when back to normal" paragraph now notes the stock draw-down; the stock paragraph now lists all completion paths.
- `docs/EVENTS.md` — the `clear_on_recover` paragraph now mentions the stock side-effect and possible events.

**Walkthrough extended** — added a step to `walkthrough.capture.ts` demonstrating the Shopping filter pill (click Shopping, pause, click back to All) so the CI-generated video tour shows the new surface.

**Version bump to 0.16.0b4** with a CHANGELOG entry for the new feature.

### Panel screenshot — Shopping filter active (empty state, no buy tasks in seed data)

Panel showing the Shopping filter selected with no matching tasks

### Panel screenshot — task list with the new filter bar (All selected)

Panel task list showing All, Overdue, Due soon, Shopping filter buttons

## Checklist

- [x] Mutation testing — 100% score on `card-filter.ts` (19/19 mutants killed)
      step through it

---
_Generated by [Claude Code](https://claude.ai/code/session_014yxDVEDH4QKyFq8i1jjGRM)_


prestomation:
/q review

Please give a critical, skeptical review of this PR. Focus on:

1. **Correctness** — Does the new `shopping` filter correctly select only `source.buy` tasks in both the panel (`panel.ts`) and the dashboard card (`card-filter.ts`)? Are there edge cases where `source` is null/undefined that could throw? Is the `horizon_days` bypass for shopping correct — should buy tasks always ignore the horizon guard?

2. **Completeness** — Does the `CardFilter` type union, the `matchesFilter` switch, and the panel's filter branch all agree? Are there any code paths that handle `TaskFilter`/`CardFilter` values exhaustively that we missed updating?

3. **Test coverage** — The two new `card-filter.test.js` tests cover the happy path and the horizon bypass. Are there missing edge cases (e.g., `source: null`, `source: {}`, `source.buy` being falsy but present)?

4. **i18n** — All 16 locale files were updated with `filter.shopping`. Are the translations reasonable? Any key-ordering issues that would break the parity test?

5. **Documentation** — The `clear_on_recover` stock-consumption docs added to README.md and EVENTS.md: are they accurate and placed in the right context? Do the cross-reference anchors resolve correctly?

6. **HA best practices** — Is the screenshot capture step in `screenshots.capture.ts` following the existing pattern correctly?

Surface the most serious issues first. Don't withhold minor ones.

---

PR #227, opened 2026-08-22
fix(card): register the dashboard card as a Lovelace resource

jableader:
Fixes: no existing issue created — found during usage. Firefox gives the most consistent
repro (also reproduced in Chrome and Safari).

## What changed

**Problem.** On a cold page load (empty cache), every `custom:home-keeper-card` renders
Home Assistant's **"Configuration error"** card and never recovers.

**Root cause.** The card is injected with `frontend.add_extra_js_url`. HA's index template
`import()`s those modules in the **same tick** as its own `app.ts`, which loads
`@webcomponents/scoped-custom-element-registry` and then **replaces** `window.customElements`.
The small card bundle wins that race and calls `customElements.define` on the *native*
registry HA immediately discards, so Lovelace — querying the replacement — never sees the
card. HACS-installed cards avoid this because they load as Lovelace **resources**, imported
after boot once the swap is done.

**Fix.** Load the card as a Lovelace **module resource** instead of an early
`add_extra_js_url` module:

- `card.py`: register/update one storage-mode module resource (idempotent, cache-busted
  `?v=` token); remove it on last-entry unload; add `lovelace` as an `after_dependency`.
- `card-index.ts`: drop the registry-swap/`loadCardHelpers` timing workaround — a plain
  guarded `customElements.define` is now enough.
- YAML-mode dashboards don't read the storage collection, so the e2e dashboard declares the
  module explicitly and `openCardDashboard` drops its reload-retry loop.

**Why a refactor, not a timing patch.** `add_extra_js_url` fundamentally loads pre-swap, so
any in-bundle fix is timing guesswork (`window.load` is unreliable; the registry is swapped
via `defineProperty`, so it can't be hooked). Loading as a resource removes the race at the
source and matches every mature card (Mushroom, button-card, vacuum-card) — no timing code to
maintain. **Breaking (YAML-mode dashboards only):** storage resources don't auto-apply to
global YAML-mode Lovelace or per-dashboard YAML dashboards; there the card must be added as
a `resources:` entry (the integration logs the exact URL on startup). The default storage
dashboard is unaffected. Ships as `0.16.0b4`.

**Testing.** Cold-load repro (fresh browser context per load) renders all 3 cards with **0**
error cards on every run — previously every card errored. Adds unit, integration and frontend
guard tests (create/update/idempotency/YAML-mode/cleanup; fails if the timing hacks return).
No visual change to the card.

## Checklist

- [-] ~Panel UI changed → current screenshots committed under `docs/images/` and
      embedded above with an HTML `` tag~ — N/A: loading-mechanism change, no visual change
- [-] ~New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended to
      step through it~ — N/A: no new surface
- [x] Version bumped to the next beta (`manifest.json` + `const.py` → `0.16.0b4`) — not a
      new feature, but a breaking change for YAML-mode dashboards warrants the bump.


prestomation:
Thanks! My apologies but I didn't actually see this when I implemented #232. I appreciate the contribution

PR #229, opened 2026-08-22
feat(inventory): mirror buy reminders onto an existing to-do list

### LINKED ISSUE #220: A few feature requests and questions

prestomation:
Fixes #220

## What changed

Item **2** of #220: *"Let reorder tasks show up on a normal HA list too."* The reporter
confirmed in the thread that one Home-Keeper-wide setting pointing at a single list is
what they want, so that's what this is.

A new integration-wide option, **`shopping_list_entity`** (`""` = off), puts every
auto-created *"Buy {part}"* reminder on a to-do list the household already uses — the
built-in shopping list, a `local_todo` list, whatever their voice assistant reads out.

The mirror is **two-way** and level-triggered:

| What happens | What the mirror does |
|---|---|
| Part goes low | The reminder is added as an ordinary line |
| Line ticked off **on the external list** | Completes the reminder with `origin=home_keeper_shopping_list`, which restocks the part and retires it. The ticked-off line is left alone as the shopper's record |
| Reminder completed **in Home Keeper** | The line is ticked off to match |
| Reminder ends **unbought** (stock topped up by hand, auto-buy off, part deleted) | The line is removed — nothing was bought |
| Target list changed, or the mirror switched off | Lines are cleared off the old list |

An already-completed item is never touched, and only lines Home Keeper added are ever
managed — the milk you put on the list yourself is none of its business.

The Settings tab's new Shopping list card, with a to-do list picked

A Buy Anode rod line on the household shopping list card

The full Settings tab, with the new Shopping list card between General and Problem sensor sync

### Shape

Same split as the problem-sensor sync: a **pure planner** (`shopping.py`, added to
`only_mutate`) and an **HA-aware driver** (`shopping_sync.py`) that reads the target over
`todo.get_items` and applies with `todo.add_item` / `update_item` / `remove_item`.

Four properties the design turns on, all covered by tests:

- **A failed call is retried, never compensated.** `plan_sync` returns the bookkeeping as
  it reads *once every op succeeded*, and each op carries its `key` so the driver puts an
  entry back when a call raises. Dropping an entry for a remove that never happened would
  orphan that line on the list forever with nothing left to remember it by.
- **An unreadable list is not an empty one.** A list absent from `items_by_entity`
  (unavailable, service raised) has nothing planned for it and keeps its tracking —
  otherwise a shopping list that briefly went away reads as "the user emptied it".
- **`needs_pass` gates the read.** The settle chokepoint fires on every completion and
  stock nudge, and almost none concern a mirrored part; a pure check over the two maps
  decides whether any `todo` service is called at all. The surfaces watching the
  *shopper's* side (the target's state listener, a periodic sweep) force a full pass,
  since that side is invisible from Home Keeper's own state.
- **It can never break its caller.** Every outbound call is best-effort (the `notifier.py`
  precedent) and `async_sync` swallows what reaches it — a pass runs off the back of a
  completion or a stock adjustment.

Bookkeeping (which line on which list stands for which reminder) persists in the storage
doc as `shopping_items` — additive, no `STORAGE_VERSION` bump, alongside `problem_notes`.
It has to be durable: the moment it matters most is *after* the reminder is deleted, and
an options save reloads the entry.

`async_settle_buy_tasks` now runs the mirror on **both sides** of the reconcile and before
the deferred reload is scheduled — before, so a just-completed reminder still exists to
tick its line off; after, for whatever the reconcile created or retired.

### Also in here

- **Fix:** completing a buy reminder from a **notification tap** called
  `async_request_refresh` instead of `async_settle_buy_tasks`, so it restocked the part
  but left the reminder standing until something else settled it. Pre-existing; found
  while wiring the chokepoint. In the CHANGELOG under Fixed.
- `recurrence.one_off_completed` hoisted out of `todo._completed_one_off` rather than
  writing a third copy of that predicate.
- `also_copy = ["ci", "CHANGELOG.md"]` in `[tool.mutmut]`. `test_release_issues.py` reads
  `CHANGELOG.md` at *import* time, so the mutmut sandbox died at collection — exactly the
  trap the neighbouring comment already documents, and it went unnoticed because no PR
  since has changed mutable Python.
- The e2e dashboard gains a second to-do card for `todo.shopping_list`, so a browser test
  can see what the shopper sees. Both to-do cards are now titled and `todoCard()` is
  scoped by title — masonry reflow made `.first()` ambiguous.

### Two things worth a maintainer's eye

1. **`(Fixes #220)` in the CHANGELOG.** #220 bundles five requests; items 3 and 4 are
   still open, so `notify-issues` would close it on the stable cut. `0.16.0b4` already set
   this precedent for item 1, so I've followed it — but you may want to drop the keyword
   until the rest lands.
2. **Pre-existing, not touched here:** the options *flow* writes `user_input` raw via
   `async_create_entry`, which replaces the whole options dict — so opening and saving the
   flow wipes `profiles`, `notifications` and `dismissed_companions` (they're deliberately
   not in the flow's schema). Worth a separate fix.

## Local verification

| Gate | Result |
|---|---|
| `pytest tests/unit` | 1146 passed, 1 skipped |
| `npx vitest run` | 565 passed |
| `tsc --noEmit` (panel) | clean |
| `ci/test-mutation-python.sh` | **97.61%** (368/377), threshold 80% |
| `ci/test-mutation-frontend.sh` | **100.00%** (12/12), threshold 80% |
| `tests/integration` (Docker HA) | 137 passed |
| `tests/e2e` (Playwright) | 59 passed |
| `vale` on added prose lines | 0 hits |

The two headline e2e tests were verified to *fail* with the mirror disabled, and the
driver's failure-path unit tests were verified to fail with their guards removed.

`mypy` was not run locally: this container is on Python 3.11, below Home Assistant's own
floor, so `pip install homeassistant` would silently resolve a months-old version and
check an API nobody runs (#199). Leaving that to CI.

## Checklist

      embedded above with an HTML `&lt;img&gt;` tag
      step through it


prestomation:
/q review Please be critical and skeptical — surface the most serious problems first, and don't withhold the minor ones. This adds a two-way mirror of auto-buy reminders onto an arbitrary external `todo` entity, so the risky surface is a reconcile loop that writes to someone else's integration and reads back from it.

**Correctness — dig hardest here.**
- `shopping.plan_sync` is the whole state machine (tracked entry × live item status × reminder open/completed/gone × target changed). Look for a combination that loses an item, strands one, or creates a duplicate. Especially: the *adoption* path (an untracked reminder claiming an existing `needs_action` line) — can it ever claim last episode's completed line and immediately "complete" the fresh reminder, i.e. an endless restock loop?
- The retry contract: ops carry a `key` so `shopping_sync._apply` restores the entry from `before` when a call raises. Is there a failure that leaves `settled` inconsistent, or an op whose failure should restore but doesn't?
- Reentrancy: `async_sync` is guarded by `_running`/`_pending` with `_MAX_PASSES`, and the reverse direction calls `coordinator.async_settle_buy_tasks()`, which calls back into `async_sync`. Convince yourself the recursion terminates in every ordering, including our own writes echoing back through the target's state listener.
- `coordinator.async_settle_buy_tasks` now awaits two mirror passes *before* scheduling the deferred entry reload. Does that widen the window where a service call 500s with "No active coordinator", or risk a torn-down store mid-pass? `_stopped` is checked after each await — is that sufficient?
- `needs_pass` skipping the read: is there a Home-Keeper-side drift it can't see that would leave the mirror permanently stale until the 5-minute sweep?

**HA best practices.** `todo.get_items` with `blocking=True, return_response=True` from a coordinator path; the `supported_features` bitmask check; `async_at_started` for the initial pass; catching bare `Exception` around foreign service calls. Anything non-idiomatic or likely to break on a future HA?

**Performance.** A pass runs off every completion and stock adjustment. Is the `needs_pass` gate enough, or does this add real cost to a large store? The periodic sweep fires per coordinator tick when anything is mirrored.

**Security / data safety.** The mirror deletes and modifies items on a list the user owns. Can it ever touch a line Home Keeper didn't add, or one the user already ticked off? `own_todo_entity_ids` refuses our own list — is the registry lookup the right gate?

**Maintainability.** Is the pure/HA split drawn in the right place? Is `shopping_items` in the storage doc the right home for the bookkeeping, or should it be in-memory with summary-based re-adoption?

---

prestomation:
/q review Second pass, please — same posture: critical and skeptical, most serious first, and don't withhold the minor ones. The previous review found nothing, so this time please go looking specifically for the failure modes below rather than assessing the design.

Since then the only change is prose (two `vale` findings in README.md and docs/EVENTS.md), so the code under review is unchanged.

Concentrate on:

1. **`shopping.plan_sync` adoption vs. the previous episode's completed line.** An untracked reminder adopts an existing `needs_action` item by summary. Buy reminders are named deterministically ("Buy {part}"), so last episode's line for the same part can still be on the list. Convince yourself — or show me otherwise — that a *completed* leftover can never be adopted and instantly complete the fresh reminder, which would be an endless restock loop.
2. **Bookkeeping lost or corrupted.** If `.storage/home_keeper`'s `shopping_items` is missing, stale, or points at a list the user has since renamed: does the mirror duplicate lines, delete someone else's line, or wedge?
3. **The `_running`/`_pending`/`_MAX_PASSES` guard.** Trace `state listener → async_sync → complete_task → async_settle_buy_tasks → async_sync` and any interleaving with the coordinator's periodic `async_schedule_sweep`. Can two passes overlap, or can the loop fail to converge and burn `_MAX_PASSES` every time?
4. **Ordering in `coordinator.async_settle_buy_tasks`.** Two awaited mirror passes now run before the deferred reload is scheduled. Does that widen the "No active coordinator" window for concurrent service calls, or let a reload tear the store out from under an in-flight pass? `_stopped` is re-checked after each await — is that actually sufficient, or is there a gap?
5. **Anything in `_apply` where a failed op leaves `settled` inconsistent** with what actually happened on the list.

Also flag anything you'd expect to break on a future Home Assistant: the `todo.get_items` response shape, the `supported_features` bitmask read, or `async_track_state_change_event` on an entity that may not exist yet.

---

prestomation:
Thanks — triaging the three minor findings.

**1. `KeyError` in the `complete_task` catch — keeping it.** It is reachable. `store.complete_task` opens with

```python
existing = self._tasks.get(task_id)
if existing is None:
    raise KeyError(task_id)
```

and the mirror is exactly the caller that can hit it: `plan_sync` reads the task map, then the driver awaits `todo` service calls before completing. Another path through `async_settle_buy_tasks` can retire the reminder in that window, so the id is stale by the time we use it. `notifier.py` catches the same pair on the same call for the same reason. The unit test only exercises `TaskValidationError` because that's the cheaper fake — the `KeyError` arm is the race, not the validation.

**2. Logging level for an unavailable target — keeping DEBUG, but the split deserves a comment.** The two cases aren't the same:

- Target **doesn't exist** (deleted, renamed, never existed) — real misconfiguration, already `_warn_once` at WARNING and named in the log.
- Target exists but reads `unavailable`/`unknown` — its owning integration is temporarily down. That's transient, Home Assistant already logs it, and a Home Keeper warning would fire on restarts where a cloud-backed list comes up after we do. It recovers by itself on the next pass.

So the misconfiguration case is loud and the transient case is quiet, which I think is the right split — but it isn't obvious from the code, so I'll add a comment saying why rather than leave the next reader to re-derive it.

**3. Feature-check timing — agreed, no change.** Your read is right: a list that loses a capability between plan and apply fails the call and the entry is restored from `before`, so the next pass retries. That's the same path as any other failed op.

One correction to the analysis, since it matters for anyone reading this later: the note that "`needs_pass` returns false for settled state ... and force=False is used for the first settle call" is right about the guard but not about why the recursion terminates. The reverse-direction re-entry is stopped by `_running`, not by `needs_pass` — a nested `async_sync` sets `_pending` and returns without doing anything at all. `needs_pass` is the cost gate (does this pass touch a `todo` service), `_running`/`_MAX_PASSES` is the termination guarantee. They're independent.

---

prestomation:
Thanks — one of these is a real improvement and is now pushed (`8458f48`). Three of the concerns don't hold up against the code, and I'd rather say why than leave them looking open.

**Taken: log when the pass budget is exhausted.** Right, and the reasoning is right. `_MAX_PASSES` exists to guarantee termination, not because ending that way is acceptable — a converging mirror settles in two. Exiting quietly left the list stale with nothing in the log. It now warns, with a test that fails if the warning is downgraded.

Writing that test turned up something worth pinning, so it's in the same commit: I could not construct a non-converging case through the real paths. Every one I tried settles, including the one that looks like it shouldn't — a reminder still open after a tick-off, because its restock quantity didn't clear the threshold. The mirror puts a *fresh* line on the list rather than adopting the ticked-off one, then stops. That's now `test_a_reminder_still_open_after_a_tick_off_gets_a_fresh_line`, and the budget test drives `_sync_once` directly since no real path reaches it.

**Not taken — `_pending` left set when a sweep runs after `_async_stop`.** It can't be. `async_sync`'s `finally` clears both flags on every exit path, including the `_stopped` early return:

```python
finally:
    self._running = False
    self._pending = False
```

**Not taken — two concurrent settles both scheduling a reload.** The check and the set have no `await` between them:

```python
if not self._buy_reload_scheduled:
    self._buy_reload_scheduled = True
    self.hass.async_create_task(...)
```

The event loop is cooperative, so no other coroutine can interleave there. What my change *does* widen is the window between `reconcile_buy_tasks()` and that guard, because the second mirror pass now sits in it. A concurrent settle can schedule its reload first — and then we correctly skip, because the reload it scheduled reloads the whole entry and covers our changes too.

**Not taken — validating the target exists before registering the state listener.** This would make it worse, not better. `async_track_state_change_event` tracks by entity id, not by entity object, and a state write for a previously-unknown id is an ordinary `state_changed` with `old_state=None`. So a target created *after* setup is picked up the moment it appears. Gating registration on current existence is what would require a reload.

On `supported_features`: the deprecation is about entity classes exposing the enum, not about reading the state attribute. On a `State` it is still an int, and `attributes.get("supported_features") or 0` masked against the flag is the idiomatic read. Happy to revisit if you have a specific HA version where that changes.

Your point about losing the target list *and* the bookkeeping together is fair and is a real gap, not a bug: nothing recovers the orphaned lines automatically. Adoption covers losing the bookkeeping alone. I've left it as designed rather than adding a reconciliation path for a case that needs storage loss to reach — flagging it here so it's a known limitation rather than an unnoticed one.

---

PR #232, opened 2026-08-22
fix(card): register the dashboard card as a Lovelace resource (Fixes #228)

### LINKED ISSUE #228: Dashboard: Configuration error Custom element doesn't exist: home-keeper-card
I have a home assistant dashboard with a card. When i refresh the browser it says: Configuration error Custom element doesn't exist: home-keeper-card

If i run developer console and disable cache the card will show as expected but as soon as i enable cache and reload the same error appears. 

A lot of cards from hacs are listed in Settings > Dashboards > Resources but not this one, could that be related?

prestomation:
Fixes #228

### The bug

`frontend.add_extra_js_url` reaches a browser exactly one way: HA's `IndexView` renders an inline dynamic `import(...)` per extra module into the **app-shell HTML**. Confirmed against the running container — a bare inline `script` element containing:

```js
import("/home_keeper_panel/home-keeper-card.js?v=97290d922ce4");
```

That response carries no `Cache-Control` and no ETag, and HA's own frontend service worker registers:

```js
registerRoute(/\/.*/, new StaleWhileRevalidate({ cacheName: "file-cache",
  plugins: [new ExpirationPlugin({ maxAgeSeconds: 86400 })] }));
```

So dashboard navigations are served from a **cached app shell**. A shell cached before Home Keeper was installed — the normal state for anyone who used their dashboard first — has no import for the card, and `home-keeper-card` is never defined.

It reads as permanent rather than flaky because HA *already* recovers from a late definition (`_customCreate` hides the error card for 2s and fires `ll-rebuild` on `customElements.whenDefined`). A stuck error card therefore proves the module never executed at all. `frontend/subscribe_extra_js` doesn't save it either: it pushes only *changes* to already-connected clients, never the initial set.

The reporter's own aside is the proof: their HACS cards worked on that same stale shell, because Lovelace **resources** are fetched over the websocket on every dashboard load.

### The fix

Register the bundle as a Lovelace resource as well — which is also the [only method HA's docs describe](https://developers.home-assistant.io/docs/frontend/custom-ui/registering-resources/) for loading a custom card. `add_extra_js_url` stays: it is the only path when Lovelace runs in yaml resource mode, and double delivery is inert because both name the same URL and `card-index.ts` guards its `define` and its `customCards` push.

Reconciliation is a pure planner (`card_resource.py`) so it is unit-tested and mutation-scored. It matches on the URL **path**, so:

- a rebuilt bundle's new `?v=` token **updates** the existing row instead of adding one;
- an install with a hand-added entry (or an absolute URL typed in by hand) is **adopted**, not duplicated;
- extra copies are **collapsed** to one;
- an up-to-date row is **left alone**, so a restart doesn't rewrite `.storage`.

The resource is dropped in `async_remove_entry` only — never on unload, which also runs on every reload.

Users get one row under **Settings → Dashboards → Resources**, which is exactly where the reporter looked and found ours missing.

Lovelace's resource collection classes are internal to that integration, and `card.py` is imported from `__init__.py` — so a runtime import of them would turn a future HA refactor into an ImportError that takes the whole integration down. It branches on `data.resource_mode != MODE_STORAGE` instead (the discriminator HA core itself uses, needing only `lovelace.const`) and keeps the class behind `TYPE_CHECKING`.

## Red → green

All three tiers failed before the fix.

**e2e** — `tests/e2e/tests/card-registration.spec.ts` reproduces #228 by stripping the card's `import(...)` out of the served dashboard HTML, which is what a stale cached shell *is*. Red, it reproduced the report exactly: three "Configuration error" cards where the Home Keeper cards belong, native to-do and calendar cards rendering fine beside them.

Three things keep it honest:

- It asserts the *unstripped* HTML did contain the import, so it can't go vacuous if HA changes how it delivers extra modules.
- It asserts the bundle was actually **fetched** afterwards, so a future shell-side fallback can't define the element with no request and still pass.
- `test.use({ serviceWorkers: 'block' })` — HA registers a service worker on first load, and a service worker answers navigations *before* `page.route` sees them, so the reload that follows was served the original unstripped shell. It passed for the wrong reason until instrumented.

It deliberately does not use `openCardDashboard` — that helper reloads 3× to absorb cold-frontend flake, which here would only re-serve the stripped shell while turning a precise failure into an opaque timeout.

**integration** — `tests/integration/test_card_resource.py`: exactly one `module` resource with a cache-busting token, and an entry reload neither drops nor duplicates it. Red: `expected a single card resource, got []`.

**unit** — `tests/unit/test_card_resource.py`: create, no-op, stale token, wrong type, duplicates, absolute URL, and the `res_type`-vs-`type` asymmetry (Lovelace takes `res_type` on write and renames it to `type` on the way in).

Restart survival was verified by hand: the row keeps its id and count across restarts.

## The CI failure this PR worked through

The first push was green everywhere except `e2e`, which failed three times on this spec alone while passing locally. Cause: Chrome classifies a response synthesized by `route.fulfill` as coming from a *public* address space, so its Local Network Access checks block the page's own websocket:

```
WebSocket connection to 'ws://localhost:8123/api/websocket' failed:
net::ERR_BLOCKED_BY_LOCAL_NETWORK_ACCESS_CHECKS
```

With no websocket the frontend never fetches the resource list, so the card could not load by the very path the test exists to prove — a false failure that looks exactly like a real one.

It reproduced only under the browser CI actually uses. `e2e.yml` runs `npx playwright install chromium` with no override, so CI drives Playwright's **headless shell**; the `CHROMIUM_EXEC` override AGENTS.md documents for the Claude Code remote environment points at an older full Chromium that does not enforce the check. Every local run had used the wrong browser. That trap is now written into AGENTS.md and `.amazonq/rules/testing-and-workflow.md`.

The flag is scoped to this spec's own `test.use({ launchOptions })`, not the project — `playwright.config.ts` is byte-identical to before.

## Two incidental fixes

- **`mutation.yml` was broken on `main`.** mutmut copies only `source_paths` + tests into `mutants/`, so `test_release_issues.py` (added in #223) died reading a `CHANGELOG.md` that isn't there, taking the whole run down at stat collection. #229 hit this independently and its wording won the merge.
- **9 surviving mutants, all the `""` fallback in `item.get("url", "")`.** No assertion can kill those — a row without a URL never matches the bundle path whatever sentinel stands in for it. Removed the sentinel instead of annotating it (`_serves_bundle` tests `isinstance(url, str)`), taking the score 86.76% → **100%**.

## Verification

Everything run locally, on the merged tree, with `CHROMIUM_EXEC` unset so the browser matches CI:

- `pytest tests/unit` — 1164 passed
- `pytest tests/integration` — 141 passed
- Playwright full suite — 60 passed
- `mypy custom_components/home_keeper` — clean, against HA **2026.8.3** on Python **3.14.6** (`ci/check-ha-version.py` caught an earlier attempt silently resolving 2026.2.3, exactly the #199 trap)
- `ruff check` / `ruff format --check` — clean
- `ci/test-mutation-python.sh` — 100.00%, 53/53
- `vale` — no findings on lines added vs `main`

Both directions re-verified after every change: green as shipped, and still red with `_async_sync_resource` stubbed out, reporting `bundle traffic: (the bundle was never requested)`.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`, integration, e2e)
- [x] `CHANGELOG.md` updated with `(Fixes #228)` under a new `## [0.16.0b6]`
- [ ] Panel UI changed → N/A, nothing under `frontend/src/` changes
- [ ] New user-facing UI surface → N/A, bug fix
- [x] Version bumped (`manifest.json` + `const.py`) — `0.16.0b5` shipped with #229 while this was in review, so the fix opens `0.16.0b6` to reach the reporter via HACS

Conventions recorded in `.amazonq/rules/architecture-and-code.md` (resource vs extra-module delivery) and `.amazonq/rules/testing-and-workflow.md` (a second delivery path needs a test that deletes the first; browser parity with CI; the interception/LNA interaction).


prestomation:
/q review Please be critical and skeptical — surface the most serious problems first, and don't withhold minor ones. This changes how the bundled dashboard card is delivered to browsers (issue #228), so scrutinise:

**Correctness of the resource lifecycle (`card.py`, `card_resource.py`)** — Can `_async_sync_resource` ever create a duplicate row? Consider: two config entries, a config-entry reload racing setup, `async_when_setup` firing after the store has already been mutated, and the lazy `ResourceStorageCollection` (I call `async_get_info()` before `async_items()` precisely because the collection is empty until loaded — is that the right public call, and is it enough?). Is `hass.data[_CARD_REGISTERED]` the right guard, and is popping it in `async_unregister_card_resource` correct for a remove-then-re-add in the same HA run?

**The path-matching rule** — `plan_card_resource` treats any resource whose URL *path* equals `/home_keeper_panel/home-keeper-card.js` as ours, adopting and rewriting it. Is adopting a user's hand-added row the right call versus leaving it and adding our own? Is there a URL shape (absolute, proxied, `hassUrl` with a subpath, trailing slash, query-in-path) where `urlsplit().path` gets this wrong?

**HA best practices** — Is writing to `.storage/lovelace_resources` from an integration acceptable, or is there a sanctioned API I'm missing? Is `after_dependencies: ["lovelace"]` + `async_when_setup` right, or should it be a hard dependency? Should removal also happen somewhere other than `async_remove_entry`? Is importing `homeassistant.components.lovelace.resources.ResourceStorageCollection` (an internal) a liability across HA versions, and is the `isinstance` narrowing against `ResourceYAMLCollection` the right way to detect yaml resource mode?

**Failure paths** — Both entry points swallow `Exception` and log. Does that hide anything a user needs to act on? Does the yaml-resource-mode branch degrade sensibly (it keeps the pre-fix behaviour, including the #228 exposure — is that defensible)?

**Test quality** — Does `tests/e2e/tests/card-registration.spec.ts` genuinely reproduce #228, or does blocking service workers make it prove less than it claims? Could it pass for the wrong reason? Are the integration assertions strong enough given registration happens off the setup path?

**Performance / maintainability** — The `?v=` token is a content hash, so a rebuilt bundle rewrites the row and fires `lovelace_updated`. Is that churn acceptable? Any duplication or naming you'd change between `card.py` and `card_resource.py`?

---

prestomation:
/q review This push fixes the e2e failure from the previous run. Please be critical and skeptical about the test-harness change specifically — most serious issues first, minor ones included.

**The `--disable-features=LocalNetworkAccessChecks` launch flag (`tests/e2e/playwright.config.ts`)** — it is applied to the whole chromium project, not just the one spec that needs it. Is that too broad? Could disabling Local Network Access checks mask a real product bug in any of the other 57 specs, or change what they prove? Would scoping it to the single spec via `test.use({ launchOptions })` be better, given that would also have to re-plumb `CHROMIUM_EXEC`? Is there a way to simulate a stale app shell that avoids `route.fulfill` on the document altogether — and would that be a truer or weaker simulation of #228?

**Does the test still test the right thing?** `card-registration.spec.ts` blocks service workers *and* now disables an LNA browser check. That is two browser behaviours suppressed. Is it still a faithful reproduction of what issue #228's reporter experienced, or has it been tuned until it passes? Concretely: if `card.py`'s Lovelace-resource registration regressed, would this test still fail? (I verified by hand that it does, but say if you see a path where it wouldn't.)

**The raised budgets** — 60s for the element, 120s for the test, against a repo-wide `timeout: 60_000`. Is a per-test override that large hiding a latent slowness, or is it justified because the resource path can only start after the websocket connects?

**Diagnostics** — the spec now attaches `console`/`pageerror`/`request`/`requestfailed` listeners and rethrows a composed error. Any leak, ordering, or noise concern? Is `String(err)` inside a rethrow losing the original stack in a way that will hurt the next debugger?

**Everything from my earlier review request still stands** for the `card.py` / `card_resource.py` resource lifecycle — please don't drop it in favour of only reviewing this delta.

---

prestomation:
Triaged both reviews. Five findings taken, two pushed back on, one already covered. Pushed as `879b86e`.

**Taken**

1. **Internal API fragility** — the sharpest finding, and worse than described. `card.py` is imported from `__init__.py`, so an ImportError on `lovelace.resources` would have taken the *whole integration* down, not just the card's second delivery path. Now branches on `data.resource_mode != MODE_STORAGE` (the discriminator HA core itself uses, needing only `lovelace.const`) with `ResourceStorageCollection` behind `TYPE_CHECKING` for the annotation only. The mode doesn't narrow the union for mypy, so the return is an explicit `cast` with a comment naming why the mode guarantees the type.

2. **LNA flag scope** — agreed, and moved into the spec's own `test.use({ launchOptions })`. `playwright.config.ts` is byte-identical to before. Cost is the `CHROMIUM_EXEC` re-plumb you predicted, since `launchOptions` replaces the config's copy rather than merging. Side effect worth knowing: the differing launch options give that file its own worker, so it now runs last.

3. **"The test doesn't verify the resource fetch itself — it infers success from the card appearing."** Correct, and the best point in the second review. It now asserts a *response* for the bundle after the shell import was removed, so a future shell-side fallback that defines the element without a fetch would fail rather than pass.

4. **`String(err)` loses the stack** — fixed, `new Error(msg, { cause: err })`.

5. **The two documentation asks** — the single-entry assumption is now in `card.py`'s docstring (naming `_abort_if_unique_id_configured` as what enforces it), and `resource_path` explains why adopting a hand-registered absolute or proxied URL is deliberate while a copy at a *different* path is left alone.

**Pushed back on**

- **"Is there a way to simulate the stale shell without document-level routing — perhaps by mocking the websocket resource list?"** That would invert the test. The resource list is the *fix*; mocking it would mean asserting the fix works by supplying the fix. The bug is that delivery rode on a cached app shell, so removing the shell's import is the only faithful simulation. Two suppressions is a fair thing to flag, but both are the harness getting out of the way — the service worker would re-serve an unstripped shell, and LNA would kill the websocket the resource path depends on. Neither touches Home Keeper. The check that matters is the one you named: I stubbed out `_async_sync_resource` and confirmed the test still fails, with `bundle traffic: (the bundle was never requested)`.

- **Timeout budgets.** Keeping 60s/120s. They're ceilings, not waits — the test passes in ~1.5s — so they cost nothing when green and only buy headroom on a loaded runner. The earlier 30s failure was the blocked websocket, not slowness, and the diagnostics now distinguish those cases directly, which was your underlying concern.

**Already covered**

- **The duplicate-row race.** Agreed it's theoretical: single-instance config flow, `_CARD_REGISTERED` guards the one `async_when_setup` registration per HA run, and `plan_card_resource` collapses duplicates on the next start. Now written into the module docstring rather than left implicit.

Re-verified end to end under the browser CI actually uses: green as shipped, still red with the registration stubbed, full e2e 58/58, integration 136, unit 1086, mypy clean against HA 2026.8.3 on Python 3.14.6, mutation 100%.

/q review Please re-review `879b86e` specifically. Two things I'd like challenged hardest: (1) is the `cast("ResourceStorageCollection", data.resources)` actually sound — is `resource_mode == "storage"` a real guarantee of that concrete type across HA versions, or have I traded an honest `isinstance` for a lie mypy can't catch? (2) does moving the LNA flag into `test.use({ launchOptions })` genuinely scope it, or does Playwright's worker reuse leak that browser flag into other specs sharing a worker? Also please re-check the new bundle-response assertion for a way it could pass spuriously.

---

PR #233, opened 2026-08-22
fix(options): stop the options flow wiping notifications and profiles

prestomation:
Fixes #

<sub>(No issue filed for this one — found while reviewing the options flow.)</sub>

## What changed

**The bug.** Home Assistant stores whatever an options flow returns from `async_create_entry` as `entry.options` **verbatim** — the whole object, not a patch. `HomeKeeperOptionsFlow.async_step_init` returned `user_input`, and the Configure dialog's form renders only 7 of the 10 option keys. So every save deleted `profiles`, `notifications` and `dismissed_companions`.

The damage was invisible. `notifier` reads those back through `options.current_options`, which defaults a missing key to `[]` — so saved notification bindings and profiles vanished and notifications simply stopped arriving, with nothing on screen to say why.

`options.py` exists to prevent exactly this ("the key list, defaults, and normalization live here so they can't drift"). The options flow was the one writer that bypassed it entirely — it imported neither helper.

### The fix

- **`options.merge_flow_input`** starts from `current_options` and changes only `FLOW_OPTIONS`. A flow key *missing* from the submission was **cleared** and resets; a key the form doesn't render is **preserved**. That distinction is load-bearing: `shopping_list_entity` has no voluptuous `default` precisely so a cleared picker drops out of the submission, and that absence is how the shopping-list mirror is turned off. A plain `{**current, **user_input}` would resurrect the old entity id — there's a test for that.
- **`current_options` collapses onto `_normalize`** over a new `_empty_options()` factory, so defaults and coercions are declared once. Reading and writing can no longer disagree about a key's shape, which is what makes `async_set_options`' `merged == base` short-circuit trustworthy. Flow submissions normalize too, so `NumberSelector`'s `7.0` is stored as `7`.
- **`config_flow._options_schema`** lifts the form to module level so a test can build it.
- **`options.py`'s two HA imports move under `TYPE_CHECKING`.** They were annotation-only; the module is now runtime-pure, so the fast unit tier covers the merge rules and it joins the mutation allowlist.

### Making it hard to repeat

There were **no tests for the config/options flow at all**, which is why this shipped. Three tiers now, plus drift guards that fail CI when a future option is forgotten:

| Guard | Catches |
| --- | --- |
| every `const.OPTION_*` reaches `_empty_options` | a new key invisible to every reader, deleted by the next flow save |
| every `const.OPTION_*` has a `_normalize` branch | a new key silently ignored by every writer |
| `strings.json` labels == `FLOW_OPTIONS` | a form field not in the tuple (discarded on save), or a tuple key with no field (cleared on save) |
| schema keys == `FLOW_OPTIONS` | the same, directly against the built `vol.Schema` |
| the shopping picker still has no `default` | a `default` here silently breaks "clear the picker to turn the mirror off" |
| `SET_OPTIONS_SCHEMA` accepts every option | a new key unreachable from the canonical service write path |

`tests/integration/test_options_flow.py` drives a real save over HA's flow API and asserts the panel-only keys come back byte-identical — the framework contract no mock can see. **It fails on the old code with `profiles == []`** (verified by reverting the fix and re-running against the container).

`quality_scale.yaml` claimed `config-flow-test-coverage: done` against coverage that did not exist; the comment now names what actually covers it. Conventions recorded in `.amazonq/rules/architecture-and-code.md` and `AGENTS.md`.

### Verified locally

| Check | Result |
| --- | --- |
| `pytest tests/unit` (py3.11, no HA) | 1176 passed, 2 skipped |
| full unit lane (py3.12 + HA, as CI runs it) | 1183 passed |
| `pytest tests/integration` (Docker, real HA) | 139 passed |
| `bash ci/test-mutation-python.sh` | **100.00%** (16/16), threshold 80% |
| `mypy custom_components/home_keeper` | Success, 43 files |
| `vale CHANGELOG.md` (added lines) | clean |

No version bump: this is a bug-fix-only PR and `0.16.0b5` is already published, so the entry sits under `## [Unreleased]`. No panel UI changed, so no screenshots or walkthrough edit.

## Checklist

- [x] Panel UI changed → n/a, nothing under `frontend/src/` changed
- [x] New user-facing feature → n/a, bug fix only

---
_Generated by [Claude Code](https://claude.ai/code/session_015T1NNjZWkp7aVA3fdKcqcm)_

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -83,6 +83,12 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
   callers instead of importing them here.
 - Keep the recurrence engine deterministic: functions take an explicit `now`
   rather than calling a clock internally.
+- `options.py` sits on the boundary: it takes a `ConfigEntry` and a `HomeAssistant`,
+  but only ever reads attributes off them, so those two imports live under
+  `if TYPE_CHECKING:` and the module is runtime-pure. That is what lets the fast unit
+  tier assert on the option merge rules every write path shares. Keep it that way —
+  reach for `hass.<something>` at import time and `tests/unit/test_options.py` stops
+  running without the HA harness.
 
 ## Datetimes & timezones
 - All datetimes are timezone-aware. Use `homeassistant.util.dt` (`dt_util`) at the
@@ -575,9 +581,30 @@ The appliance/asset feature lives in `assets.py` (pure model — no HA imports, 
   **coordinator's periodic refresh** uses to auto-delete completed one-offs
   (`recurrence.one_off_expired` collects expired ids; `_purge_expired_one_offs` deletes
   via `store.delete_task`), and `shopping_list_entity` (a `todo.*` entity id; `""` =
-  off) driving the shopping-list mirror. Put a new option's default/coercion in
-  `options.py` and add it to all three surfaces (flow schema, `SET_OPTIONS_SCHEMA`,
-  Settings `settingsSchema`) with `strings.json`/`services.yaml` parity.
+  off) driving the shopping-list mirror. Put a new option's default in
+  `options.py`'s `_empty_options` **and** its coercion in `_normalize` — both, or the
+  key is invisible to every reader and dropped by every writer — then add it to all
+  three surfaces (flow schema *and* `FLOW_OPTIONS`, `SET_OPTIONS_SCHEMA`, Settings
+  `settingsSchema`) with `strings.json`/`services.yaml` parity. `tests/unit/test_options.py`
+  fails the build if a `const.OPTION_*` misses either half.
+  - **The options *flow* merges; it never replaces.** Home Assistant stores whatever
+    an options flow returns from `async_create_entry` as `entry.options` **verbatim**
+    — the whole object, not a patch. The Configure dialog renders only
+    `options.FLOW_OPTIONS`; `profiles`, `notifications` and `dismissed_companions` are
+    panel-only, so returning `user_input` deleted every one of them on each save, and
+    notifications then stopped firing with nothing on screen to say why (`notifier`
+    reads a missing key back as `[]`). `async_step_init` returns
+    `options.merge_flow_input(entry, user_input)`, which starts from `current_options`,
+    resets any `FLOW_OPTIONS` key the submission *omits* — that absence is how clearing
+    the `shopping_list_entity` picker turns the mirror off, and it's the one field with
+    no voluptuous `default` for exactly that reason — routes everything through
+    `_normalize`, and touches nothing else. Guarded at three levels:
+    `tests/unit/test_options.py` (every `const.OPTION_*` reaches `_empty_options` and
+    `_normalize`; `strings.json`'s labels match `FLOW_OPTIONS`),
+    `tests/unit/test_config_flow.py` (the form's schema keys equal `FLOW_OPTIONS`, and
+    the shopping picker still has no `default`), and
+    `tests/integration/test_options_flow.py` (a real save over HA's flow API leaves the
+    panel-only keys byte-identical — the framework contract no mock can see).
   - **A single-value picker needs a clear-coercion on the panel side.** `ha-form`'s
     entity selector emits `undefined` when cleared, and JSON drops it — so the key
     never reaches `_normalize`'s `if key in updates` merge and "turn it off" silently
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -263,6 +263,13 @@ rules. Keep the rules and `AGENTS.md` consistent with each other.
   baselined silently on startup). A new event isn't done until it's in `docs/EVENTS.md`
   and, if device-facing, in `device_trigger.py` with translation-parity labels. Events
   need no new service. See `.amazonq/rules/architecture-and-code.md` and `docs/EVENTS.md`.
+- **An options flow merges; it never replaces.** Home Assistant stores what an options
+  flow returns from `async_create_entry` as the *entire* `entry.options`, and the
+  Configure dialog renders only `options.FLOW_OPTIONS` — so return
+  `options.merge_flow_input(entry, user_input)`, never `user_input`, or every key the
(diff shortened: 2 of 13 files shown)


prestomation:

This PR fixes the options flow replacing the whole `entry.options` object (deleting profiles, notifications and dismissed companions on every save) and adds drift guards so a future option key can't be forgotten the same way.

Topics I'd like scrutinized:

**Correctness**
- `options.merge_flow_input`: is the "a `FLOW_OPTIONS` key missing from `user_input` means the user *cleared* it" rule right in every case? Is there a path where voluptuous omits a key for a reason other than the user clearing it (validation error, a selector that submits nothing, HA version differences in how an options flow's `user_input` is built)? That would silently reset a field the user never touched.
- Collapsing `current_options` onto `_normalize(dict(entry.options), _empty_options())`: the id lists now coerce through `[str(x) for x in ...]` instead of `list(...)`. Any stored value where that changes observable behaviour for `problem_sync.py` / `companions.py`, which read `entry.options` raw and therefore *don't* get the coercion? Is that raw-read asymmetry itself a bug worth fixing here?
- `_empty_options()` hands out fresh lists each call — is every mutation path on the returned dict actually safe, including `_normalize`'s `merged = dict(base)` shallow copy?
- The options flow relies on the update listener's fire-and-forget reload while the service path awaits it. Does routing the flow through `merge_flow_input` change anything about that race, or about `async_set_options`' `merged == base` short-circuit now that reads and writes share one coercion table?

**Maintainability**
- Is `FLOW_OPTIONS` + a test-enforced schema/`strings.json` parity the right shape, or should the schema be generated from the tuple so the two literally cannot drift? I rejected generation as less readable — push back if you disagree.
- `tests/unit/test_options.py` parses `__init__.py` with `ast` to check `SET_OPTIONS_SCHEMA` coverage, and reads `strings.json` off disk. Is that too clever/brittle versus importing?
- Moving `options.py`'s HA imports under `TYPE_CHECKING` to make it runtime-pure: reasonable, or does it invite someone to reintroduce a runtime HA import and quietly break the fast unit tier?

**HA best practices**
- Is there a first-party Home Assistant idiom (e.g. `SchemaOptionsFlowHandler`) that would make this merge unnecessary, and would it handle the deliberate no-`default` shopping picker correctly?
- Anything wrong with how `tests/integration/test_options_flow.py` drives the flow over `/api/config/config_entries/options/flow`?

**Performance / security** — the flow now builds `current_options` on every form render; anything hot enough to matter, and any way user-supplied option values reach somewhere unescaped?

---

prestomation:
Thanks — took three of the four suggestions, and one was already covered.

**Taken:**

1. **Empty entries in the id lists.** Agreed, and it's slightly worse than you framed it: `[str(x) for x in ...]` turned a `None` into the literal string `"None"`, which would then sit in the entry forever matching nothing. Now `[str(x) for x in (updates[key] or []) if x]`, filtering on the original value so `None`/`""`/`0` all go — none is a valid registry id. Pinned by `test_id_lists_drop_empty_entries`.
2. **The `_normalize` coercion table.** Documented, one line per branch explaining *why* it exists, since each value can arrive from a form, a service call or an automation and none of those is obliged to send the stored shape.
3. **`merge_flow_input`'s docstring.** Now says it's the options flow's entry point and nothing else's, and points the other two surfaces at `async_set_options`.

**Already there:** the "no `default` on the shopping picker" test — `test_the_shopping_picker_has_no_default` in `tests/unit/test_config_flow.py`. It asserts `not callable(getattr(picker, "default", None))` and that the `suggested_value` description is what pre-fills it instead.

**On your correctness note** that the drift guards can't catch "forgot to add a default" on a *new* field: true, and I don't think a guard is the right tool, because a new field without a default isn't wrong — it just means "empty submission clears this", which is usually what you want for a picker. The hazard is the inverse (adding a `default` to `shopping_list_entity`), and that one is guarded. So I documented the contract in `_options_schema`'s docstring instead, where someone adding a field will read it.

**On the `[str(x)]` asymmetry with `problem_sync.py` / `companions.py` reading `entry.options` raw** — agreed it's harmless today for the reason you gave. I left it alone rather than widen this PR; converting those two to `current_options` is a reasonable follow-up but it's a behaviour change in the sync reconcile path and deserves its own change.

Also merged `main` in (0.16.0b6 shipped while this was open) and re-ran everything against the merged base: 1199 unit (py3.12 + HA), 143 integration against real HA, mutation **100% (37/37)** now that `_normalize` is in scope too, mypy and ruff clean, vale clean on added lines.

---

prestomation:
/q review Critical and skeptical again please, on the delta since your last review — most serious first, and don't withhold minor ones.

Since that review this branch: (1) filters falsy entries out of the id lists in `_normalize`, (2) documents the coercion table, `merge_flow_input`'s scope, and the `default` contract in `_options_schema`, and (3) merged `main` in (0.16.0b6, the card-resource change).

Specifically:

- **The new `if x` filter.** Is filtering on the *original* value right, or should it filter on the stringified one? `[str(x) for x in xs if x]` drops `0` and `False`; `[s for s in map(str, xs) if s]` would keep `"0"` and `"False"` and drop only `""`. Is there any Home Assistant id — entity, device, area, label, or a companion domain — where a falsy input is legitimate and I've just started silently dropping it? This runs on the `set_options` service path where an automation can send anything.
- **Is this a silent-data-loss risk?** It drops without logging. Should a dropped entry warn, the way `notifications.split_targets` warns on a rejected target?
- **The merge with `main`.** Conflicts were in `pyproject.toml` (`only_mutate`), `tests/conftest.py` (`_PURE_MODULES`) and `CHANGELOG.md` — all additive, both sides kept. `options` is ordered last in `_PURE_MODULES` because it does `from . import notifications, profiles, shopping`; is that ordering actually load-bearing given the stub-parent `__path__`, or am I relying on something incidental? And did the merge miss any interaction between the new `card_resource.py` and the options write path (it hooks entry setup/unload)?
- **Anything in the resolved `CHANGELOG.md`** that would confuse `release.yml`'s `notify-issues`, given `## [Unreleased]` now sits above the already-shipped `## [0.16.0b6]`?

---

PR #234, opened 2026-08-22
feat(inventory): measure part stock in units, not just whole spares

### LINKED ISSUE #220: A few feature requests and questions

prestomation:
Fixes #220

## What changed

Request 3 from #220: *"Stock, reorder_at and the delta on adjust_part_stock all seem to expect whole numbers. That falls over for anything you use by volume rather than by the item."* The reporter counts fabric softener in "thirds of a bottle" as a workaround, and every completion still takes a flat 1, so three top-ups read as three whole bottles gone and the reorder fires about 3× too early.

Two new optional fields on a part, and decimals throughout.

- **Stock unit** — what the numbers mean (`ml`, `m`, `bottles`). Free-form and purely presentational: Home Keeper never converts between units, it only shows the one you chose. The unit follows the amount to every surface — the part's chips, a linked task's detail line, the stock `number` on the appliance's device page, the low-stock `binary_sensor` attributes, and a new `unit` field on the three `home_keeper_part_*` stock events.
- **Used per completion** — how much one completion draws down. Unset means one whole spare, so a filter replacement behaves exactly as before; `0.33` makes three refills empty one bottle.
- **Decimals everywhere they were ints** — `stock`, `reorder_at`, `restock_quantity`, `consume_quantity`, and the `delta` on `home_keeper.adjust_part_stock` (`-250` and `-0.33` are both valid now).

A part that sets neither field is unchanged in every respect: it counts whole spares, uses one per completion, and its stock box still refuses "2.5 filters" (the `number` entity's step follows the part, and only goes fine once the part shows it deals in fractions).

**Two details worth review attention.**

Every write goes through `assets._round_stock`, which rounds to 3 places and **collapses a whole result back to an `int`**. That is what makes 0.1 taken ten times land on exactly zero instead of a `1.4e-17` remainder that would read as "still in stock" forever, and it keeps the ordinary count-the-filters case round-tripping as `3` rather than `3.0` through storage, the panel and every event payload. Validation also rejects NaN/inf explicitly — they survive `float()` and then compare falsely against every bound.

The two per-completion amounts have a **read-side floor of 1** (`part_consume_quantity` / `part_restock_quantity` return 1 for unset, junk, zero or negative), so parts written before these fields existed keep consuming exactly one spare. No storage migration.

### The bug the e2e run caught

`_PART_SCHEMA` in `__init__.py` is strict and coerced every quantity to `int`, so `add_asset`/`update_asset` rejected a part with a unit outright and would have floored a fractional stock — while the websocket path (a plain `dict`) accepted both. There was already a regression test for exactly this class of drift (`test_add_asset_service_accepts_part_stock`); it now has a sibling covering the new fields and decimals, plus one asserting the stock `number` reports its unit and a sub-1 step.

### Tests

- **Unit** — decimals, rounding, the int-collapse, NaN/inf rejection, the unit's length cap at exactly 16, the read-side floor for both per-use amounts, and the reported scenario end to end (a bottle at `0.33` per use lasts three completions; 0.1 ten times reaches exactly `0`).
- **Frontend** — `formatQuantity` (trailing zeros, float noise, unit appended), `selNumber`'s opt-in step, the part editor's new fields and their gating, and the chips reading "In stock: 0.67 bottles" / "Uses 0.33 bottles per completion".
- **Integration** — a measured part round-tripped through the service API, and the stock `number`'s unit/step.
- **e2e** — a new `measured-stock.spec.ts`: the unit on the appliance page and in the editor, a `-0.5` adjustment coming back as `749.5`, and the number entity's unit. The e2e store seeds a "Descaling solution" part (750 ml, reorder at 500, 250 ml per completion) so the screenshots and the tour have something real to show.
- **Mutation** — Python 89.85%, TypeScript 100% (both above the 80% gate).

Full unit suite, vitest, ruff, mypy, the integration suite (145 passed) and the e2e suite (64 passed) all run locally. mypy ran against HA 2026.2.3, since this sandbox's newest available Python is 3.14.0rc2 and pip backtracks below HA's floor — CI checks it against current.

## Screenshots

The part editor for the measured part:

The part editor for a descaling solution measured in millilitres: Stock 750, Reorder at 500, Stock unit ml, Used per completion 250

The same part on the appliance page, where the unit rides with both amounts:

A part row reading In stock: 750 ml and Uses 250 ml per completion

A part that sets no unit keeps its old behaviour, with the two fields simply empty:

The sediment pre-filter's editor: Stock 4, Reorder at 1, empty Stock unit and Used per completion, Auto-create buy task on with Restock quantity 4

The appliance page with the new part row in place:

The garage water heater's appliance page, its parts list now including the descaling solution measured in millilitres

## Checklist

      step through it (two beats: the measured part's chips, then its editor fields)
- [x] New user-facing feature → version bumped to the next beta (`0.16.0b7`, `manifest.json` +

---
_Generated by [Claude Code](https://claude.ai/code/session_01Lc3iKTSHDjfeTMmWhxfm8B)_


prestomation:
/q review Be critical and skeptical. This turns spare-part stock from integers into decimals and adds a per-part unit plus a per-completion draw-down amount. Surface the most serious issues first, but don't withhold minor ones.

**Correctness.** `assets._round_stock` rounds every write to 3 places and collapses a whole result back to `int` — find the case where that is wrong: a quantity whose true value needs more precision, a sequence of fractional draw-downs that drifts or fails to reach exactly zero, a comparison in `stock_transition` / `part_is_low` that behaves differently for `0.30000000000000004` vs `0.3`, or a place where `int` vs `float` leaks somewhere it matters (JSON storage, event payload consumers, entity state, the `todo`/shopping mirror). `stock_transition`'s out-of-stock test changed from `new == 0` to `new <= 0` — is any caller able to pass a negative `new`, and does the precedence between `out` and `low` still hold for fractional thresholds? Check the read-side floor in `_positive_quantity` (unset / zero / negative / junk → 1): is silently coercing a stored zero to one the right call at every call site, or does it hide a real user intent somewhere?

**Migration and back-compat.** Parts stored before this change carry no `stock_unit` / `consume_quantity`. Verify nothing in the read paths, the reconcilers, the buy-task lifecycle or the shopping-list mirror can behave differently for such a part than it did before this PR. Are there stored values (a `restock_quantity` of 0 written through the old service) whose behaviour changes?

**Privilege model.** `stock_unit` is a new user-supplied string that reaches the panel, the event bus, the `number` entity's `unit_of_measurement` and the low-stock sensor's attributes. It is capped at 16 chars and stripped, but not otherwise sanitised — check every render path for escaping, and check whether `card_projection` should still be withholding it from non-admin callers.

**HA best practices.** `number.py` now computes `native_step` and `native_unit_of_measurement` per part, from the coordinator's data, on every property read. Is a per-part varying step legitimate for a `NumberEntity`, does changing a unit at runtime confuse the entity/state machine or recorder, and should this entity carry a `device_class` or `suggested_display_precision`? The step-follows-the-part heuristic (fine step only once the part shows a unit or a fractional value) — is it too clever?

**Performance.** `native_step` walks the asset's parts on every read; `_round_stock` runs on every write. Anything hot enough to matter?

**Maintainability.** `_normalize_stock` is now shared by four fields with different rules layered on top (`consume_quantity` must be positive, the others may be zero). Is that the right factoring, or should the positivity rule live in one validator? Same question for the frontend's `MIN_POSITIVE_QUANTITY = 0.001` floor, which fakes an exclusive minimum a number selector doesn't offer.

---

prestomation:
/q review Second push (0d44dae) — docs-only fixes on top of the change you already reviewed: an `ANCHOR_ROUTES` entry in `website/scripts/sync-docs.mjs` so the README's cross-section anchor to the new measured-stock subsection resolves to the right generated page, and two `docs/EVENTS.md` sentences rewritten for the prose linter.

Worth a skeptical look at the anchor map specifically: it is hand-maintained, and nothing but the Docusaurus build catches a missing or wrong entry. Is there a cheaper guard (a check that every README same-page anchor either resolves within its own generated page or has a route), and is the entry I added pointing at the right page given "Stock you measure rather than count" is a `####` under "Parts & wear items" under "Appliances & virtual devices"?

Everything from the previous request still stands for the code itself.

---

prestomation:
Triaged the review — three findings acted on in `44c7b45`, three declined with reasoning.

**Acted on**

1. **`restock_quantity` write/read inconsistency** — a real find, thank you. The writer accepted zero while `part_restock_quantity` floored it to one, so a stored zero was a record claiming something the code never did. `_normalize_restock_quantity` now folds zero to `None`. I *folded* rather than *rejected* it, deliberately asymmetric with `consume_quantity`: that field is new so nothing can already hold a zero, whereas `restock_quantity` predates the validator and its service schema took `vol.Coerce(int)`, so a zero may be sitting in someone's store today — raising would make an untouched part unsaveable the next time its appliance is edited. Both the reasoning and the asymmetry are now in the docstrings and in `.amazonq/rules/`.

2. **`stock_transition`'s `new <= 0`** — both callers do clamp (`max(0.0, …)` in `consume_part_stock` and `adjust_part_stock`), so there's no live path, but "verify all callers" is only as durable as the person who last read them. Pinned with a test: a negative `new` reports out-of-stock rather than sliding into the restocked branch, and a separate test asserts neither caller can drive stock below zero.

3. **A cheaper anchor guard** — took the suggestion. `ANCHOR_ROUTES` moved into the pure `doc-map.mjs`, and `unroutedReadmeAnchors` now checks every README same-page anchor either resolves inside its own generated page or carries a route (skipping unpublished sections and fenced code). It reproduces the exact break this PR hit, in milliseconds instead of a four-minute site build. Seven cases in `tests/frontend/doc-anchors.test.js`.

**Declined**

4. **Regex-strip `stock_unit` to `[a-zA-Z0-9 /-]`** — this would break the field for most of the world. Legitimate units include `m³`, `°C`, `µL`, `Ω` and non-Latin scripts, and Home Keeper ships 16 locales; a household counting stock in `厘米` isn't an attack. The defence belongs at the render sites, and it's there: the part chips build through `escapeHTML(label)` and the task detail line through `escapeHTML(t(…))`, so the unit is escaped on every panel path. The other consumers are structurally safe (JSON event payloads, JSON entity attributes, and HA core's own rendering of `unit_of_measurement`), and the value never reaches a file path or a query. Happy to revisit if you can name a render path that doesn't escape.

5. **`suggested_display_precision = 3`** — this would make it worse, not better. HA formats to *exactly* that many decimals, so a part counted in whole spares would read "4.000 filters" on its device page. The current behaviour already shows "750 ml" and "0.67 bottles" correctly, because the value carries only the precision it actually has (`_round_stock` collapses a whole result back to `int`).

6. **Unit changes confusing long-term statistics** — doesn't apply: `NumberEntity` sets no `state_class`, so these entities never enter the statistics tables. Only recorder short-term state history is involved, which tolerates an attribute change.

On the int-collapse comparison you flagged as "makes reviewers nervous": `round(2.9995, 3)` is `2.999` (banker's rounding at three places, well clear of the next integer), and after rounding to three places any value within 5e-4 of an integer is *exactly* that integer as a float, so `rounded == int(rounded)` is reliable rather than lucky. I've left the reasoning in `_round_stock`'s docstring.

---

prestomation:
/q review Third push (44c7b45) — please re-review critically, most serious first, and don't withhold minor findings. Three things are new since your last pass:

**`_normalize_restock_quantity` folding zero to `None`.** Attack the choice: is there any path where a caller meant a literal zero and now silently gets the default of one at bump time? Does the fold interact badly with `_merge_parts` (incoming-wins, including a `None` that clears), with the level-triggered buy-task reconciler, or with the shopping-list mirror — for instance a part whose reminder is already open when its `restock_quantity` folds? And is the asymmetry with `consume_quantity` (reject vs fold) actually defensible, or should both behave the same way at the cost of one of the two failure modes?

**`unroutedReadmeAnchors` in `website/scripts/doc-map.mjs`.** This is a guard, so its own false negatives are the risk. Does `slugify` match GitHub/Docusaurus on headings the README actually has — duplicated headings that get a `-1` suffix, headings with inline code or links, non-ASCII? Does `stripFences` handle indented or nested fences? Can a link spanning a newline, a reference-style link, or an HTML `<a href="#...">` slip past the regex? And is `headingSections` right to let a later duplicate slug overwrite an earlier one?

**The new tests** — `stock_transition` with a negative `new`, and the zero-fold cases. Are they asserting behaviour or just re-stating the implementation?

Everything else in the diff is unchanged from your previous review, and my reasoning on the three findings I declined is in the comment above — push back if you disagree with any of it.

---

prestomation:
Good catch on the duplicate-heading gap — fixed in `e999eeb`, along with the link-form one.

**Duplicate headings.** You're right that this was a live false-alarm waiting to happen: two headings with the same text collapsed onto one slug in the map, so a valid `](#installation-1)` would have been reported as `unknown`. `headingSections` now numbers repeats the way GitHub does (`-1`, `-2`, …) and keeps the first occurrence on the bare slug. Tested both directions: a link to the second heading passes, and a link to a third that doesn't exist is still reported.

**Link forms.** Also right that the JSDoc claimed more than the regex delivered. `anchorsIn` now checks reference definitions (`[ref]: #a`) and raw `<a href="#a">` alongside inline links, each with a test. A link split across a newline is still missed — rather than pretend otherwise, that limitation is now named in the docstring, with the Docusaurus build as the acknowledged backstop. A guard that overstates its coverage is worse than one that doesn't, since it invites people to stop running the slow check.

**One correction on the `_merge_parts` walkthrough**, since it'll otherwise sit in the thread as fact: `_merge_parts` is *not* preserve-on-incoming-`None` for `restock_quantity`. Only `last_replaced` gets that treatment (plus the upload-only file fields, which are restored unconditionally); every other field including `restock_quantity` is plain incoming-wins, deliberately, so stock tracking and auto-buy can be switched back off. Your conclusion is unaffected — both paths land on `None`, and as you say `part_restock_quantity` floors zero to 1 on read regardless, so the fold only makes the stored record honest about what the code already did.

Your read on the two tests matches my intent, and I've left them as they are.

---

prestomation:
/q review Fourth push (e999eeb), scoped to `website/scripts/doc-map.mjs` and its test — everything else is unchanged from your last pass.

Be skeptical of the duplicate-heading fix specifically. `headingSections` now assigns `slug`, `slug-1`, `slug-2` in document order using a `seen` counter. Does that match GitHub and Docusaurus in the cases that bite: a repeat whose *first* occurrence is an H2 section title and whose second is an H4 in a different section (the counter is global across sections — is the anchor?), a heading whose literal text is already `foo-1` colliding with the generated suffix for a second `foo`, and headings that differ only in punctuation and so slugify identically? If any of those diverge, the guard is wrong in the direction that matters (a false pass).

Also worth attacking: `anchorsIn` runs three regexes over the same prose, so a link matched by two patterns would be reported twice — can that happen, and does a duplicate report matter? And `stripFences` runs before the reference-definition scan, so a `[ref]: #a` inside a fence is skipped while its `[text][ref]` usage outside one is not seen at all; is that the right asymmetry?

---

prestomation:
Thanks — no code change needed from this round, but three of those answers need correcting so the thread doesn't record them as facts about this repo.

**Repeat-heading numbering is not global on the generated site.** GitHub numbers across one document, so `#installation-1` is right in README — that's what the guard models. But the User Guide isn't one document: `sync-docs.mjs` splits README into a page per `##` section, so two same-named headings in *different* sections each end up the only one on their own page and get the bare slug there, no suffix. Docusaurus can't number across files it renders separately. The case that has to survive on one page — a repeat *within* one section — does number identically both ways, which is why the guard is still worth having. `eb6af63` (pushed before your reply landed) documents exactly this as a known hole.

**The `stripFences` asymmetry isn't a catch, it's a silent miss.** `anchorsIn` scans reference *definitions* (`[ref]: #a`), never `[text][ref]` *usages*. So a definition inside a fence is stripped and nothing is reported at all — the usage outside the fence is never looked at, so it can't "fail as unknown". Practically irrelevant (a reference definition inside a code fence isn't a thing anyone writes), but the mechanism is the opposite of what you described.

**The README does use reference-style links** — the badge block at the top, defined at the bottom of the file. They target `https://` URLs rather than anchors, so the `(#\S+)` definition pattern skips them, and they sit in the preamble that `splitByH2` never assigns to a section. Two reasons they can't false-positive, which is the useful part: the guard returning `[]` on the real README is a result that was actually exercised against reference links, not one that never met them.

Your other two points hold: punctuation-only differences collapse to the same base slug and get numbered, and the three patterns are structurally exclusive so nothing is double-reported.

**Status:** all 18 checks green on `eb6af63`, `mergeable_state: clean`.

---

PR #236, opened 2026-08-24
feat(sensor): custom meter start, and the reading in completion history

### LINKED ISSUE #235: Feature Request: Custom start usage for Usage/Meter Sensor Task

prestomation:
Fixes #235

## What changed

Three asks from the issue, all about a usage/meter task's relationship with its number.

### 1. A custom starting reading

A usage task anchored its baseline at whatever the sensor read the moment it was created, so every task made for a machine already partway through an interval started a full interval late. The task form now has a **Starting reading**: odometer at 48,000, oil every 10,000 miles, last change at 45,000 → *"3,000 of 10,000 used"* immediately, due at 55,000.

Most of the backend was already there. `models.normalize_sensor` has always validated and stored `sensor.baseline` (it's documented for glue integrations in `docs/INTEGRATING.md`), and `set_task_meter` already moved it — the panel simply never sent one. Blank still means "anchor at the live reading"; `0` is a real anchor, so the payload uses an explicit blank check rather than the `|| 0` idiom `target` uses, and the field is a bare number selector because `selNumber()` pins `min: 0` while the backend has no range gate.

**Last completed** is unhidden for sensor tasks at the same time. `build_task` has always handled a seeded completion on one — recording history without arming, expressly to anchor `sensor.also_every` — and only a `!isSensor` in the schema predicate hid it. Without it the calendar half of "every 10,000 miles or 12 months" restarts today while the meter half starts 3,000 miles back.

The task form with Starting reading set to 700 against a sensor reading 780 h, and the When it comes due strip reading: Counting from 700 h, so 80 h of 100 h is already used and the task becomes due at 800 h

### 2. The reading, in the history

Completing a sensor task in a numeric mode records what the sensor read, beside the note/cost/photo/who. `complete_task` resolves the number once — the caller's wins over the live read, which is what makes back-dating work — and hands the same value to `_reset_usage_baseline`, so the log and the anchor cannot drift apart.

Task detail for Replace printer nozzle: the history row for Jun 1 2026 shows at 660 h, and the progress bar above reads 120 h of 300 h used

### 3. Editable, and the meter follows

Correcting the reading on the **latest** completion of a usage task re-anchors `sensor.baseline` to it. This is a deliberate, documented exception to `update_completion`'s "never re-arms a task": the baseline *is* the reading at the last completion, so leaving it behind makes the store contradict its own history — 45,000 in the log beside a progress bar counting from 48,000. The task may then arm, which is the right answer. `delete_completion` / `move_completion` stay meter-neutral by the same rule: the anchor follows the reading, and neither changes one.

The completion edit dialog opened from a history row, with a Meter reading field seeded to 660 alongside Note, Cost and Who did it

## Decisions worth a second opinion

- **`reading` is deliberately *not* in `COMPLETION_METADATA_FIELDS`.** That list doubles as the allowlist for `completion_required_fields`, so adding it would let any task — a floating one with no sensor — be marked as requiring a number the user is never asked for and that can never arrive: an uncompletable task, from the panel, with no error explaining why. New `COMPLETION_CAPTURED_FIELDS` + `COMPLETION_ENTRY_FIELDS` (the union) instead. A disallowed `reading` raises rather than being dropped, mirroring `_reject_fields`.
- **A starting reading above the live one is not rejected.** `evaluate_usage`'s debounced meter-reset re-anchors it, which is right for a counter about to be zeroed; teaching it to respect a "user-set" baseline would need provenance on the field and would strand a task after a genuine reset. The form warns instead, and also says when a baseline already puts the task past target (it arms seconds after Create, which the user should see coming).

## Latent bugs found along the way

- **Four literals duplicated the completion field list** and each would have silently skipped a new key: `sensor.py`'s attribute loop, `websocket_api._ws_metadata`, `__init__._COMPLETION_METADATA_KEYS`, and a unit test. All four now read the const.
- **`_openCompletionEdit` didn't seed the field it doesn't own.** `update_completion` treats an omitted key as "clear it", so without seeding, editing a *note* on a sensor task would have silently destroyed the captured reading. The new e2e case is what pins it.
- **`setBackstop` in the screenshot harness inferred the switch state from how many number fields were on screen.** Adding one made the switched-off count equal the old switched-on count, so it stopped clicking and its own assertion still passed — the backstop shot would have quietly photographed the wrong form. It now reads the switch's `checked` state.
- A naive `last_completed` seed could reach a sensor task unqualified — every stored datetime is meant to be aware, and a naive one poisons the store until the file is hand-edited.
- `taskRecordsReading` disagreed with its Python twin for a sensor task with no binding; caught by porting the same test to both sides.

## Verification

Run locally, all green: `pytest tests/unit` (1265), `vitest` (621), `tsc --noEmit`, `ruff check`/`format`, `vale` (no hits on added lines), and the full Playwright suite (64) plus the new completion-dialog case on both the local Chromium and CI's headless shell. Mutation: **100%** Python, **96.85%** TypeScript against the 80% gate.

## Checklist

      step through it

<sub>Note on the walkthrough: the tour is extended to type a starting reading, but the capture currently fails locally at an earlier, unrelated step — the Markdown note editor, which this branch doesn't touch. Capture is a soft gate, so CI posts a note instead of blocking. Rebased onto `main` after #234, whose `0.16.0b7` section this change now shares.</sub>


prestomation:
/q review Please be critical and skeptical — surface the most serious problems first, but don't withhold minor ones. This change makes a usage sensor task's meter baseline settable at creation and correctable from the completion history, and records the bound sensor's reading on every completion. Areas I most want torn into:

**Correctness of the baseline invariant.** The claim the whole design rests on is "a usage task's `baseline` *is* the reading on its latest completion". Look for paths that break it: `store.complete_task` resolves one reading and feeds it to both the history entry and `_reset_usage_baseline` — is there any ordering, exception, or early-return where those diverge? `store._reanchor_from_completion` only re-anchors when `ts == task["last_completed"]` and the task is usage mode — is that the right predicate given `move_completion` can change which completion is latest, and given `delete_completion` deliberately leaves the meter alone? What happens on a task whose completions are out of chronological order (a seeded back-date), where `last_completed` and "last appended" disagree?

**The re-arm exception.** `store.update_completion` now mutates `sensor.baseline`, which can make the watcher arm the task on its next tick. That intentionally breaks its old "never rewinds or re-arms" contract. Is the narrowing sound, or can this fire in a case a user would find surprising — e.g. editing an *old* completion on a task that has since been re-baselined by `set_task_meter`?

**Aliasing.** `recurrence.apply_completion(dict(existing), …)` is a shallow copy, so `updated["sensor"] is existing["sensor"]`. Both `_reset_usage_baseline` and `_reanchor_from_completion` mutate that shared dict. Is there a path where the mutation lands on the stored task before a later validation failure, leaving the store inconsistent?

**The captured-vs-requirable field split.** `reading` is in `COMPLETION_CAPTURED_FIELDS`, deliberately not `COMPLETION_METADATA_FIELDS` (which gates `completion_required_fields`). Is `COMPLETION_ENTRY_FIELDS` now used everywhere an entry's keys are enumerated, or did I miss a literal? Does `normalize_completion_metadata(allow_reading=...)` raising rather than dropping create a problem for any existing caller — e.g. a glue integration replaying metadata from a `home_keeper_task_completed` payload onto a non-sensor task?

**Back-dating and edge values.** A back-dated `complete_task` records today's live reading unless the caller passes one. Is the mitigation complete? Also check `0` and negative readings survive every hop (panel → `metadataMsg` → websocket → `normalize_completion_metadata` → `recurrence.update_completion`'s clear-on-empty check), since `0` is falsy and a legitimate meter value.

**HA best practices / performance.** `_reanchor_from_completion` scans `completions` linearly (capped at 500) on every edit; `sensor.py` now emits `usage_baseline` and `last_completion_reading`. Anything there that misbehaves on a large history, or that should be a repair issue rather than silent behaviour?

**Frontend.** `forms.ts` `usageHint` has several `??` chains and four branches — any input that renders `NaN`, `undefined`, or a misleading sentence? `buildTaskPayload` only sends `sensor.baseline` inside the `usage` branch to avoid the backend's `_reject_fields`; is that airtight across mode switches? And `utils.readingUnit` falls back to the entity's `unit_of_measurement` for threshold bindings but returns nothing for `attribute` bindings — is that the right call?

---

PR #237, opened 2026-08-24
release: 0.16.0b8

### LINKED ISSUE #235: Feature Request: Custom start usage for Usage/Meter Sensor Task

prestomation:
## What changed

#236 (custom starting reading + reading-in-history for usage/meter tasks) merged to
`main` reusing the `0.16.0b7` version string, but `v0.16.0b7` had already been tagged
and published ~50 minutes earlier by #234's merge. `release.yml` read the unchanged
version, saw the `v0.16.0b7` tag already existed, and skipped silently — so the
feature landed on `main` without ever going out as an installable beta.

This PR has two parts:

**1. The release bump #236 should have carried:**
- `manifest.json` + `const.py` (`PANEL_VERSION`) → `0.16.0b8`
- `CHANGELOG.md` — moved the "Start a usage meter…" entry (`Fixes #235`) out of the
  now-immutable `## [0.16.0b7]` section (which matches what actually shipped in that
  release: the stock-units feature and the options-flow fix) into a new
  `## [0.16.0b8]` section

**2. A CI safeguard against this recurring** (per Amazon Q's review on this PR):
`lint.yml`'s new `changelog-release-gap` job runs
`ci/check-changelog-release-gap.py`, which compares `CHANGELOG.md`'s top
`## [X.Y.ZbN]` section between a PR's merge-base and `HEAD`. It fails when the
version is unchanged but the section's content changed *and* that version is
already a published release tag — exactly the gap #236 fell into. Verified against
the real repo history: run against #236's actual base/head commits it fails with the
message telling the author to bump the version; run against this PR's own diff (a
genuine version bump) it passes. Documented in `RELEASE.md`, `AGENTS.md` and
`.amazonq/rules/testing-and-workflow.md`.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v` — 1276 passed, 2 skipped; `ruff check`
      / `ruff format --check` clean)
- [x] `CHANGELOG.md` updated — `(Fixes #235)` moved into the new `0.16.0b8` section
- [ ] Panel UI changed → N/A (no UI change)
- [ ] New user-facing UI surface → N/A (#236 already extended the tour; this PR just
      fixes the version it ships under, plus a CI-only safeguard)
      `const.py`)

### DIFF
--- a/.amazonq/rules/testing-and-workflow.md
+++ b/.amazonq/rules/testing-and-workflow.md
@@ -229,6 +229,17 @@ gate); CI publishes it to the job summary.
   `PANEL_VERSION` → next `bN`, plus a matching `## [X.Y.0bN]` CHANGELOG section)
   so it reaches beta testers. Fold into the current top beta if it's still
   unreleased; otherwise open the next `bN`. Bug-fix/developer-only PRs don't.
+- **`lint.yml`'s `changelog-release-gap` job guards the "fold into the current top
+  beta" rule above.** `release.yml` keys off `manifest.json`'s version and skips
+  tagging silently when that version is already tagged — so a PR that edits the top
+  `## [X.Y.ZbN]` CHANGELOG section without bumping the version, after that version
+  has already shipped, merges clean and then never actually ships (#236 did exactly
+  this: it landed ~50 minutes after `v0.16.0b7` was tagged, reusing the b7 heading,
+  and sat unreleased on `main` until #237 caught it). `ci/check-changelog-release-gap.py`
+  compares the top section between the PR's merge-base and `HEAD`: unchanged content,
+  or a version bump, or a still-unreleased top section, all pass; changed content
+  under an already-tagged version fails. Pure logic lives in `check()`, tested in
+  `tests/unit/test_check_changelog_release_gap.py`.
 - **Always add the `preview-release` label to a new-feature PR** once it's open, so
   `preview-release.yml` publishes an installable ephemeral pre-release
   (`X.Y.Z.dev<pr>`) from the PR head for pre-merge HACS testing (auto-deleted on
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -59,3 +59,19 @@ jobs:
           fail_level: any
           reporter: github-pr-review
           filter_mode: added
+
+  changelog-release-gap:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # ci/check-changelog-release-gap.py needs the base branch's history and
+          # every release tag to tell an already-shipped section from one still
+          # being iterated.
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: actions/setup-python@v7
+        with:
+          python-version: "3.13"
+      - name: Guard against editing an already-released CHANGELOG section
+        run: python3 ci/check-changelog-release-gap.py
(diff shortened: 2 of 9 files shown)


prestomation:
/q review Please review critically and skeptically. This is a release-only PR (no code changes): it bumps `manifest.json`/`const.py` `PANEL_VERSION` from `0.16.0b7` to `0.16.0b8`, and reorganizes `CHANGELOG.md` by moving the "Start a usage meter…" entry (`Fixes #235`) out of the `## [0.16.0b7]` section into a new `## [0.16.0b8]` section, since `v0.16.0b7` had already been tagged/released before that entry's commit (#236) landed on `main`.

Please scrutinize:
- **Correctness**: does the CHANGELOG section split correctly preserve what actually shipped in the real `v0.16.0b7` GitHub release vs. what's new in `0.16.0b8`? Any markdown/heading issues that would break `ci/release-issues.py`'s parsing (used both for the release notes body and the `notify-issues` job)?
- **Process/maintainability**: is bumping straight to `b8` (skipping nothing) the right fix, or should anything else be reconciled given `release.yml` silently skips when a tag already exists?
- Any other release-process risk from this pattern (a feature merging without a version bump) recurring — should there be a CI check that flags an unchanged `PANEL_VERSION` when a `## [X.Y.ZbN]` CHANGELOG section already has a matching published tag?

Surface the most serious issues first, but don't withhold minor ones.

---

prestomation:
/q review Thanks for the earlier review and the CI-safeguard suggestion — I've added it in a second commit. Please review critically and skeptically, specifically:

- **`ci/check-changelog-release-gap.py`**: it compares `CHANGELOG.md`'s top `## [X.Y.ZbN]` section between the PR's merge-base and `HEAD` (via `git show <ref>:CHANGELOG.md` + `git merge-base`), and fails when the version is unchanged, the section's *content* changed, and a `v<version>` tag already exists. Is the four-way case split in `check()` (version bumped / content unchanged / still-unreleased top section / already-released-and-edited) actually exhaustive and correct? Any way a legitimate CHANGELOG edit (e.g. wording fix to an old but still-top, unreleased section) gets a false positive?
- **CI wiring** in `lint.yml`: the new `changelog-release-gap` job uses `fetch-depth: 0` + `fetch-tags: true` so tags are available for `git rev-parse --verify refs/tags/vX.Y.Z`. Any risk this doesn't reliably fetch tags on `actions/checkout@v7`, or performance concern from the full unshallow fetch on every PR?
- **Testability**: `check()` takes `tag_exists`/`section` as injected callables so `tests/unit/test_check_changelog_release_gap.py` can test the pure logic without a real git repo. Is that the right boundary, or should more of the git plumbing (`_git`, `_file_at`, `_tag_exists`) be exercised directly rather than mocked out entirely?
- I manually verified the script against real history: run with `--base` pointing at #236's actual parent commit, `check()` correctly flags the bug this PR fixes; run against this PR's own diff (a real version bump) it passes clean. Does that verification method have a gap?

Surface the most serious issues first, but don't withhold minor ones.

---

prestomation:
/q review Thanks — addressed the one nit from the last review: added a docstring note in `check()` explaining that a version bump short-circuits before any tag lookup, plus a test (`test_version_bump_short_circuits_before_checking_any_tag`) that passes a `tag_exists` which raises `AssertionError` if called, proving it's never invoked once the top heading itself changed. Please confirm this closes out the review, and flag anything else critical/skeptical you still see.

---

PR #240, opened 2026-08-24
feat(sensor): meter countdown on the overview, and restore progress on undo

### LINKED ISSUE #235: Feature Request: Custom start usage for Usage/Meter Sensor Task

prestomation:
Fixes #235

The two follow-ups from the latest comment on #235 (on top of the merged #236).

## What changed

**Undoing a meter task's completion keeps your partial progress (the bug).**
Completing a usage/meter task moves `sensor.baseline` forward to the completion
reading, so its progress resets to zero. Deleting that completion (undoing an
accidental "done") left the baseline stranded there, so the meter stayed at zero
instead of reverting to where it was — 3,000 of 10,000 miles was lost.

- Each usage completion now records the baseline it replaced as `meter_start`
  (internal bookkeeping stamped by the store, so it survives edits, moves and
  restarts).
- `delete_completion` restores it when the **anchoring** (latest) completion is
  undone; undoing an older row leaves the baseline alone. An entry recorded before
  `meter_start` existed falls back to the latest remaining completion's reading,
  else clears so the watcher re-anchors.
- The decision is a pure helper, `sensor_tasks.baseline_after_delete`, with unit
  tests, plus an integration regression for the complete→delete store path.

**Meter countdown on the Tasks overview (the enhancement).** A dormant usage/meter
task used to read only "Monitored". It now reads a live countdown — *"in 7000
miles"* — the meter analogue of the *"in 3 days"* a time-based task shows.
`dueLabel` takes `hass` and computes `target − max(0, reading − baseline)` via a new
`meterRemaining` helper; threshold/state tasks and meters with no reading yet still
read "Monitored". New `due.in_units` key across all 16 locales.

Tasks overview showing the 'Replace printer nozzle' usage task reading 'in 180 h' in the Monitored group, while triggered battery tasks still read 'Monitored'

Bumps to `0.16.0b9` (`manifest.json` + `const.py`) with a matching CHANGELOG
section and a README note.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`, `bash ci/test-frontend.sh`) — 1285
      unit + 637 frontend green; mutation gates pass (TS 95.8%, Py 95.4%)
- [x] `CHANGELOG.md` updated — user-facing changes only, with `(Fixes #235)`
- [x] Panel UI changed → current screenshot committed under `docs/images/` and
      embedded above with an HTML `` tag (also asserted in
      `screenshots.capture.ts`)
- [ ] New user-facing UI surface → n/a (label text on an existing surface; no new
      tour step)
      `const.py`), `preview-release` label to be applied

---
_Generated by [Claude Code](https://claude.ai/code/session_019qySFtxAebaziFS6isvSv8)_


prestomation:
/q review Please be critical and skeptical, most serious issues first, and don't withhold minor ones. Focus on:

- **Correctness of the meter-baseline restore.** `store.delete_completion` now restores `sensor.baseline` via `sensor_tasks.baseline_after_delete`. Scrutinize the cases: undoing the *only* completion (restore `meter_start`), undoing the *latest* of several (restore its `meter_start`, which should equal the prior reading), undoing an *older* row (must NOT move the baseline), and a pre-`meter_start` legacy entry (fall back to latest-remaining reading, else clear to `None`). Is `was_latest = ts == existing["last_completed"]` the right anchor test, computed before removal? Any interaction with `move_completion`, `update_completion` re-anchoring, or the watcher's re-baseline/reset-candidate logic that could double-move or contradict the baseline?
- **`meter_start` stamping.** It's stamped directly on the completion entry in `complete_task` (not a `COMPLETION_*_FIELDS` metadata key). Does it correctly survive `update_completion` (fields-only edit), `move_completion` (whole-entry), storage save/load, and the `MAX_COMPLETION_HISTORY` trim? Any path that rebuilds/filters completion entries and would drop it?
- **Frontend countdown.** `dueLabel`/`meterRemaining` compute `target − max(0, reading − baseline)` from live `hass` state. Edge cases: unavailable/non-numeric readings, attribute vs state readings, a meter past target (clamp), unit resolution via `readingUnit`. Any place a stale or missing `hass` yields a wrong label instead of "Monitored"?
- **HA best practices / privilege model, maintainability, and any performance concern** in the changed store paths.

---

PR #244, opened 2026-08-25
feat(task-mirror): sync profile-filtered tasks to external to-do lists

### LINKED ISSUE #239: Glue integration: two-way task sync between Home Keeper and Todo lists
## Summary

A glue integration that mirrors Home Keeper tasks to Todoist and keeps completions in sync in both directions, so your home maintenance chores show up in Todoist alongside everything else — and checking a task off on either side marks it done on both.

## Intended experience

- **Tasks appear in Todoist automatically.** When a Home Keeper task becomes due (change the fridge filter, replace the water filter, etc.), a matching task shows up in a configured Todoist project. You don't have to manually create anything in Todoist.

- **Check off in either place, it's done everywhere.** Complete the task in Home Keeper → it's marked complete in Todoist. Check it off in Todoist (from your phone, the web, anywhere) → it's marked complete in Home Keeper, with the completion logged in its history.

- **Recurrence just works.** When a recurring Home Keeper task reschedules after you complete it, a fresh task appears in Todoist for the next due date. The old completed task stays in your Todoist history as a record. No manual cleanup needed.

- **Self-healing.** If HA restarts, or the sync was interrupted, the integration reconciles both sides on startup — tasks that should exist in Todoist are created, tasks completed on one side but not the other are brought in sync. Nothing gets duplicated or lost.

- **Configurable.** Choose which Todoist project to sync into, toggle two-way sync on or off, and optionally filter so only certain Home Keeper tasks (by area or tag) show up in Todoist. It works fine if Todoist or Home Keeper is removed — nothing breaks, the tasks just stop syncing.

- **Connected companion.** The integration registers itself under Home Keeper's companion discovery, so it shows up as a connected integration in Home Keeper's settings alongside any other companion integrations.

prestomation:
Fixes #239

## What changed

Adds **to-do list sync**: Home Keeper keeps an external `todo.*` entity — a Todoist project via HA's core todoist integration, Google Tasks, a `local_todo` list — in step with the tasks a Profile selects, two-way.

Issue #239 asked for a Todoist glue integration in a separate repo. Through design discussion this generalized to *any* to-do entity and moved inside Home Keeper, reusing the shopping-list mirror's architecture: a pure planner (`task_mirror.py`) plus an HA-aware driver (`task_mirror_sync.py`), the same split as `shopping.py`/`shopping_sync.py`.

**Sync is configured inside a Profile.** *Settings → Profiles → the profile → Sync to a to-do list* holds the list picker and two switches. The Profile is both the filter and the timing: its **Include** tier decides when a task appears (*Overdue only* when it falls due, *Overdue and due soon* three days ahead, *Every scheduled task* as soon as it is scheduled). One list per Profile; a household that wants two lists writes two Profiles, which it needed anyway to say what goes on each. Clearing the picker is both the off switch and the delete, so there is no separate mirror record and no way to point at a Profile that no longer exists.

A profile editor with its Sync to a to-do list group open, showing the list picker and both switches

The Family chores to-do card showing mirrored tasks with their due dates

### How it behaves

- **Outbound.** A task the Profile surfaces gets an item carrying its due date and notes, where the list supports them (capability-gated in the planner, so a list without due dates is never told to update the same item forever). A task that stops qualifying has its open item removed.
- **Inbound.** Ticking an item completes the task with `origin: home_keeper_todo_mirror`, logged in the completion history like any other Done.
- **Recurrence.** Completing in Home Keeper ticks the item off and leaves it as the record; the next due cycle adds a fresh item beside it.
- **Two per-Profile switches.** *Two-way sync* off makes the inbound direction inert. *Treat removed items as completed* (on by default) exists because Todoist hides checked-off items from Home Assistant, so a vanished item reads as done — uid-gated, so an add that never confirmably landed is re-added rather than completed. Off gives strict recreate-on-delete for providers that report completions properly.
- **Deliberate exclusions.** Auto-buy reminders (the shopping mirror owns those) and completion-blocked tasks — today a synced `problem` sensor, which belongs in a Profile but not on a list, since only the integration owning the sensor can decide it is fixed. An item nobody can ever tick off is worse than no item, the same call `notifications.actions_for` makes about buttons.
- **Self-healing.** Passes run off task-lifecycle events (gated by a pure `needs_pass` drift check so an idle mirror costs no service calls), the target lists' state changes, a periodic sweep, and an HA-started initial reconcile. Every `todo.*` call is best-effort; a failed op restores its bookkeeping so the next pass retries rather than orphaning a line.

## Testing

Re-run on the rebased tree:

- **1515 unit tests**, including a full planner edge-case matrix (vanish × two-way × uid, frozen entries, target switches, adoption, cross-mirror claim sets, the unreadable-list rule) and a driver suite covering failure semantics and the re-entrancy budget.
- **73 Playwright e2e tests**, against a seeded `local_todo` list, asserting on list contents in both directions — including that clearing the picker takes that Profile's lines back off. `main`'s id-visibility specs pass alongside the new `task-mirror` ones.
- **695 frontend tests**, `tsc --noEmit` clean, ruff lint and format clean.

Carried over from before the rebase, not re-run here: the Docker integration suite, mutation (98.96% on the changed Python, 100% on the changed `forms.ts` lines — gate is 80%), `mypy` and vale. This environment has no Python ≥ 3.14, so installing the current Home Assistant for `mypy` would silently resolve a stale release rather than the one CI checks; `lint.yml` covers it.

## Notes for review

- **Rebased onto `main` at 640a7b7** (`0.18.0b3`). Conflicts were `CHANGELOG.md`, `pyproject.toml`, `tests/conftest.py` and three screenshots; the `pyproject`/`conftest` ones were both sides adding an entry to the same list (`resolve` and `task_mirror`), resolved by keeping both, with `task_mirror` left in its documented position ahead of `options`.
- **Version is now `0.18.0b4`.** The branch carried `0.18.0b3`, which shipped from `main` as a published tag while this PR was open — PEP 440 sorts it at or below the release, and `changelog-release-gap` fails a PR that edits an already-released section. The feature moved into a fresh `## [0.18.0b4]` section above `main`'s `0.18.0b3` one, which is left intact.
- **Screenshots recaptured on the merged build.** `17-panel-settings`, `22-panel-notifications` and `profiles-card` were recaptured on both sides and conflicted, so they were regenerated here to show the Profile's Sync group and `main`'s id-visibility panel together (the Settings capture reads `v0.18.0b4`). The rest of `docs/images/` is untouched — the harness rewrites every PNG, and that churn isn't this PR's.
- #248 made synced `problem` sensors ordinary members of a Profile, which silently pulled them into every mirror's task set. The mirror now filters as a *delivery* surface via `managed_by.completion_blocked`, which is the split `profiles.matches_filter`'s own docstring prescribes.
- The sync UI began as a separate Settings card with per-mirror rows and a profile dropdown; it was reworked into the Profile after a design review, which removed the "mirror names a profile that is gone" state entirely.
- `docs/EVENTS.md`, `.amazonq/rules/architecture-and-code.md`, `README.md` and an `IDEAS.md` note (a deferred `profile` filter for `list_tasks`) all updated.

## Checklist

      step through it


prestomation:
/q review This draft is landing in stages; so far it carries the interface layer for the new task mirror (To-do list sync): `task_mirror.py` normalizers + op dataclasses, the `OPTION_TASK_MIRRORS`/`ORIGIN_TODO_MIRROR` consts, the options plumbing (`options.py` `_normalize` branch, `SET_OPTIONS_SCHEMA`), the option-guard test updates, and docs (EVENTS.md origin, .amazonq rules block, CHANGELOG 0.17.0b1). Please review critically and skeptically, most serious issues first, without withholding minor ones. Scrutinize: **correctness** of the options merge semantics for the new panel-only `task_mirrors` key (partial updates, flow-save preservation, normalization idempotency, garbage coercion); **the frozen planner interface** — do the op dataclasses and documented bookkeeping shape (`mirror_key`, uid-gated vanish, `last_completed` snapshot detector) have gaps or ambiguity that will bite the state machine implementing them; **naming/maintainability** vs the sibling shopping mirror; and **docs accuracy** — whether the EVENTS.md and rules text promise anything the described design can't keep. Flag anything in the CHANGELOG bullet that reads wrong for a user.

---

prestomation:
/q review New commit `8b138c0` adds the panel's To-do list sync settings editor (forms.ts schema + round-trips, panel.ts card cloned from the profiles editor, ten `mirror.*` keys across all 16 locales, vitest + Stryker coverage). Please review critically, most serious first: **correctness** — the `boolOr` round-trip defaulting (missing booleans must read as ON to match the backend normalizer, unlike the notification pattern's `!!`), the `profile_id` `''`↔`null` mapping, the debounced optimistic persist vs. a reload race, delete-removes-the-right-row; **security/escaping** — friendly names and any user content reaching innerHTML through the new card; **HA best practices** — the `selEntity` exclude of Home Keeper's own todo entities, and whether a mirror row with no entity behaves sanely everywhere; **i18n** — anything in the new locale strings that reads wrong to a native speaker; and **maintainability** vs. the profiles/notifications editor patterns it clones. Screenshots for the new surface land with the e2e phase of this draft.

---

prestomation:
Triage of the review above — fixes taken, push-backs with reasoning:

**Taken**
- *Explicit flow-preservation assertion for `task_mirrors`* — added in `5dd4465`: the key now sits beside profiles/notifications in `test_merge_flow_input_preserves_the_keys_the_form_does_not_render`, with a seeded-fixture sanity assert.
- *`profile_id` docstring should say `""` coerces to `None`* — agreed; landing with the planner commit (that file is mid-implementation on this branch).
- *`CAP_*` constants should point at where capabilities come from* — agreed, same commit: the comment will name `task_mirror_sync` as the layer deriving them from `supported_features`.
- *`require_tag_scan` sharp edge* — real, and by design (a phone tap is exactly the remote completion the physical-scan flag exists to refuse). The driver will warn once, drop the bookkeeping entry, and let a fresh open item reappear; the README section this PR still owes will document it.

**Push-backs**
- *Scalar-in-list "phantom config" gap*: `normalize_task_mirrors` filters with `isinstance(m, dict)` **before** calling the single-item normalizer, so `["string"]` is dropped, not normalized — the `raw if isinstance(raw, dict) else {}` line only defends direct single-mirror calls. An empty `entity_id` after normalization must stay legal regardless: the panel's Add button deliberately creates an unconfigured row the user then fills in.
- *`mirror_key` colon collision*: keys are parsed with `split(":", 1)` — from the left, once — and the mirror id (always ours, uuid hex) can never contain `:`, so a task id containing colons still round-trips. That invariant is stated in the `_KEY_SEP` comment.
- *Undo scenario*: already specified — the module docstring's snapshot rule ("strictly newer means completed since mirrored… an undone completion moves the value backwards and therefore reads as plain content drift") is precisely the undo case; the planner suite pins it with boundary tests.
- *In-flight add vs. failed add*: the driver pops tracking on a failed `add_item` (retry next pass) and the uid-gate keeps an unconfirmed add from ever completing a task; a landed-but-not-yet-listed item is bound by summary adoption on the next pass. This is the shopping mirror's exact, shipped design, and core todo entities refresh state before `todo.add_item` returns, so the stale-read window is theoretical.
- *`CompleteOp` symmetry*: it stays in the plan on precedent — `shopping.SyncPlan` carries `complete` the same way, and keeping every decision in the pure planner is what makes the state machine mutation-testable.
- *`ORIGIN_TODO_MIRROR` naming*: origins name **where the completion came from** for automation authors (`home_keeper_shopping_list` = the shopping list, `home_keeper_todo_mirror` = a mirrored to-do list); parallel with the module name matters less than that read.
- *EVENTS.md two-way caveat*: capability degradation (a list without `UPDATE_TODO_ITEM` etc.) is documented in the architecture rules and surfaced by a warn-once in the driver; EVENTS.md stays about event semantics. The README section will carry the user-facing requirements.

---

prestomation:
/q review Commit `e487d95` lands the pure task-mirror planner (`plan_sync` two-phase state machine, `desired_by_mirror`, `completed_since`, `lists_to_read`, `needs_pass`) plus its 94-test suite (98.7% mutation score). Please review critically and skeptically, most serious first: **correctness of the state machine** — the vanish × two_way × vanish_as_completed × uid matrix, the settled/frozen-entry rules (a tick with two-way off must freeze, a tick HK already knows about must NOT complete again but must allow a same-pass fresh add), target-switch remove+re-add, the cross-mirror claim set with duplicate summaries, adoption never claiming a completed item; **timezone/parsing** — `completed_since` instant comparison across offset spellings and naive values, the `next_due` → date-only reduction near midnight/DST; **the needs_pass gate** — any drift axis it can miss that neither the state listeners nor the sweep would catch; and **the two docstring follow-ups from your earlier review** (empty-string `profile_id`, `CAP_*` provenance) now in this commit. Surviving mutants are listed in the commit message reasoning — challenge any equivalence claim that looks wrong.

---

prestomation:
/q review Commit `b936214` lands the sync driver (`task_mirror_sync.py`), the store's `task_mirror_items` bookkeeping, coordinator/`__init__` wiring, and the now-public `notifier.effective_filter_tasks`. The feature is functionally complete; the preview release rebuilds from this head. Please review critically and skeptically, most serious first: **loop prevention and re-entrancy** — inbound completions echoing through the task-event listeners, the running/pending fold, the pass budget, and whether any event path can starve or livelock; **failure semantics** — restore-on-failed-remove/update, pop-on-failed-add, the TaskValidationError (scan-required) drop-don't-restore path, and KeyError handling; **HA correctness** — listener teardown via `entry.async_on_unload` on reload, the HA-started deferral, blocking service calls from within a coordinator refresh path (the sweep schedules a task rather than awaiting — check for deadlock or ordering hazards), and `async_settle_buy_tasks` re-entry; **performance** — eight bus listeners gated by `needs_pass`, registry reads per pass, and anything that would make an idle mirror cost service calls; and **store semantics** — the silent no-event map and unchanged-write skip. Remaining in this PR after this commit: integration/e2e specs, screenshots, and the walkthrough step.

---

prestomation:
On the one open verification item: the sweep never blocks the coordinator refresh. `async_schedule_sweep` is a `@callback` that only does `hass.async_create_task(self.async_sync(force=True))` and returns — the blocking `todo.*` service calls run in that detached task, never awaited from `_async_update_data`. This is the same shipped pattern as `shopping_sync.async_schedule_sweep`, and the settle-ordering point is the deliberate persist-before-settle sequence the module comment documents (also inherited from the shopping mirror).

---

prestomation:
Rebased onto `main` (was 4 commits behind; `0.17.0` stable plus `0.18.0b1`/`b2` shipped in the meantime). Three things the rebase forced, worth flagging:

- **Version corrected to `0.18.0b3`.** The branch carried `0.17.0b1`, which has since become a published tag — PEP 440 sorts it *below* the shipped `0.17.0`, so HACS would have offered the stable as an "upgrade". The CHANGELOG section moved to `## [0.18.0b3]`, keeping `(Fixes #239)`.
- **Completion-blocked tasks are now filtered out of mirrors** (`0b594b65`). #248 made synced `problem` sensors ordinary members of a Profile, which silently pulled them into every mirror's desired set. They belong in a Profile but not on a to-do list: only the integration that owns the sensor can mark one done, so the item could never be ticked off — and `vanish_as_completed` would keep resurrecting it. The mirror now filters as a *delivery* surface, reading the same `managed_by.completion_blocked` the panel and notification buttons use, which is the split `profiles.matches_filter` itself documents.
- **README follows the renamed Include tiers** (*Overdue only* / *Overdue and due soon* / *Every scheduled task*).

Local gates after the rebase: 1447 unit tests, ruff, `mypy` over all 46 files, 673 frontend tests, panel build — all green.

/q review Please re-review the rebased branch, most serious first, focused on rebase fallout rather than re-litigating earlier rounds: **semantic merge hazards** — my planner and driver now sit on top of #248/#251 (problem sensors in Profiles, renamed status tiers) and #250 (to-do due dates in local time); does the new `is_completion_blocked` filter belong in `desired_by_mirror`, and does anything else in the mirror still assume pre-#248 filter behaviour? Is my date-only `due_date` reduction still correct given #250's timezone fix? **Version/release correctness** of the `0.18.0b3` bump against the published tags. And whether the auto-merged locale files and `panel.ts` came through coherently.

---

prestomation:
The design review landed: sync moved out of its own Settings card and **into each Profile** as a collapsible *Sync to a to-do list* group. Commits `f836340e` (refactor) and `7645b884` (verification + captures) complete it. The `task_mirrors` option is deleted; a profile now carries `sync: {entity_id, two_way, vanish_as_completed}`, clearing the picker is both the off switch and the delete, and there is no Delete button in the group. Bookkeeping keys became `mirror_key(profile_id, task_id)`. A profile stored before this reads back sync-off, so nothing migrates.

Gates: 1444 unit, 165 integration, 70 e2e, 682 frontend, mutation 98.96% Python / 100% Stryker, mypy + ruff + vale clean. Screenshots re-captured and the walkthrough re-pointed through Profiles.

/q review Please review the reshaped design critically and skeptically, most serious first. **The refactor's blast radius:** `desired_by_mirror`/`plan_sync`/`needs_pass` lost their separate mirror-config argument and the "profile can't be resolved" branches were deleted as unreachable — is that reachability argument actually airtight, and did removing them leave any state (a profile deleted mid-pass, sync cleared while items are tracked, two profiles pointed at one list) without a home? **Bookkeeping identity:** keys moved from a standalone mirror id to the profile id; consider entries written before the change, a profile deleted and recreated, and whether anything can now collide or orphan. **The migration claim** that a pre-existing profile reads back sync-off, including via the options flow and `set_options`. **Panel state:** the group's open/closed rule layers explicit expand over explicit collapse over a data-derived default, keyed per profile — any way that strands a user, and does the shared debounce between the filter form and the sync form risk one overwriting the other? And **anything in the docs or CHANGELOG that still describes the old model.**

---

prestomation:
The sync UI has been reworked since the last review round. It was a separate Settings card holding mirror records, each pairing a **profile dropdown** with a list; it is now a collapsible **Sync to a to-do list** group inside each Profile's own editor, with no Delete button — clearing the entity picker is the off switch and the delete. The `task_mirrors` option is deleted outright: a profile carries `sync: {entity_id, two_way, vanish_as_completed}`, bookkeeping keys become `mirror_key(profile_id, task_id)`, and the "mirror names a profile that no longer exists" state is gone by construction. A profile saved before this reads back sync-off, so nothing migrates.

Commits `f836340e` (the refactor), `7645b884` (verification + captures re-targeted), `d8bbd01f` (prose lint).

/q review Please review the reworked design critically and skeptically, most serious first. **Correctness of the collapsed model** — the planner and driver now take profiles directly, and the backend deliberately passes *all* profiles (not only synced ones) to `plan_sync`/`needs_pass` so "profile deleted" stays distinguishable from "sync switched off"; is that distinction actually preserved on every path, and does clearing the picker reliably remove the lines that profile put on the list? **Dead-path removal** — two "misconfigured mirror" hold-branches and phase two's empty-target guard were deleted as unreachable; prove them unreachable or find the input that reaches one. **Frontend state** — the sync group's open/closed rule layers an explicit-expand set over an explicit-collapse set over a data-derived default, and both `ha-form`s in a profile share one debounce key with each holding the other half in a closure; look for a save that drops a concurrent edit, and for the cleared-picker value actually persisting as `""`. **Migration and storage** — profiles written by the previous shape of this same PR carry a stale `sync`-less dict; `mirror_key` now derives from profile ids, so any bookkeeping written earlier is orphaned. **Regressions against `main`** — #248/#251 (problem sensors in Profiles, renamed tiers) and #250 (to-do due dates in local time) all landed under this branch. Don't withhold minor findings.

---

prestomation:
/q review This branch was just rebased onto `main` at 640a7b7 (force-push, 19 commits). Please be critical and skeptical, and focus on what the rebase could have broken rather than re-reviewing the feature from scratch. Surface the most serious issues first, but don't withhold minor ones.

Specifically:

- **Conflict resolutions.** `tests/conftest.py` and `pyproject.toml` both had `main` adding `resolve` and this branch adding `task_mirror` to the same list; I kept both. In `conftest.py`'s `_PURE_MODULES` the ordering is load-bearing (a module first pulled in by a sibling and then re-executed leaves two copies loaded) — is `task_mirror` correctly placed after the siblings it imports and before `options`, and does adding `resolve` ahead of it change anything?
- **Semantic conflicts the rebase can't see.** `main` landed `api_surface.py` (#260) and name-accepting service fields via `resolve.py` (#263) *after* this branch forked. This branch adds no new services or websocket commands, and its sync config nests inside the existing `profiles` option — but please check whether anything here should have been declared in `api_surface.py`, or should be routed through `resolve.py`.
- **The `__init__.py` / `coordinator.py` / `notifier.py` auto-merges.** These merged without conflict but combine two features' wiring. Is `TaskMirrorSync`'s setup/teardown correct next to `main`'s changes — listeners, `async_on_unload`, sweep cadence, and the `_effective_filter_tasks` → `effective_filter_tasks` visibility change?
- **Versioning.** `0.18.0b3` shipped from `main` while this PR was open, so I moved the feature into a new `## [0.18.0b4]` section and bumped `manifest.json` + `const.py`. Does that satisfy `changelog-release-gap`, and is `main`'s `0.18.0b3` section left correct?
- **Correctness / HA best practices / performance** in `task_mirror.py` and `task_mirror_sync.py` generally: error paths and best-effort `todo.*` calls, re-entrancy, timezone/DST handling on due dates, and whether the event-gated `needs_pass` check can miss a drift case.

---

PR #249, opened 2026-08-26
fix(panel): keep the sidebar panel up across a config-entry reload (Fixes #247)

### LINKED ISSUE #215: Tree view
I can setup parent/child relationships between devices, but I have no way to visualize it. An option to toggle the device list from flat list to a tree view would be nice.

### LINKED ISSUE #247: Home Keeper is crashing after updating to v0.16.0
Since the update to v0.16.0, the UI is crashing a lot and kicking me back out to my default dashboard. At first I thought it was only happening when pressing "Add Notification" but now it seems to just be random. Seemingly can't make it more than 10 seconds or so. Seeing this in logs, not sure if related:

https://pastebin.com/RejcBVXj

prestomation:
Fixes #247

### The bounce (the reported crash)

`async_unload_entry` called `frontend.async_remove_panel`, so a config-entry
**reload** deleted `home-keeper` from `hass.panels` and put it back once setup
finished. Home Assistant's `partial-panel-resolver` answers a panel disappearing
out from under the page by moving the viewer to the default panel — so a reload
throws whoever is reading the panel back to their dashboard.

That explains both halves of the report. The reporter's first trigger, **Add
notification**, saves through `options.async_set_options`, which reloads the entry.
It then "just went random" because reloads are routine here: a synced problem sensor
appearing, an auto-buy reminder created or retired, a purged one-off, a language
change.

Nothing about the registration is entry-scoped: it names a static module URL that
`async_register_static_paths` serves for the whole HA run, and setup re-registers it
identically. So the panel is now dropped in `async_remove_entry`, plus on unload when
`entry.disabled_by` is set — HA sets that *before* unloading, and a disabled entry is
the one unload that isn't coming back. Services still go on the last loaded entry's
unload; re-registering those is invisible. `card.py` already took this stance for the
card's Lovelace resource, for the same reason.

### The blocking-call warnings (the pasted log)

The log in the issue is HA's blocking-call detector catching `backend_i18n` opening
its string tables straight from the loop thread. That module deliberately has no Home
Assistant import — the pure modules (`problem_tasks`, `inventory`,
`companions_catalog`) call it — so it cannot dispatch its own reads the way
`notifier.py` does for #150, and its `functools.cache` only helps *after* the first
read. That first read lands on whichever loop-bound caller gets there first.

`async_setup_entry` now warms both tables through `hass.async_add_executor_job`
before it touches the store, so every later `resolve_*` is a cache hit. `preload`
always warms English alongside the configured language, because that is the fallback
both resolvers reach for.

Those warnings are **not** the reported crash, and the reporter's own "not sure if
related" was right. Their log shows 4 occurrences across three hours — the cache fills
once per process, so it cannot produce a symptom recurring every ten seconds. Fixed
because it is a real Platinum-scale violation that arrived in the same report.

### Tests (all failing before the change)

| Test | Asserts |
| --- | --- |
| `tests/integration/test_panel_stability.py` | A reload — and an options save, the reporter's own trigger — produces **no** `panels_updated` events at all. Two before the fix (remove, then re-add). |
| `tests/integration/test_event_loop_blocking.py` | HA's own blocking-call detector, read back through `/api/error_log`, names nothing of ours. Provokes both string tables first so it can't pass by doing nothing. |
| `tests/unit/test_backend_i18n_preload.py` | `preload` warms *every* cached table in the module, not just the two that exist today; resolving after it reads no files. |

Verified in a real browser too, and it needs no help to reproduce. Sitting on
`/home-keeper` and reloading the config entry sent Chromium to `/home/overview` on
**3 of 3 runs** against `main`, and left it on `/home-keeper` on **3 of 3** with this
branch. The removal and re-registration are only ~19 ms apart, so the window is far
tighter than it needs to be — the frontend loses that race every time, which is why
the reporter could not get through ten seconds of use.

### Release

Cuts **0.17.0 stable**, rolling the `0.17.0b1` tree view up into `### Added` as
AGENTS.md prescribes for a stable section. That is deliberate rather than another
beta: #247 is a regression every 0.16.0 user is hitting now, and this repo releases
off a single `main` line, so a `0.16.1` patch off the stable tag isn't a shape the
release process supports. `ci/release-issues.py --version 0.17.0` resolves #215 and
#247, so `notify-issues` closes both.

### Docs

`.amazonq/rules/architecture-and-code.md` gains the two conventions this
established: frontend registrations outlive the config entry and are never torn down
on an ordinary unload, and `backend_i18n`'s tables are warmed in the executor at
setup.

## Checklist

- [x] Tests run locally — unit 1291 passed, frontend 647 passed, integration 156
      passed, e2e 65 passed, `ruff check`/`format` clean, `vale` clean on the new
      CHANGELOG lines
- [x] `CHANGELOG.md` updated, with `(Fixes #247)` in the bullet's first paragraph
- [x] Panel UI unchanged (backend only) — no screenshots needed
- [x] No new user-facing UI surface — walkthrough untouched
- [x] Not a new feature — no beta bump or `preview-release` label; version goes
      straight to `0.17.0` stable as discussed above

> [!NOTE]
> `mypy` was not run locally: it needs Home Assistant installed, and HA's Python
> floor is above what this session's interpreter offers (3.11), so a local install
> would silently resolve a stale HA — exactly what `ci/check-ha-version.py` exists to
> catch. `lint.yml` covers it, and passed.

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -455,6 +455,26 @@ variables (`var(--primary-color)`, `var(--divider-color)`) — see the upload pr
 bar (`.hk-upload`). HA has broken us this way before (issue #144, `ha-dialog`'s slot
 API).
 
+### Frontend registrations outlive the config entry — never tear one down on unload
+The sidebar panel and the card's Lovelace resource are registered against the *HA
+run*, not the entry: both name a static module URL that
+`async_register_static_paths` serves until restart, and setup re-registers each one
+identically. An ordinary unload is almost always the first half of a **reload**, and
+reloads are routine here — saving options, a synced problem sensor appearing, a
+purged one-off, a language change. Dropping the registration in
+`async_unload_entry` therefore deletes it for as long as setup takes, on a schedule
+the user never asked for.
+
+For the panel that is user-visible damage, not churn: HA's `partial-panel-resolver`
+answers `home-keeper` disappearing out of `hass.panels` by navigating to the default
+panel, so a reload throws anyone reading the panel back to their dashboard. It reads
+as random because it is a race with the frontend's `get_panels` refetch — a fast
+reload usually wins, a slow one never does (#247). Tear both down in
+`async_remove_entry` instead, plus the panel when `entry.disabled_by` is set (HA
+sets it *before* unloading, and a disabled entry is the one unload that isn't coming
+back). Services are different: re-registering them is invisible, so they still go on
+the last loaded entry's unload.
+
 ### A dashboard asset ships as a Lovelace resource, not just an extra module URL
 `frontend.add_extra_js_url` reaches the browser exactly one way: `IndexView` renders an
 inline `import("<url>")` per extra module into the app-shell HTML — a response with no
@@ -798,6 +818,20 @@ The appliance/asset feature lives in `assets.py` (pure model — no HA imports, 
   `hass` itself — the HA-aware caller (`store.py`, `websocket_api.py`,
   `companions.py`) threads `hass.config.language` in, the same pattern
   `store.reconcile_part_tasks` already established for wear-part task names.
+- **Warm the tables in the executor before anything asks for a string.**
+  `backend_i18n` has no HA import by design, so it cannot dispatch its own reads the
+  way `notifier.py` does (#150) — and its `functools.cache` only helps *after* the
+  first read, which lands on whichever loop-bound caller gets there first (the
+  problem-sensor reconcile during setup, a websocket error reply, a CSV export). HA's
+  blocking-call detector logs every one of those as `Detected blocking call to
+  read_text ... inside the event loop` (#247). `async_setup_entry` runs
+  `backend_i18n.preload` through `hass.async_add_executor_job` before it touches the
+  store, so every later `resolve_*` is a cache hit; `preload` always warms English
+  alongside the requested language, because that is the fallback both resolvers
+  reach for. Add a table to this module and you add it to `preload` —
+  `tests/unit/test_backend_i18n_preload.py` checks *every* cached table, and
+  `tests/integration/test_event_loop_blocking.py` reads HA's own log back to catch
+  anything the unit lane can't see.
 
 ## Companion discovery (implemented)
 - Home Keeper surfaces integrations that work with it in the panel's **Settings →
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
 - **Home Keeper no longer throws you back to your dashboard.** Anything that reloaded
   the integration took the sidebar panel down until setup finished, and Home Assistant
-  navigates away from a panel that vanishes under it. The panel now survives a reload.
+  drops you out of a panel that vanishes under it. The panel now survives a reload.
   (Fixes #247)
 - **"Detected blocking call" warnings from Home Keeper are gone.** The integration read
   its backend string tables on Home Assistant's event loop the first time it needed a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.17.0] - 2026-08-26
+
+### Added
+
+- **Tree view for the appliance list.** A View toggle on the Appliances tab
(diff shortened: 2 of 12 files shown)


prestomation:
/q review Be critical and skeptical, and put the most serious findings first — but don't withhold the minor ones. This PR changes config-entry lifecycle and adds an executor hop during setup, so scrutinise:

**Correctness / lifecycle.** `async_unload_entry` no longer unregisters the sidebar panel except when `entry.disabled_by` is set; `async_remove_entry` does it instead. Is `entry.disabled_by` reliably set *before* `async_unload_entry` runs on the disable path across supported HA versions? What about the unload paths I haven't enumerated: setup failing after a reload's unload (entry ends `SETUP_ERROR` with the panel still registered), HA shutdown, a second config entry, `async_remove` on an entry that was already disabled (does the panel get removed twice — is `async_unregister_panel` safe to call when the panel is gone)? Is leaving the panel up while the entry is unloaded actually safe for the frontend — do the websocket commands degrade cleanly to `not_loaded`, and does the static path outlive the entry as `panel.py` claims?

**Correctness / the executor preload.** `backend_i18n.preload` is called once per `async_setup_entry` with `hass.config.language`. Are there callers that resolve a string for some *other* language, or before `async_setup_entry` runs (a global `async_setup`, an HTTP view registered earlier, `companions.async_reconcile` from the coordinator's first refresh)? Does `functools.cache` across a config-entry reload or a language change do the right thing, and is the `dict.fromkeys((lang, _DEFAULT_LANG))` dedupe correct when `lang == "en"`?

**HA best practices.** Is `hass.async_add_executor_job` at the very top of `async_setup_entry` the right place, or should this be a `singleton`/`async_get_loaded_integration`-style component-level warm-up? Is registering a panel that outlives its config entry an accepted HA pattern or an abuse of `frontend.async_register_built_in_panel`?

**Test quality.** Do the three new tests actually pin the behaviour, or can they pass vacuously? `test_panel_stability.py` asserts *zero* `panels_updated` events — is that over-tight (could unrelated activity in a real install fire one, making it flaky in CI)? `test_event_loop_blocking.py` reads `/api/error_log` and greps for `home_keeper` — is that robust, and could it be polluted by other tests sharing the container? `test_backend_i18n_preload.py` introspects `vars(module)` for cached tables — will that silently find zero tables and pass if the implementation changes shape?

**Maintainability / performance.** Is the added executor hop on the setup path worth it versus just moving each read, and does it measurably delay setup? Any duplication between the new integration tests and the existing `test_card_resource.py` reload test that should be shared?


---

prestomation:
/q review Pushed `1afdc5c`, a one-line CHANGELOG wording change (`navigates` → `drops`) to clear the `ai-tells.OverusedVocabularyVerbs` vale gate. No code changed since your last pass.

Be critical and skeptical, most serious first, and don't withhold nits. Scoped to what this push touches: does the reworded bullet still accurately describe the fix — Home Assistant moving the viewer off a panel that leaves `hass.panels` mid-reload — or has "drops you out of" lost or overstated something the original said? Does it still read as a user-facing changelog line rather than an implementation note, and does it hold to the project's three-sentence budget with `(Fixes #247)` in the first paragraph? Also worth a skeptical look: the `## [0.17.0]` section rolls `0.17.0b1`'s tree-view bullet up into `### Added` while the `## [0.17.0b1]` section below keeps its own copy — is that the right shape for a stable cut here, and is anything shipped since `v0.16.0` missing from the section that `notify-issues` will read?


---

prestomation:
/q review Pushed `0ddbce8` — a comment-only addition to `async_unload_entry`, no behaviour change. Full CI was green on the previous head (`1afdc5c`): integration, e2e, mypy, hacs, walkthrough, test and the rest.

Be critical and skeptical, most serious first, nits included. This push exists to correct one claim from your first pass, so please re-derive it rather than agreeing with either of us. You wrote that a setup failure after a reload's unload "leaves no panel to clean up" because registration happens late in `async_setup_entry`. I read that as true only for the *first* setup: on a reload the panel is already registered from the previous successful setup, and it now survives the unload, so a failed re-setup leaves the sidebar entry up against an entry in `SETUP_ERROR`/`SETUP_RETRY`. Which of us is right? Trace `async_reload` → `async_unload` → `async_setup` and say plainly if the new comment is wrong.

If the comment is right, is the trade-off it argues for actually the better one — a sidebar entry whose commands all answer `integration_not_loaded`, versus dropping the entry and hiding that Home Keeper is installed? Consider what a user in a permanent `SETUP_ERROR` sees, and whether a panel that renders its shell but fails every command is worse than no sidebar entry at all.

One related edge I chose not to widen the PR for — tell me if you think it deserves a fix rather than silence: `_not_loaded` resolves its message through `backend_i18n`, so a websocket command reaching a never-successfully-set-up integration would do the one blocking read `preload` exists to prevent. Reachable only by a direct websocket client (the panel isn't registered until a setup succeeds), and `functools.cache` is module-level so any prior successful setup keeps it warm. Real enough to fix, or correctly left alone?


---

PR #251, opened 2026-08-26
fix(profiles): show synced problem sensors, let Snooze work on them, and name the status tiers

### LINKED ISSUE #248: In profiles - no way to select multiple filters
Image

Why can't I show both "overdue" and "due soon"? If I select "due soon", some problems synced from tasks don't show up.

prestomation:
Fixes #248

## What changed

Three things, all from #248.

### 1. Profiles dropped synced problem sensors outright — the reported bug

`profiles.matches_filter` and its TypeScript twin `card-filter.profileMatches` both returned early on a `problem_sensor` source, **under every status**. So picking any Profile in the panel or on a card made every mirrored problem vanish — the reporter's second comment. The plain task list, the `todo` entity and an unfiltered card all showed them, so the profile matcher was the only place hiding them.

An armed mirror carries `next_due` = the moment its sensor went bad, so it is real overdue work and now matches like any other task. A dormant one (sensor back to OK) is undated and still excluded by the rule above.

### 2. Snooze now works on a synced problem task

The first pass excluded these from *walk* notifications, because `complete`, `snooze` and `skip` were all rejected by the store, and a walk advances only on a successful action — one at the head of the queue would re-send forever. On review that was working around the problem rather than fixing it.

`store.snooze_task` no longer rejects a problem-sensor mirror. Splitting the guard that way is the point: **complete and skip both assert the problem is dealt with**, which only the originating integration can decide, so they stay rejected. **Snooze asserts nothing of the sort** — it defers the reminder and leaves the problem standing. It also survives the sync: `reconcile_problem_tasks` reads armed as `next_due is not None`, so a snoozed mirror stays armed while its sensor is bad and still auto-clears when the sensor returns to OK.

With a working verb, the delivery-layer exclusion goes away. `notifications.actions_for` replaces it: for a `completion_blocked` task it drops complete and skip, and offers Snooze **even when the notification's own button set omits it** — without that one button a walk could still park forever. The old behaviour offered buttons the store refuses, and `notifier` swallows the rejection, so they read as dead buttons. Now only buttons that work are shown.

Keyed on `managed_by.completion_blocked`, the marker every surface already uses to hide *Done* — not the `problem_sensor` source — so a future completion-blocked source is covered without another edit.

### 3. "No way to select multiple filters" — the status values are nested tiers

`due_soon` has always meant "overdue **or** due within 3 days", so "Overdue + Due soon" is exactly what "Due soon" already did. A single select labelled *All due / Overdue / Due soon* hides that completely. The options now name what they cover, with a helper under the field:

| before | after |
| --- | --- |
| Overdue | **Overdue only** |
| Due soon | **Overdue and due soon** |
| All due | **Every scheduled task** |

No storage change and no migration — the stored `status` values are untouched. All 16 locales updated, plus a helper under the problem-sensor sync toggle in both the panel's Settings tab and the options flow.

## Screenshots

The Tasks tab filtered by a saved Profile — **"Sump pump problem"** is the synced task that used to vanish here:

The Tasks tab filtered to the My chores Profile, listing Sump pump problem with an Auto-synced chip and a Clears automatically caption alongside the other overdue tasks

Settings → Profiles, with the relabelled Include tiers and the new helper:

Two expanded profile editors showing Include set to Overdue only and Overdue and due soon, each with the helper Each choice includes the one above it

Settings → Problem sensor sync, with the helper under the toggle:

The Home Keeper Settings tab; under Sync problem sensors as tasks a helper explains that a synced task clears only when the source integration resolves the problem and that reminders offer Snooze in place of Mark done

## Verification

| Gate | Result |
| --- | --- |
| `pytest tests/unit` | 1302 passed, 2 skipped |
| `bash ci/test-frontend.sh` | 649 passed |
| `bash ci/test-python-integration.sh` | 158 passed on a clean seed |
| `bash ci/e2e-up.sh` | 67 passed (headless shell, the browser CI uses) |
| `bash ci/test-mutation-python.sh` | **100%** — 165/165 |
| `bash ci/test-mutation-frontend.sh` | **100%** — 4/4 on `card-filter.ts` |
| `tsc --noEmit` | clean |
| `vale` (diff-scoped) | clean |

New tests: armed/dormant problem sensors in `test_profiles.py` and both frontend matcher suites; two conformance cases in the shared fixture; `actions_for`/`is_completion_blocked` branch coverage plus a `build_notification` case asserting the rendered verbs; integration coverage that snooze is accepted and sticks past a reconcile while skip is still refused, and that a walk now carries the mirror; an e2e regression that picks a Profile and asserts the Sump pump card survives it.

Editing `build_notification` pulled it into the mutation scope and exposed two gaps that predate this branch — nothing asserted the `"Home Keeper"` fallback title for a nameless task, nor the exact `data.group` the companion app stacks notifications by. Both are payload contract a user sees on their phone, so they got real assertions rather than a pass on the 80% threshold.

Two notes on checks I could not run here. The full `mypy` job needs Home Assistant installed, which does not resolve on this container's Python 3.11 (HA's floor is 3.14) — the changed pure modules pass `mypy --strict` on their own, and CI runs the real job. And a local `vale` run reported clean while the rule regexes, checked directly per AGENTS.md, found three real `VerbTricolon` hits in new README prose; those were fixed and both checks are now clean.

## Checklist

- [x] `CHANGELOG.md` updated — `(Fixes #248)` in the `### Fixed` bullet's first paragraph
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with HTML image tags
- [ ] New user-facing UI surface → not applicable: no new surface, so the walkthrough tour is unchanged. Snooze on these tasks is a notification button, which the tour does not cover
- [x] Version bumped to `0.18.0b1` (`manifest.json` + `const.py`) — required here regardless, since `v0.17.0` is already tagged and `changelog-release-gap` fails an edit to a released section without a bump


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones either. This changes who a saved Profile selects, so I'd like the risky parts attacked specifically:

**Correctness of the behaviour change.** `profiles.matches_filter` and `card-filter.profileMatches` no longer drop `problem_sensor`-sourced tasks. Is there any state where an armed mirror now leaks somewhere it shouldn't, or where the two matchers can disagree? They are pinned only by `tests/fixtures/profile_filter_cases.json` — is that fixture now under-specifying the problem-sensor cases? Consider a task that is armed but whose sensor is `unavailable`, a mirror mid-reconcile, and a task carrying `managed_by` but no `source`.

**The walk exclusion.** `notifier._send` narrows a walk queue with `notifications.walk_queue`, keyed on `managed_by.completion_blocked`. Attack the placement: it sits *above* the `if not queue` early return, so a walk matching only blocked tasks reports `matched: 0` and the walk-advance path emits "all caught up" rather than stalling — is that the right reading of the advance logic in `async_setup_notifications._handle`, or does some path still resurface a blocked task? Is keying on `completion_blocked` rather than the `problem_sensor` source over-broad — does it now exclude anything from a walk that a user would reasonably expect to be walked through? And is a `matched` count that differs between a digest and a walk over the *same* Profile defensible, or a trap for anyone reading the `home_keeper.notify` service response?

**Maintainability / boundaries.** Is "filter says who, delivery says how" the right seam, or should this have been an explicit per-profile opt-in field instead? Is `is_walkable`/`walk_queue` in the right module (`notifications.py` owns `STYLE_WALK`; `profiles.py` owns `due_queue`)?

**HA best practices and performance.** `_settingsCard` gained a generic `computeHelper` that probes `settings.{field}_help` per field and compares against the returned key to detect a miss — is that idiomatic for `ha-form`, and is the sentinel comparison safe? Any concern about the extra per-render `t()` lookups, or about `walk_queue` re-allocating the queue on every send?

**Anything about the copy.** The status options were relabelled across 16 locales; the tiers are cumulative, so flag any wording that still reads as mutually exclusive.

---

prestomation:
/q review Second push — the design changed since your first pass, so please re-review with fresh eyes and be critical and skeptical. Most serious issues first, and don't withhold the minor ones.

**What changed.** The previous version excluded `completion_blocked` tasks from *walk* notifications, because the store rejected complete/snooze/skip and a walk advances only on a successful action. That has been replaced: `store.snooze_task` no longer calls `_reject_synced_problem` (complete and skip still do), and `notifications.actions_for` now trims a blocked task's button set to Snooze + Open, forcing Snooze in even when the notification's configured `actions` omits it. `notifications.walk_queue`/`is_walkable` are gone.

Please attack specifically:

**The relaxed store guard.** Is snooze genuinely safe to allow on a problem-sensor mirror, or does it break an invariant `_reject_synced_problem` was protecting? I claim it survives `reconcile_problem_tasks` because that reads armed as `next_due is not None`, so a snoozed mirror stays armed while the sensor is bad and still auto-clears on a `problem is False` pass. Check that reasoning against the reconciler's three-way `is_problem` handling — especially the indeterminate (`None`) branch, and a snooze that lapses while the sensor is still bad. Does a snoozed mirror do anything odd to the edge-triggered overdue/due-soon transitions, the per-task entities, or the to-do list?

**Forcing a button the user didn't configure.** `actions_for` inserts Snooze into a blocked task's button set even when the notification's `actions` excludes it. Is that defensible, or should such a task instead be dropped from that particular walk? Is there any configuration where the forced button still can't advance the walk?

**Did I actually fix the stall, or move it?** Trace the walk-advance path in `notifier.async_setup_notifications._handle` and confirm a Snooze tap on a blocked task reaches the advance block rather than the `except (KeyError, TaskValidationError)` return.

**Privilege model.** `docs/SECURITY.md` and the rules describe these tasks as externally-owned and un-completable. Does allowing snooze weaken the boundary, and is the service/websocket parity still right (the panel and card have no snooze affordance today — should they)?

**Everything from the first review still applies** — matcher parity between `profiles.matches_filter` and `card-filter.profileMatches`, the shared conformance fixture, HA best practices, performance, and the relabelled status copy across 16 locales.

---

PR #252, opened 2026-08-27
fix(todo): date to-do items in Home Assistant's timezone

### LINKED ISSUE #250: Todo item due date is one day earlier than the panel and next_due
**Version:** v0.16.0b9
**HA Core:** 2026.8.3
**Timezone:** Europe/London (currently BST, UTC+1)

### Summary

`todo.home_keeper_tasks` renders an item's `due` date by calling `.date()` on the parsed `next_due` without converting to Home Assistant's local timezone. When a task's `next_due` is stored with a `+00:00` offset and falls at 23:00Z, the todo item shows the previous day. The panel and `home_keeper.list_tasks` show the correct day, so the two surfaces disagree by one.

### Reproduction

Floating task, 1 week interval, last completed via the panel's date picker.

`home_keeper.list_tasks`:

```
name:            Upstairs Vacuum Clean
recurrence_type: floating
interval:        1 week
last_completed:  2026-08-19T23:00:00+00:00
next_due:        2026-08-26T23:00:00+00:00
```

`2026-08-26T23:00:00+00:00` is **Thursday 27 August 00:00 BST**. The panel shows the 27th.

`todo.get_items` on the same task, at the same moment:

```
summary: Upstairs Vacuum Clean
uid:     2373f084-1d9d-4912-b27b-ccace31af455
due:     2026-08-26
```

One day early.

Two other tasks in the same instance show the same shift: `next_due 2026-08-29T23:00:00+00:00` renders as `2026-08-29` (locally the 30th), and `next_due 2026-10-15T23:00:00+00:00` renders as `2026-10-15` (locally the 16th).

### Root cause

`custom_components/home_keeper/todo.py`, in `todo_items`:

```python
due = dt_util.parse_datetime(due_iso) if due_iso else None
...
due=due.date() if due else None
```

`parse_datetime` returns an aware datetime carrying whatever offset the stored string has. `.date()` then takes the date **in that offset**, not in HA's local timezone.

Tasks whose `next_due` carries a local offset (`+01:00` here) come out correct by coincidence — the stored offset happens to equal local. Only `+00:00` values are wrong, and only while local time differs from UTC.

### Why some tasks store +00:00 and others don't

Completions made "now" are stored with the local offset (`2026-08-23T09:10:51.249933+01:00`) and render correctly. Completions entered through the panel's date picker are stored as local midnight expressed in UTC (`2026-08-19T23:00:00+00:00`), and those are the ones that shift.

So the bug only affects tasks whose last completion was set by date rather than by completing them live.

### Suggested fix

```python
due=dt_util.as_local(due).date() if due else None
```

There may also be a case for normalising the offset at write time in the date-picker completion path rather than only at render — your call which layer is the right one to fix.

### Notes

This is invisible in any UTC±0 timezone, and invisible everywhere during northern-hemisphere winter, which may be why it hasn't surfaced before.

It matters because anything consuming the todo entity's `due` field — a notification digest, an automation, a dashboard card — inherits the shift silently. The value is a plausible date, just the wrong one.

prestomation:
Fixes #250

## What changed

`todo_items` called `.date()` on the parsed `next_due`. `dt_util.parse_datetime` returns an aware datetime carrying **whatever offset the stored string has**, so `.date()` took the calendar date *in that offset* rather than in Home Assistant's zone.

The reporter's numbers, on Europe/London in BST: `next_due` of `2026-08-26T23:00:00+00:00` is Thursday 27 August 00:00 local. The panel and `list_tasks` said the 27th; `todo.home_keeper_tasks` said the 26th.

Why only some tasks: a completion made *now* is stored with the local offset and happens to project correctly. One entered through the panel's date picker goes through `haDateTimeToIso` → `new Date(...).toISOString()`, which is the right instant expressed in UTC — and those are the rows that shift. Invisible at UTC±0, and invisible everywhere during northern-hemisphere winter.

**A second instance of the same bug, found while tracing it.** `store._stamp_part_replacement` did the same thing to a completion timestamp, and that value becomes a wear part's `last_replaced`, which `reconcile` re-anchors the part's recurrence to. A back-dated part replacement therefore shifted the whole wear cycle a day, not just a label. Same root cause, same trigger, same one-line fix.

Both now convert with `dt_util.as_local` before projecting to a date.

### Why render time and not write time

The issue offers both layers. The stored instant was always correct — only the projection to a calendar date was wrong — and converting on read also repairs tasks already on disk, which normalizing at write time cannot. A write-time change would also be undone by any automation calling `complete_task` with a `Z` timestamp.

### Checked and not affected

`calendar.py` (emits aware datetimes, HA localizes them), the `next_due` timestamp sensor (aware, `device_class: timestamp`), `notifications.py` (day arithmetic on deltas, never a date projection), and the panel (`toLocaleDateString()`). `assets.py` and the metadata date sensor slice date-only strings, where there is no offset to get wrong.

### Tests

- `tests/unit/test_todo.py` — the reporter's case (UTC-stored due under UTC+1 → next day) and its mirror (UTC+1-stored due under UTC-4 → previous day), so the conversion cannot pass by doing nothing, plus the already-matching-offset case that was always right. `_local_tz` pins "local" by swapping the `dt_util` reference on the `hk.todo` module; see the second commit for why it must not touch Home Assistant's own global.
- `tests/integration/test_lifecycle.py` — the container runs on America/New_York, so this pins the date Home Assistant's own `todo.get_items` hands to a card or an automation. A unit test that stubs `dt_util` cannot see that contract.
- `tests/integration/test_assets.py` — a back-dated wear-part completion has to stamp `last_replaced` with the local date.

All four fail on the parent commit with exactly the reported one-day shift, and pass on this one.

### A note on the second commit

The first push was red: *1313 passed, 3 errors*. The new tests patched `dt_util.DEFAULT_TIME_ZONE`, and `pytest-homeassistant-custom-component`'s `verify_cleanup` fixture asserts that global is still UTC at teardown — it tears down before monkeypatch's undo, so every test that set it errored. It surfaced only in the lane that installs real Home Assistant; the hand-built stubs in the unit file kept a local run green.

`97d4527` swaps the module's own `dt_util` reference instead, so nothing reaches into Home Assistant's state and the suite behaves the same either way. The failure was reproduced locally first with a fixture mimicking `verify_cleanup`'s teardown ordering, and both directional tests still fail on the unfixed entity.

### Release

`v0.18.0b1` is already published, so this opens `0.18.0b2` (`manifest.json` + `const.py` + a new CHANGELOG section) rather than editing a shipped section — `changelog-release-gap` would otherwise fail, and the entry would document a fix that never ships.

## Checklist

- [x] Tests run locally — 1306 unit + 160 integration, all green
- [x] `CHANGELOG.md` updated with `(Fixes #250)`
- [x] Panel UI changed → n/a, no change under `frontend/src/`
- [x] New user-facing feature → n/a (bug fix); version still bumped, because the top CHANGELOG section was already released

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.18.0b2]
+
+### Fixed
+
+- **To-do items show the right due date outside UTC.** A task due at local midnight showed
+  the previous day on `todo.home_keeper_tasks` while the panel showed the correct one. A
+  wear part's last-replaced date shifted the same way when the replacement was back-dated.
+  (Fixes #250)
+
 ## [0.18.0b1]
 
 ### Fixed
--- a/custom_components/home_keeper/const.py
+++ b/custom_components/home_keeper/const.py
@@ -8,7 +8,7 @@ PLATFORMS = ["todo", "calendar", "button", "sensor", "binary_sensor", "number"]
 # Frontend panel.
 # PANEL_VERSION is the single source of truth that release.yml validates against
 # manifest.json's "version" (mirrors Pawsistant's CARD_VERSION check).
-PANEL_VERSION = "0.18.0b1"
+PANEL_VERSION = "0.18.0b2"
 PANEL_URL_PATH = "home-keeper"  # sidebar route -> /home-keeper
 PANEL_STATIC_URL = "/home_keeper_panel"  # static path that serves the JS bundle
 PANEL_JS_FILENAME = "home-keeper-panel.js"
--- a/custom_components/home_keeper/manifest.json
+++ b/custom_components/home_keeper/manifest.json
@@ -21,5 +21,5 @@
   "requirements": [
     "Babel>=2.12"
   ],
-  "version": "0.18.0b1"
+  "version": "0.18.0b2"
 }
--- a/custom_components/home_keeper/store.py
+++ b/custom_components/home_keeper/store.py
@@ -1536,7 +1536,17 @@ class HomeKeeperStore:
         asset = self._assets.get(src["asset_id"])
         if not asset:
             return
-        when_date = when.date().isoformat() if hasattr(when, "date") else str(when)[:10]
+        # ``as_local`` first (#250): *when* carries the offset the caller supplied — UTC
+        # for a completion back-dated through the panel's date picker — and a bare
+        # ``.date()`` would take the calendar date in that offset. ``last_replaced`` is
+        # a date-only string that ``reconcile`` re-anchors the part's recurrence to, so
+        # a one-day shift here shifts the whole wear cycle. The ``hasattr`` guard keeps
+        # the true branch a datetime: a plain ``date`` has no ``.date()`` method.
+        when_date = (
+            dt_util.as_local(when).date().isoformat()
+            if hasattr(when, "date")
+            else str(when)[:10]
+        )
         for part in asset.get("parts", []):
             if part.get("id") == src.get("part_id"):
                 part["last_replaced"] = when_date
--- a/custom_components/home_keeper/todo.py
+++ b/custom_components/home_keeper/todo.py
@@ -83,7 +83,14 @@ class HomeKeeperTodoListEntity(
                     uid=task["id"],
                     summary=task["name"],
                     status=TodoItemStatus.NEEDS_ACTION,
-                    due=due.date() if due else None,
+                    # ``as_local`` before ``date()``, never ``due.date()`` (#250). A
+                    # stored ``next_due`` carries whatever offset it was written with —
+                    # a completion made "now" gets the local one, one entered through
+                    # the panel's date picker gets UTC — and ``.date()`` takes the
+                    # calendar date *in that offset*. A task due at local midnight then
+                    # reads as the previous day here while the panel reads it right,
+                    # and everything downstream of the entity inherits the shift.
+                    due=dt_util.as_local(due).date() if due else None,
                     description=task.get("notes") or None,
                 )
             )
--- a/tests/integration/test_assets.py
+++ b/tests/integration/test_assets.py
@@ -6,6 +6,8 @@ can attach to that virtual device (its per-task entities merge onto the same pag
 and that the asset CRUD services round-trip.
 """
 
+from datetime import UTC, datetime, timedelta
+
 from conftest import call_service, list_states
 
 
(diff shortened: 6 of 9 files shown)


prestomation:

This fixes a timezone bug where a stored `next_due` / completion timestamp was projected to a calendar date in *its own offset* instead of Home Assistant's zone. Two one-line `dt_util.as_local(...)` conversions, in `todo.py` (`todo_items`) and `store.py` (`_stamp_part_replacement`).

Topics I want scrutinized:

**Correctness / completeness of the sweep.** I claim only these two sites project a date from a stored offset. Please check that independently — is there any other place (websocket handlers, `sensor.py`, `assets.py`, `reconcile.py`, `inventory.py`, notification digests, the CSV export) where an aware datetime becomes a date, a `YYYY-MM-DD` string, or a day-granular comparison without going through `as_local` or `dt_util.now()`? I deliberately left `panel.ts`'s export-filename stamp alone as cosmetic — argue if that's wrong.

**Edge cases.** DST boundaries (a `next_due` landing inside a spring-forward gap or a fall-back repeat); a naive `next_due` reaching `todo_items` from malformed or migrated data (is `as_local` on a naive value a behaviour change from `.date()`?); a task whose `next_due` sits exactly at local midnight; and the `hasattr(when, "date")` guard in `_stamp_part_replacement` — is `as_local` safe on everything that reaches the true branch?

**Data consequences.** `last_replaced` anchors the wear-part recurrence via `reconcile`. Does changing this stamp shift `next_due` for parts replaced *before* this fix, on the next reconcile? Should there be a migration, or is leaving old rows alone the right call?

**Layer choice.** I fixed at render time rather than normalizing the offset when the panel's date picker writes the completion. My reasoning is in the PR body — push back if write-time normalization (or doing both) is actually the better design.

**Test quality.** Do the unit tests assert enough to catch a wrong-direction conversion, and is the integration coverage in the right place given the container runs on America/New_York? The `test_assets.py` case derives its dates from "yesterday in UTC" so it stays valid year-round — check that reasoning holds across a DST change.

**HA best practices.** Is `dt_util.as_local` the idiomatic call here, or should the entity be storing/returning something else entirely for `TodoItem.due`?


---

prestomation:
/q review Follow-up push — please be critical and skeptical, most serious issues first, and don't withhold minor ones.

The previous head was red: `pytest and coverage report` and `test` both failed with *1313 passed, 3 errors*. My new regression tests patched `dt_util.DEFAULT_TIME_ZONE`, and `pytest-homeassistant-custom-component`'s `verify_cleanup` fixture asserts that global is still UTC at teardown — it tears down before monkeypatch's undo, so every test that set it errored. It only showed up in the lane that installs real Home Assistant; the hand-built stubs in the unit file meant a local run stayed green.

`97d4527` swaps the `dt_util` reference on the `hk.todo` module instead, so the substitution never touches Home Assistant's state.

What I'd like checked on this commit specifically:

**Is the shim honest?** `_local_tz` replaces `todo.dt_util` with a `SimpleNamespace` carrying the real `parse_datetime` and a fixed-offset `as_local`. Does that still assert something real, or has it drifted into testing the test? I argue the unit-level contract is "the entity asks `as_local` before taking a date" and that HA resolving the configured zone belongs to the integration suite — say so if you think that's a cop-out and the assertion should live elsewhere.

**Did I leave a trap for the next person?** `_install_ha_stubs` still defines `DEFAULT_TIME_ZONE`/`as_local` on the stub `homeassistant.util.dt` for the tests that don't use `_local_tz`. Nothing patches it now, so it only ever resolves to UTC. Is keeping it right, or is a stub attribute nothing varies a liability?

**Is this pattern already broken elsewhere?** Does any other test in `tests/unit/` reach into Home Assistant module state (`DEFAULT_TIME_ZONE`, other `dt_util` globals, registries) in a way `verify_cleanup` would catch once that lane sees it — or that leaks across tests because these suites share `sys.modules` stubs?

**Ordering assumption.** I reproduced the failure with a local fixture that requests `monkeypatch` to force the teardown ordering. If you think the real `verify_cleanup` ordering differs and my repro proved the wrong thing, please say so.

The underlying `as_local` fix in `todo.py` / `store.py` is unchanged from your first pass — re-flag anything there you'd like a second look at.


---

PR #260, opened 2026-08-28
feat(docs): model the integrator-facing API surface and generate its reference

prestomation:
## What changed

The Developer Guide had no complete reference for what an integration can build on, and the pieces that came closest were hand-maintained copies of facts the code already knew:

- `docs/EVENTS.md` restated `events.py`'s payload builders field for field, with nothing keeping them equal.
- `README.md`'s Services list covered 17 of 33 services.
- Nothing checked that `services.yaml`, `strings.json` and the actual `hass.services.async_register` calls agreed — there are no `SERVICE_*` constants, only bare literals at each registration site.

**That gap already shipped a bug.** `set_task_meter` was registered (`__init__.py:1312`) and present in `services.yaml`/`strings.json`, but missing from the hand-written `_SERVICES` teardown tuple — 32 entries against 33 services — so it stayed callable against an unloaded integration until Home Assistant restarted.

### The model

`custom_components/home_keeper/api_surface.py` is a new pure module: a declarative index of every surface an integrator can touch — services, events and their payloads, device triggers, entity platforms and attributes, config-entry options, plus the internal websocket commands and HTTP views. It follows the `options.py` precedent (pure, unit-testable, consumed by the runtime, pinned by tests).

**The runtime consumes it**, so lists can't drift by construction:

- `async_unload_entry` iterates `SERVICE_NAMES` and the `_SERVICES` tuple is gone. That is the `set_task_meter` fix — as a class of bug, not a line.
- `device_trigger.py` builds `TASK_TRIGGERS` / `ASSET_TRIGGERS` from `triggers_for()`.

**It declares names and structure only.** Every string Home Assistant already localizes is resolved at generation time from `services.yaml` / `strings.json`. The one exception is an event's one-line summary, because a bus event has no Home Assistant string source.

`SURFACE_KINDS` lists the surfaces Home Keeper *doesn't* offer too — device conditions, intents, repairs, conversation, backup, media source — each with a status and a one-sentence reason. A list of what exists can't tell you what was forgotten.

### The generated reference

`ci/generate_api_docs.py` renders `website/developer/api.md` from the model plus those two string files. It runs from `npm run sync` (so from `npm start` / `npm run build` too), after `sync-docs.mjs` clears that directory. The page is gitignored with the rest of the generated tree — nothing committed, nothing hand-written, nothing that can go stale in git.

Because service labels and descriptions come from `strings.json`, the reference and Home Assistant's own service dialog read from one string and cannot disagree. `test_service_prose_comes_from_strings_json_not_services_yaml` pins that by feeding the renderer sources that deliberately disagree.

The page lands in the Developer Guide sidebar at position 2, between *Integrating* and *Glue integrations*. Its first section is the surface-coverage table, then one subsection per action with its localized label, description, admin/response badges and field table, then the event catalog with a payload table per shape, then device triggers, entities, options and localized errors.

### The gate

- `tests/unit/test_api_surface.py` (24 tests) parses the component's own source with `ast` — the technique `test_exception_translations.py` already uses — and compares it to the model: registered services, `supports_response`, admin gating, `services.yaml`/`strings.json` keys, every `const.EVENT_*`, payload spines probed against the real builders, websocket command types and `require_admin` flags, `HomeAssistantView` urls, device-trigger labels, entity platforms.
- `tests/unit/test_generate_api_docs.py` (20 tests) covers the renderer, including determinism and that every modelled surface reaches the page.
- `tests/integration/test_api_surface.py` covers what static analysis can't: that a running Home Keeper registers exactly the modelled services, and that disabling the entry removes every one of them.

Both gates were verified to bite. Deleting a `ServiceSpec` fails the unit test naming the service; re-introducing the original teardown defect fails the integration test with `still_registered_after_unload: ['set_task_meter']`.

### Documentation split

`docs/EVENTS.md` keeps what a table can't say — when each event fires, edge triggering, startup baselining, tag scans, worked automations — and loses its six catalog tables and four payload tables. `docs/INTEGRATING.md` and `README.md` gain a pointer; the tutorial is unchanged.

Also fixes a pre-existing bug found on the way: `docs/GLUE_INTEGRATIONS.md` was in `DEV_DOCS` and published at `/developer/glue-integrations`, but missing from `DOC_ROUTES`, so every link to it rewrote to a GitHub blob URL instead of the on-site page. `DOC_ROUTES` moved into the side-effect-free `doc-map.mjs` and a test now asserts every `DEV_DOCS` entry has a route.

### Verified locally

`ruff check` + `ruff format --check`; `pytest tests/unit` (1350 passed); `vitest run` (652 passed); `vale` on the changed docs (clean, and no new findings on added lines elsewhere); `npm run build` for the site with `onBrokenLinks` and `onBrokenAnchors` set to throw; the full Docker integration tier (162 passed) and the e2e tier (67 passed). The rendered page was checked in a browser against the built site.

The strict-typing gate could not be reproduced here — this environment is on Python 3.11 and Home Assistant's floor is 3.14.2, so installing it would silently resolve a stale release (the trap `ci/check-ha-version.py` exists for). The new and changed modules type-check cleanly under an equivalent standalone mypy run; CI's `mypy` job is the real check.

## Checklist

- [x] `CHANGELOG.md` updated — not needed. The one behavioural change (a service lingering after unload until restart) is user-invisible and references no issue; everything else is developer-facing. `v0.18.0b2` is already published, so touching that section would trip `changelog-release-gap` without a version bump, and `AGENTS.md` doesn't ask for a fresh beta on a bug-fix-only / developer-only PR.
- [x] Panel UI changed → not applicable, nothing under `frontend/src/` is touched.
- [x] New user-facing UI surface → not applicable, no new panel surface.
- [x] New user-facing feature → not applicable, so no beta bump and no `preview-release` label.

https://claude.ai/code/session_011QF5rfz9vFLGZfSVmUzLQs


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first without withholding minor ones.

This PR adds `api_surface.py` (a declarative model of the integration's public surfaces), makes the runtime consume it, generates the Developer Guide's API reference from it, and gates the whole thing with AST-based unit tests. Topics I'd like scrutinized:

**Correctness of the runtime change.** `async_unload_entry` now iterates `api_surface.SERVICE_NAMES` instead of a local `_SERVICES` tuple, and `device_trigger.py` builds `TASK_TRIGGERS`/`ASSET_TRIGGERS` from `triggers_for()`. Does the new import (`__init__.py` → `api_surface` → `const`, `options`) introduce any import-cycle or setup-ordering hazard? `device_trigger.py` is loaded by Home Assistant *before* the config entry is set up — is deriving its module-level maps from another module safe at that point? Is `TRIGGER_SCHEMA`'s `vol.In(set(_EVENT_BY_TYPE))` still built over the same set?

**Soundness of the gate, not just its passing.** `tests/unit/test_api_surface.py` reads the component's source with `ast`. Where can it be fooled into a false *pass*? Specifically: a service registered outside the one `hass.services.async_register(DOMAIN, "<literal>", …)` shape; a handler whose `_verify_admin` call is present textually but unreachable; a websocket command declared with a non-literal `type`; an `HomeAssistantView` whose `url` isn't a `PREFIX + '/…'` concatenation (`test_http_views_match_source` compares only the literal tail — is that too weak?). Also: `test_payload_spines_match_the_event_builders` probes the builders with empty dicts — does that actually pin key *order* and would it catch a conditionally-added key?

**The generator.** `ci/generate_api_docs.py` resolves prose from `strings.json` and structure from `services.yaml`. Is `load_surface()`'s hard-coded dependency-ordered module list fragile — what happens when `options.py` gains an import? Does `cell()` escape everything a GFM table cell needs (pipes inside code spans, backslashes, HTML)? Is `selector_summary()` missing selector shapes that `services.yaml` actually uses? Is rendering genuinely deterministic given `PAYLOAD_SPINES` is a `dict`?

**Maintainability.** Is `api_surface.py` a genuine single source of truth, or have I built a parallel list that merely can't drift silently — and where is the boundary? The websocket and HTTP tables are modelled but not rendered; is that dead weight? Are `SURFACE_KINDS` and the four-status vocabulary carrying their keep?

**Docs risk.** `docs/EVENTS.md` lost its catalog and payload tables, and canonical docs now link to the generated page by its published `https://prestomation.github.io/...` URL, which `sync-docs.mjs` rewrites to a site-relative route. Does that rewrite rule have a failure mode (a URL that shouldn't be rewritten, a PR-preview base path)? Did the edit drop any semantics a reader needed?

**HA best practices.** Is a declarative surface model an idiomatic thing for a Platinum-scale integration to ship inside `custom_components/`, or should it live outside the component?

---

prestomation:
Both CI failures on `015fc8f` are fixed in `7811ad0`, and I've acted on the review. Please be critical and skeptical again on the new commit, most serious issues first.

**`preview` (docs build) — root-caused, not patched around.** The generator executed `notifications.py` on its way to `api_surface`, and that imports Babel, which a docs runner has no reason to install. The chain existed only because `api_surface` imported `options` for two tuples of strings. I dropped that import rather than adding Babel to the docs workflow: `OPTIONS` now names its keys from `const` directly, and `test_options_match_the_options_module` pins it to `options.ALL_OPTIONS` and `FLOW_OPTIONS`. That is also a strictly better test — the previous one compared a tuple against the thing it was derived from, which asserted nothing. `test_api_surface_imports_stay_light` now holds the module to importing nothing but `const`, so this failure mode is closed rather than merely fixed, and `load_surface()`'s module list collapses to two entries (your recommendation 1, in the form that made the list unnecessary).

Reproduced the failure environment locally — a venv with PyYAML and no Babel — and confirmed the generator runs there.

**`vale`.** Three real hits in prose I added. Local Vale 3.9.1 reported none of them, which is the divergence AGENTS.md warns about, so I checked my added paragraphs against the rules' own regexes with continuation lines joined the way Vale blocks them. The README pointer was the interesting one: the `VerbTricolon` match started in pre-existing text and only completed because my sentence extended that blockquote's `[^.!?]+` tail, so I moved it into its own paragraph instead of rewording someone else's sentence.

**On your other recommendations:**

- **Populated builder probes (2)** — agreed, this was a real gap. `test_payload_spines_match_the_event_builders` now probes each builder twice, once with empty inputs and once with every field it reads carrying a value.
- **Documented blind spots (3)** — done, in the test module's docstring: the literal-reading assumptions, and the specific note that `test_admin_only_services_verify_admin` matches call text and so cannot see an unreachable gate.
- **A regeneration gate (4)** — I don't think this one applies. The page is never committed (it renders into the gitignored `website/developer/`), so there is no stored output to diff against. `test_render_is_deterministic` covers the property you were protecting, and `test_render_covers_every_modelled_surface` catches a broken generator on a PR that never builds the site.
- **`services.yaml` / `strings.json` can still fall out of sync** — you were right that only the service *names* were pinned. Added `test_service_fields_match_between_yaml_and_strings`, since the reference takes a field's structure from one file and its label from the other; a field in just one renders half-blank.
- **Websocket and HTTP tables modelled but not rendered** — deliberate, and I'd like a second opinion on it. `AGENTS.md` calls the websocket a UI-latency optimization and explicitly never a substitute for the service, so publishing it would invite integrators onto a surface that isn't a contract. Modelling it still buys `test_every_websocket_command_is_registered`, which closed a hole nothing checked: the decorator list and the `async_register` list could already disagree.

What I'd most like scrutinized on this commit: whether restating the option keys in `api_surface` and pinning them by test is genuinely equivalent to deriving them (I claim yes, and that the pin is now a real assertion rather than a tautology), and whether `test_api_surface_imports_stay_light` is the right guard or too blunt.

---

prestomation:
/q review One small follow-up in `1b53ef8`, plus a status note. Same ask as before: critical and skeptical, most serious first.

**CI on `7811ad0` came back clean** — `preview` (the docs build that failed on the first commit) is green, along with `vale`, `mypy`, `ruff`, `Python`, `TypeScript`, `hacs`, `test` and coverage. `integration` and `e2e` were still running when I pushed this follow-up; both passed locally (162 and 67).

**What changed in `1b53ef8`:** removed `api_surface.service()`. The coverage report flagged it as the only uncovered code in the new module, and it turned out to be genuinely dead — two tests called it until I rewrote them to skip services the model doesn't know, so a missing `ServiceSpec` reports once in `test_every_registered_service_is_modelled` rather than echoing as a `KeyError` through every other service check. An unused public helper is the dead weight this module is meant not to carry, and `api_surface.py` is now fully covered.

Worth a specific look, since it's the kind of thing a reviewer catches better than a coverage percentage: are `triggers_for()` and `events_by_payload()` — the two functions left — pulling their weight, or would the call sites read better inline?

---

PR #261, opened 2026-08-28
feat(panel): hi-fi redesign — drawer editing, Settings rail, appliance sub-tabs

### LINKED ISSUE #253: Home Keeper fails against the Home Assistant beta
Nightly Home Assistant `beta` run failed: https://github.com/prestomation/ha-home-keeper/actions/runs/33101808733

prestomation:
Implements the [Panel UI Hi-Fi](https://claude.ai/design/p/d0474607-a721-4a11-ba3b-99215eed05f9?file=Panel+UI+Hi-Fi.dc.html) design across all four panel surfaces plus phone widths, and covers the responsive layout it introduces. No functionality is removed.

## What changed

**Design tokens.** A `:host` block maps the comp&#39;s palette onto Home Assistant theme variables, or `color-mix()` off them for the tints HA does not publish. Nothing hard-codes a colour, so dark and custom themes keep working. Two primitives the rest reuses: `.hk-eyebrow` (uppercase micro-label) and `.hk-indent` (a rule down the left of fields that exist only because of a choice above them).

**Tasks list.** Scope pills carry counts and share one row with Profile, Group by and the single primary action, so the old full-width action bar is gone. Group by became a dropdown — five visible options dominated the row. A row&#39;s descriptive chips sit beside the name (two, then a `+n` that unfolds the rest in place), and the due/overdue pill moved to the end of the line next to Done. Enclosure means pressable: the status pill lost its outline and Done gained a ring, so the row&#39;s one control is the one drawn as a control.

The Tasks tab: scope pills with counts on one filter row, task rows with a coloured left edge and a due pill at the end of each line

**Editing opens in a drawer beside the list**, not a card above it. The list keeps its place and stays fully live, the row being edited stays lit while the rest recedes, and Save sits in a header that does not scroll away. Delete and History join it in a footer. The form is one `ha-form` per section (Basics, Schedule, Placement, Completion) under its own heading, with the fields a recurrence choice reveals indented behind a rule.

Creating a task in the right-hand drawer, with the list behind it dimmed and section headings in the form

**Settings gets an anchor rail** naming every section and what it is set to: a green dot where a feature is on, amber where notifications are configured with no mobile app to deliver them, a count for the sections holding lists. Each card states its current value in a line under its name. The problem-sensor exclusions are indented behind the switch that decides whether they apply.

The Settings tab with a rail listing General, Shopping list, Problem sensor sync, Profiles, Notifications and Companions, each with a status dot or count

**An appliance is read next to the list it came from.** The list stays as a 268px pane with the open appliance marked in it, and the seven stacked sections become sub-tabs — Parts, Tasks, Documents, Details, Related, History — each carrying how much it holds. Each is a URL of its own (`/appliances/:id/documents`), so Back leaves a sub-tab and one can be linked to. A bare `/appliances/:id` still resolves, which matters because it is already written into every registered appliance device&#39;s `configuration_url`.

An appliance detail beside the appliance list, showing sub-tabs and the parts list with spare-stock bars

## Phone widths

The tabs move to the bottom of the screen, Add task becomes a floating button, and the drawer becomes a full-height sheet.

**The filter row wraps instead of scrolling sideways.** A joined segment cannot wrap — it is one pill with hairlines between its buttons — so below 700px it comes apart into separate chips, which can. Nothing is parked off the edge of the screen behind a swipe with no affordance, and it survives a locale whose labels run half again as long as English.

<p>
  The Tasks tab on a phone, with the scope pills wrapped onto two rows above the list
  The Appliances tab on a phone
</p>

**Settings opens on an index of its sections.** A screen with no room for a rail beside six expanded sections has no room for the six sections either, so the rail&#39;s chip strip is gone. Each section is a row naming it, what it holds, and its dot or count; tapping one opens that section by itself with a back arrow. The route `/settings/:section` joins the table, so Back leaves a section like any other page and a section can be linked to. An unrecognised section falls back to the index rather than an arbitrary default — unlike an appliance sub-tab, the index is a destination in its own right.

<p>
  The Settings section index at phone width: six rows, each naming a section, what it holds, and its status dot or count
  One Settings section open on a phone, with a back arrow to the index
</p>

**Responsiveness stays in CSS.** All three parts — rail, index, and every section — render at every width. The layout carries which section the URL names as `data-section` and the named card carries `.hk-sec-current`, both facts about the route rather than the viewport, and the media query picks between them. The one exception is the drawer&#39;s *modality*, which is an attribute rather than a style.

A breakpoint here has to allow for the fact that a media query measures the viewport while the panel gets the viewport minus HA&#39;s ~256px sidebar. The sheet threshold is 1150px and the appliance master pane steps aside at 1000px for that reason: at 900px a 400px drawer sat beside a 320px list and broke task names one character per line.

### Deliberate deviations from the comp

- **Field chrome stays HA&#39;s.** The comp puts labels above the box and helper text behind a question-mark affordance; `ha-form` owns that and has no property for either. Replacing it would cost the entity/device/area/label/tag/datetime selectors.
- **Parts are not a five-column table.** With the appliance list beside it the pane is ~700px, which wraps &#34;Every 12 months&#34; onto two lines — and a per-row grid cannot align columns across rows anyway, since each row is its own grid container. Each part still answers the same three questions in the same three places, with a bar for how much of the reorder point is left.
- **No &#34;Show N more&#34; truncation and no Duplicate action** — agreed out of scope; Duplicate has no service or websocket command behind it today.

## Testing the layout, not just photographing it

The redesign introduced the panel&#39;s entire responsive layout — before it `panel.ts` had no `@media` rules at all — and seven of seventy-six e2e tests looked at any of it. The rest ran at 1280px and would not have noticed the narrow rules regressing.

Rather than write a second suite, **the same test bodies now run in three viewport projects**: `desktop` (1280×720), `tablet` (820×1180) and `phone` (390×844). A spec opts in with a tag — `@responsive` for assertions that hold at every width, `@narrow` for a layout with no desktop form, `@phone`/`@tablet` for one band. Untagged is desktop-only, which stays the default and covers the Lovelace specs.

**84 tests → 186** (desktop 80, phone 54, tablet 52). Desktop drops four because the `settings-narrow` tests that used to set their own phone viewport now run in the phone *and* tablet projects instead — two widths each rather than one. CI goes 3m28s → **5.6 min against a 15-minute cap**.

**Tablet is 820px, and it is the interesting one.** It sits under the drawer&#39;s 1150px sheet threshold and under the 1000px where the rail and the master pane step aside, but over the 700px phone chrome — so the top tabs are still up while the drawer is already a sheet. Nothing exercised that combination.

What made it possible without duplicating test bodies is that both tab bars are in the DOM at every width and CSS picks, so `gotoTab()` clicks whichever is *visible*. That unpins seventeen call sites without reading the viewport. `expectTabActive()` fixes an assertion that was about to go quietly vacuous: three deep-linking tests said &#34;we got back to the list&#34; by checking `#tab-tasks` was visible, and the phone bar is visible on every view.

The projects do not replace `page.setViewportSize()`. A test that crosses a breakpoint *within* one page — proving the split is CSS, or reaching `_syncDrawerModality()`&#39;s `matchMedia` listener, the only viewport read in the panel — stays untagged and resizes itself.

A new `responsive-layout.spec.ts` pins what nothing asserted: the chrome swap, the FAB, no horizontal overflow, 44px tap targets, the wrapped row order, the master pane stepping aside, the sheet geometry. It measures relationships and computed keywords, never coordinates, and both fixes below were re-broken to confirm the assertions fail.

**The walkthrough now records the phone layout too.** The tour said it could not — &#34;a 390px viewport would sit in a third of a 1280×800 frame beside a grey void&#34; — which is true only of resizing inside one recording, since `recordVideo.size` is fixed at context creation. A second context at 390×844 records it correctly framed (verified: the mp4 is 390×844, no letterboxing). It is a different set of beats, not the desktop script at phone width.

### Bugs this surfaced, all fixed with tests

`ha-form` emits its whole `data` object on every change. Seeding each section with the entire form made every section re-assert a stale snapshot of the others, so typing a name and then picking a recurrence put the name back to empty and the save created nothing. Sections are now seeded with only their own fields.

With the appliance detail rendered beside the list, `_hydrate` wired the row openers twice, so every click pushed two history entries and Back out of a task opened from an appliance landed on the same task.

Two CSS specificity ties: restoring the first chip&#39;s border repainted the active chip white-on-white, and the chip-overflow &#34;open&#34; rule lost to the &#34;hide&#34; rule so the chips stayed folded. The `+n` control had also never worked — it is a child of the row, so the rule hiding everything past the second chip hid the control that reveals them.

Focusing an HA element right after an `innerHTML` assignment throws (its `focus()` dereferences a shadow root that does not exist yet), and the unguarded throw was propagating out of `_render()` and skipping everything after it.

**A wrapped phone task row carried an empty line.** `.hk-row-spacer` pushes Done to the right end of a *single-line* row; once the row wraps, `margin-inline-start: auto` does that job instead. Left at `order: 0` beside a `.grow` that is `flex: 1 1 100%`, it could not share a line and took a full-width one of its own — measured as a phantom 10px gap under the task name.

**The phone task list was photographed with a Profile filter still applied**, left behind twenty capture steps earlier. A filtered list hides the scope pills, so the shot documented the phone layout while omitting its headline change. **&#34;Home Assistant has started!&#34; was also sitting across the task detail&#39;s history** — the capture always runs against a freshly-booted container.

`field.due` was missing from all 16 locales while its 66 `field.*` siblings were present, so the one-off due-date picker showed its raw translation key. Pre-existing on `main`.

Two capture bugs, both from the drawer: it is `position: sticky`, so a full-page shot rendered it at the page&#39;s scroll offset with blank paper above; and its scroll container is capped at `100vh`, which a full-page capture expands to the whole document.

`tests/e2e/mobile.capture.ts` and `mobile.config.ts` were referenced by no workflow and no script — the phone screenshots were hand-produced and could rot silently. Deleted; the views they had that nothing else did joined the documented set.

## Accessibility

Three reviews went over the redesign; the accessibility one found the two most serious problems, both introduced here.

- **Focus survived nothing.** `_render()` replaces the whole shadow tree, so every filter click, sub-tab and rail entry dropped the keyboard at ``. It is restored by the stable `data-*` attribute each control already carried.
- **The phone sheet was not a dialog.** It covered the list while leaving ~30 tabbable controls underneath it. The content behind is now `inert`, the sheet is `role=&#34;dialog&#34; aria-modal=&#34;true&#34;`, focus enters it, Escape closes it and focus returns to the opener. Beside the list, where it is a panel rather than an overlay, none of that applies and the list stays live.
- Filter chips carry `aria-pressed` and their group a name; the dropdowns have accessible names; the rail&#39;s status dots have a text equivalent; `prefers-reduced-motion` is respected; the phone controls meet the 44px target the file defines — now asserted rather than asserted-about.
- The sub-tabs and phone bar had declared `role=&#34;tab&#34;` with no tabpanel, roving tabindex or arrow keys. They are navigation between URLs, so they are buttons with `aria-current=&#34;page&#34;`.

Contrast is measured off rendered pixels in both themes rather than eyeballed. Done went from **2.85:1 to 6.05:1** (light) and 6.34:1 (dark); &#34;Integration offline&#34; was white-on-amber at **1.96:1** and the archived chip white-on-grey at **1.88:1**; the selected filter chip was white on the accent at 3.26:1. The `*-ink` tokens now mix 58% rather than 78% of the hue.

## Merged with `main`, twice

`0.18.0` shipped while this branch was in review, and #265 landed after that. Both were merged in rather than rebased — the branch has an open review history, and a merge commit keeps existing checkouts valid.

- **The first merge** resolved twelve text conflicts plus every committed screenshot. `_wireCopyButtons` moved above the `if (this._detail.kind !== &#39;asset&#39;)` split in `_hydrate`, since an appliance detail falls through to the list wiring where a task detail returns early. And #263 had replaced readable fixture ids with uuid4s, so `task_door_battery` and friends no longer exist in the seed — `a11y.spec.ts` merged *cleanly* and still needed that migration by hand, which is the failure mode to watch: git is happy, the record is gone, the spec fails at runtime. Every screenshot was re-taken, since the merge brought main&#39;s id rows into pages this branch had rewritten.
- **The second merge** brought in #265 (the HA 2026.9 device-registry shim). Python-only, so the sole conflict was the CHANGELOG.

**The version is `0.19.0b2`, and that was a CI catch rather than a judgement call.** `0.18.0b3` was clearly wrong once `0.18.0` shipped — PEP 440 sorts it below the stable, so HACS would have offered the stable as an &#34;upgrade&#34; to a beta tester. But the branch then sat in `## [0.19.0b1]`, and merging #265 made that a *published* section (`v0.19.0b1` is tagged). `changelog-release-gap` failed the push and was right to: `release.yml` keys off an unchanged manifest version, would have seen the tag already existed, and skipped — the entire redesign would have landed on `main` documented but never shipped. That is the #236 failure mode the gate exists for. The redesign now has its own `## [0.19.0b2]`, and `## [0.19.0b1]` keeps only #265&#39;s `(Fixes #253)`.

Pre-existing issues the same reviews raised that this branch does **not** touch are filed as #262.

## Checklist

- [x] Tests run locally — 731 vitest, 1530 pytest unit, 186 Playwright across three viewport projects, all green
- [x] Mutation testing: **100%** on the changed surface (`forms.ts`, `utils.ts`), threshold 80%
- [x] `CHANGELOG.md` updated
- [x] Panel UI changed → current screenshots committed under `docs/images/`, every changed image read and inspected, embedded above with SHA-pinned HTML image tags
- [x] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended (drawer, appliance sub-tabs, Settings rail routing) **and a phone tour added**; both transcoded and reviewed locally
- [x] New user-facing feature → version bumped to `0.19.0b2` (`manifest.json` + `const.py`), `preview-release` label applied
- [x] Conventions recorded in `.amazonq/rules/architecture-and-code.md` and `testing-and-workflow.md`

https://claude.ai/code/session_01DgmwWdJetyx8Gc5vU1SC7h


prestomation:
/q review This is a large panel redesign. Please be critical and skeptical, surface the most serious issues first, and don't withhold minor ones.

Topics I'd most like scrutinised:

**Correctness.** The task form was split from one `ha-form` into one per section (`taskSchemaSections` in `forms.ts`), and each section is seeded with only its own fields via `pickFormData`. Two bugs already came out of that — a section re-asserting a stale snapshot of the others, and form-seeded defaults no longer reaching the edit state. Please look hard for a third: any handler in `_renderTaskForm`'s `onChange` that still reads a field without checking it is in this section's event, anything where the "adopt what the form seeds" merge could resurrect a value the user cleared or a field belonging to a hidden branch of the schema, and whether `taskFormSchemaKey`'s re-render gate still behaves with partial events.

**Routing.** `parseRoute`/`buildPath` gained an optional appliance sub-tab segment. Check the round-trip, the fallback for an unknown segment, and that `/appliances/<id>` — already baked into every registered appliance device's `configuration_url` — still resolves. Sub-tab switches `replace` rather than `push`; is that right for Back?

**Event wiring.** `_hydrate`'s early return for detail pages was split so an appliance detail also gets the list wiring. Duplicated listeners there already caused double history pushes. Are there other listeners now attached twice on that route?

**HA best practices / theming.** Every colour goes through `--hk-*` tokens that resolve to HA theme variables or `color-mix()` off them. Please check for anything that would look wrong in a dark or custom theme, and whether `appearance`/`variant` on `ha-button` are the right API to lean on.

**Performance.** `_render()` rebuilds the whole shadow tree; the appliance route now renders the master list *and* the detail on every pass. Is anything doing avoidable per-render work (the filter counts iterate every task × 4 scopes)?

**Maintainability.** The `STYLES` literal grew a lot. Naming, duplication, and whether the sticky/fixed layout rules are comprehensible from the comments.

---

prestomation:
/q review Be critical and skeptical, and surface the most serious problems first — but don't withhold minor ones. Two new commits since your last pass, both about the panel at phone width.

**1. The Tasks filter row (`46e1ce2`).** The scope pills used to sit on one `nowrap` line with `overflow-x: auto` below 700px. They now come apart into separate chips that wrap. Please scrutinise:
- **CSS cascade.** The joined-segment rule `.hk-seg-btn:first-child { border-left: 0 }` ties on specificity with `.hk-seg-btn.active`; restoring the border there already repainted the active chip white-on-white once. Are there other ties in the `@media (max-width: 700px)` block that silently undo a base rule? Check `.hk-card-row` ordering too — chips, status and actions now carry explicit `order` values.
- **Theming.** Everything is `var(--hk-*)` off HA theme variables and `color-mix()`, never a literal. Does anything in these two commits break under a dark or custom theme?

**2. Settings becomes an index → section on a phone (`a7ba213`).** New route `/settings/<section>`, a section index, and a back bar.
- **Correctness of the routing.** `parseRoute`/`buildPath` now round-trip a `section`, and `_applyLocation` compares it. Is there a path where the URL and `_settingsSection` can desync — a deep link, a tab tap, `history.back()` from a section with no pushed entry, or `_switchView` popping to root?
- **The load-bearing claim.** Responsiveness is meant to stay CSS-only: the panel renders rail + index + all six sections at every width, puts `data-section` on the layout and `.hk-sec-current` on one card, and the media query picks. Does any of `_render`/`_hydrate` actually read the viewport? And is rendering all six sections on a phone (five of them `display: none`) a real cost — six `ha-form` trees with live `hass` bindings — or noise?
- **Privilege model.** The panel is `require_admin`; nothing here touches a service or websocket command, but please confirm the new route can't reach anything the old `/settings` couldn't.
- **HA best practices** for the new `ha-card`-based index rows and the `ha-button` back bar, and whether the keyboard/focus story on `.hk-index-row` is adequate.

Also worth a look: `_settingsSectionList()` is now the single source for both the rail and the index — is the summary derivation (`_settingsSummary`, keyed on card id) a sensible seam, or should the key be the section rather than the DOM id?

---

prestomation:
/q review Be critical and skeptical, most serious first, and don't withhold minor findings. One new commit (`484f4bf`) acting on three reviews — lost behaviour, contrast, and keyboard/screen-reader access. Please scrutinise:

**1. The drawer's split personality.** It is a side panel above 1150px and a modal bottom sheet below it, and those two need opposite behaviour: beside the list the list must stay fully live (the inline form it replaced never blocked it), while covering the list it must be `inert` or a keyboard user tabs into ~30 controls under an opaque overlay. `_syncDrawerModality()` reads `matchMedia` to decide which it is — the one place in the panel that reads the viewport. Is that seam correct, and is there a state where it desyncs (a resize across the breakpoint with the drawer open, a re-render mid-edit, unmount with the document `keydown` still attached)?

**2. Focus restoration across a wholesale re-render.** `_render()` replaces `shadowRoot.innerHTML`, so `_focusKey()` records the active element's stable `data-*` attribute and `_restoreFocus()` finds its replacement after. Is the key unambiguous — two segments can share an option value, which is why it also carries the group. Any path where it focuses the *wrong* control, or steals focus from a field the user is typing in?

**3. `_focus()` swallowing exceptions.** HA's `ha-button.focus()` throws when called right after an `innerHTML` assignment, and the unguarded throw was propagating out of `_render()` and skipping everything after it. The guard is a bare try/catch. Is that hiding a class of real error it shouldn't?

**4. CSS specificity.** Two bugs in this area already: restoring the first chip's border tied with `.hk-seg-btn.active` and repainted it white-on-white, and the chip-overflow "open" rule lost a tie to the "hide" rule so chips stayed folded. Please sweep the new rules in `STYLES` for others — particularly `.hk-status ha-assist-chip:not(.hk-overdue):not(.hk-managed):not(.hk-warn)`, the `nth-child(n+3):not(.hk-chip-more)` pair, and `.hk-settings-layout[data-section] ~ .ver`.

**5. Contrast durability.** The `*-ink` tokens moved 78% → 58% of the hue mixed into `--primary-text-color`. Measured light and dark and all the pairs now clear 4.5:1 — but is 58% right across *custom* themes, where `--primary-color` could be anything? Is there a pairing that inverts?

**6. `aria-current="page"` on the sub-tabs and phone bar.** They previously claimed `role="tab"` with no tabpanel, roving tabindex or arrow keys. Is plain navigation the right call, or does the appliance sub-tab (which re-renders in place at a new URL) genuinely want the tab pattern implemented properly?

Also: the chip-overflow button is new state (`_chipsExpanded`) that is deliberately *not* in the URL, unlike everything else in this panel. Reasonable, or an inconsistency worth fixing?

---

prestomation:
/q review Be critical and skeptical, most serious first, and don't withhold minor findings. Six new commits: a merge with `main`, and then multi-viewport e2e coverage for the responsive layout this branch introduced.

**1. The merge (`ee1bbb2`).** `0.18.0` shipped mid-review, so this resolved twelve text conflicts plus every screenshot. Two resolutions are judgement calls I'd like challenged:
- `_wireCopyButtons(root)` moved *above* the `if (this._detail.kind !== 'asset')` split in `_hydrate`, because an appliance detail now falls through to the list wiring where a task detail returns early. Is it wired exactly once on both paths, and can the fall-through reach it twice?
- #263 replaced readable fixture ids with uuid4s. The conflicted hunks took main's constants, but `a11y.spec.ts` is new on this branch and merged **cleanly** while still naming `task_door_battery` — migrated by hand. Please sweep for any other clean-merge site that still names a record the seed no longer has.

**2. The tag → project selection (`34b91f3`).** Three projects; a spec opts in with `@responsive` / `@narrow` / `@phone` / `@tablet`, untagged is desktop-only. Playwright greps the project name as well as the tags, which is why every tag is `@`-prefixed. Is the tag set closed and unambiguous, is anything now running at a width where its assertion is *true but meaningless*, and did I mis-tag anything that should have stayed desktop-only? In particular `measured-stock.spec.ts`'s last two tests take no `page` at all — harmless waste, or a signal they belong elsewhere?

**3. `gotoTab()` / `expectTabActive()` / `openSettingsSection()`.** The claim is that these read what is *visible* rather than the viewport, preserving the "only CSS decides layout" invariant. `openSettingsSection` branches on `await row.isVisible()` after settling on `.hk-settings-layout` — is that a race under a re-render, and is the earlier `index.or(rail)` strict-mode trap I hit lurking anywhere else? `expectTabActive` replaced three assertions that were about to go vacuous at phone width; are there others in the suite with the same shape?

**4. `responsive-layout.spec.ts`.** The doctrine is computed keywords for `position`/`flex-wrap`/`overflow-x`, bounding boxes for relations, no absolute coordinates, and a non-empty check before every loop. Please hunt for assertions that could pass vacuously, tolerances that are too loose to catch a real regression (or too tight to survive a font change), and anything that duplicates the modality contract `a11y.spec.ts` owns. The scroll-invariance checks assert the page actually scrolled first — is that sufficient to prove "anchored to the viewport"?

**5. The `.hk-row-spacer` fix.** `display: none` inside the 700px block, because at `order: 0` beside a `flex: 1 1 100%` sibling it took an empty line. Is hiding it right, or should it carry an `order` instead — and does hiding it change anything on a row that happens *not* to wrap at that width?

**6. The capture-config pin (`capture-config.ts`).** Without it every capture would inherit three projects and write each committed PNG three times. Is `captureConfig()`'s merge of `use` correct, and is there any path left where a harness spreads the base config directly?

**7. The two-context walkthrough.** `recordVideo.size` is fixed at context creation, so each width gets its own context and `test()`. Is splitting into two tests right for a soft gate, is `recordTour`'s `finally` safe if the tour throws before the first navigation, and does `ci/capture-video.sh` degrade sensibly when one variant is missing?

Also: CI cost. Desktop stays 84 tests; phone and tablet add 47 each, ~4.7 min locally against a 15-minute cap. Anything here likely to be flakier at a narrow width than a wide one — particularly clicks near the fixed bottom bar and the floating Add button?

---

prestomation:
Thanks — that review was against `1e7c51b`. Your one recommendation was already addressed in `44aef34`, pushed just after: the two `measured-stock` tests that take no `page` no longer run at three widths.

I took the tag off the describe and put it on the two tests that actually open a browser, rather than extracting the API-only pair to a separate file as you suggested. Same outcome for CI time, and it keeps the `afterEach` that restores the seeded stock covering all four — a separate file would need its own copy of it, and the last test depends on the fractional-adjustment test running ahead of it in the same project. Desktop is unchanged; phone and tablet drop two apiece, 190 → 186.

Two corrections to the review, for the record:

- **`recordTour`'s `finally`.** You said `page.video()` returns null when nothing was recorded. It doesn't — with `recordVideo` set on the context, it returns a `Video` regardless. The `if (video)` guard is for the no-`recordVideo` case, not the empty-recording one. The behaviour is still right (a partial recording on a thrown tour is fine, and the failing test is what reports it), but not for the reason given.
- **`.hk-row-spacer`.** You read line 216 as the fix. It's the assertion; the fix is `display: none` in the 700px block of `panel.ts`. You reached the right conclusion about `order` not being the answer, which is the same reason I didn't use it.

I verified your capture-config claim independently — no `*.config.ts` spreads `baseConfig` directly, all eight go through `captureConfig()`.

---

/q review One new commit since your last pass (`44aef34`), plus please re-check two things you assessed on the older head.

**1. The new commit.** `measured-stock.spec.ts` moves `{ tag: '@responsive' }` from the describe onto its two browser tests. Does the `afterEach` still cover all four, and is the ordering dependency the last test has on the fractional-adjustment test still guaranteed — both now run only in the `desktop` project, but is that a guarantee or an accident of declaration order? And is per-test tagging the right shape here, or does mixing tagged and untagged tests in one describe make the file harder to reason about than splitting it would?

**2. Re-check on the current head: is any *other* describe over-tagged the same way?** I tagged ten describes wholesale. Please look for tests inside them whose assertion cannot vary with viewport — anything that never touches `page`, or that only asserts on service/entity API state — and name them rather than telling me the pattern is fine.

**3. You said "no other assertions in the suite follow that vacuous shape" for `expectTabActive`.** That is the claim I most want a second look at, because it is the one that fails silently. Please check specifically for assertions that prove *presence* of something rendered at every width while meaning *state* — a locator that resolves in both layouts but only means what the test intends in one of them. `#back-btn`, `#hk-list` and `.hk-index-row` are the ones I'd start with.

---

prestomation:
/q review Two commits since your last pass, both from main moving underneath this branch. Be critical and skeptical; most serious first.

**1. `00ae5034` — merging #265 in.** It is Python-only (the HA 2026.9 device-registry shim) and touches no frontend source, so the only conflict was the CHANGELOG: both sides had independently opened a `## [0.19.0b1]` section. Resolved as a union. Please check the merge did not silently drop anything from either side — `coordinator.py` is the file to look hardest at, since it is the one #265 rewrote and this branch had also touched.

**2. `21152a9a` — cutting `0.19.0b2`.** `changelog-release-gap` failed on the previous push and was right to: #265 had already shipped as `v0.19.0b1`, so the section this branch's entries were sitting in is a *published* one. `release.yml` keys off an unchanged manifest version, would have seen `v0.19.0b1` already tagged, and skipped — the whole redesign would have landed on main documented but unreleased. That is the #236 failure mode.

The fix splits the redesign into `## [0.19.0b2]` and bumps `manifest.json` + `const.py`, leaving `## [0.19.0b1]` as the released section carrying only #265's `(Fixes #253)`.

What I want scrutinised: **is the split correct, or does it strand something?** Specifically — does `ci/release-issues.py` still find every `(Fixes #N)` it needs to notify, given `(Fixes #253)` now sits in a section that is already tagged and will not be re-published? #253 was notified by the `v0.19.0b1` release, so I believe it is fine, but that is the kind of thing that fails silently a release later. And is `0.19.0b2` the right number rather than `0.19.0`, given the redesign is the headline feature of the 0.19 line?

**3. Anything from your last review I have not addressed.** You flagged the redundant API-only tests, which `44aef34` fixed. If your second review (the one asking about over-tagged describes and vacuous presence-vs-state assertions) turned anything up that I have not acted on, please restate it here — this PR is about to merge and I would rather catch it now than after.

---

PR #263, opened 2026-08-29
feat(services): accept a name wherever a service takes an id, and show ids in the panel

prestomation:
Fixes #

## What changed

Home Keeper's ids are `uuid4`s (`models.build_task`, `assets.build_asset`) that appeared in **no UI a person reads**, yet all 34 services identify their target by one — every `task_id` / `asset_id` / `part_id` / `document_id` is a bare `selector: text:`, so the Actions dialog renders an empty box. The only documented ways to learn a `task_id` were to call `add_task` with `return_response` at creation time, or read uuids out of a `list_tasks` dump. `README.md` never mentioned `task_id` at all while pointing users at `complete_task`/`snooze_task`/`skip_task` for automations. #220 is a live example: a user hand-wrote a script calling `adjust_part_stock` from a dashboard button.

**Every `*_id` field now accepts the object's name as well as its id.**

```yaml
action: home_keeper.complete_task
data:
  task_id: Replace furnace filter
```

### Following HA core's convention

`todo.update_item` is the direct precedent — a list of user-named items in an integration-owned store. It uses **one field, not two**: `item`, labelled *"Item name or UID"*, resolved by `_find_by_uid_or_summary` (`value in (item.uid, item.summary)`), with the duplicate-names caveat in the field description. There is no `*_name` sibling anywhere in core.

So: one field, id first, **no rename** — existing automations and the `docs/INTEGRATING.md` contract are untouched. The labels become "Task name or ID" and friends, and the descriptions gain a sentence.

This also matches what the repo already did for profiles/notifications (`profiles.resolve_profile`, `notifications.resolve_notification`, surfaced as *"Name or id of a saved profile"*).

### One deliberate departure from core

Core takes the **first** name match and never mentions the others. Home Keeper **raises** instead: names here are not unique by design (`INTEGRATING.md` tells contributors to expect collisions) and the services reached this way include `delete_task` and `delete_asset`. Silently deleting one of two identically named tasks is worse than an error naming both.

```
ServiceValidationError: Several tasks are named Replace furnace filter.
Use an id instead: 10561eb5-1ab5-4151-aa8e-5f6269a0c28a, task_furnace_filter
```

A name matching *nothing* is passed through untouched, so the handler's existing not-found error quotes what the user actually typed (`Task not found: Wash the dog`) — no error messages changed.

### Where the id comes from

The panel now shows it with a copy button, which is what makes the id form reachable when a name is ambiguous. `copyText` (`utils.ts`) falls back to an off-screen textarea + `execCommand`, because `navigator.clipboard` is **secure-context only** and plenty of people reach HA over plain HTTP — the same trap `randomId` already documents. It reports failure rather than claiming a copy that never happened.

Task and appliance pages get a labelled detail row; parts and documents get a compact form (no label, no divider) so the densest lists keep their shape.

Task detail page showing an ID row with the task's id and a copy button

Appliance detail page showing the appliance id in the Details card, plus a compact id under each document and each part

### Also documented

`todo.update_item` **already** completes a Home Keeper task by name — `todo.py:84` sets `summary=task["name"]`, and `async_update_todo_item` routes `COMPLETED` into `store.complete_task`. That worked before this PR and was documented nowhere; the README now says so.

### Scope notes

- Resolution lives in a new pure `resolve.py` (HA-free, added to the mutmut allowlist). Websocket commands stay id-only — the panel holds full objects and never types a name.
- Fixes a **pre-existing** mutmut gap: `test_generate_api_docs.py` reads `website/scripts/doc-map.mjs`, which was not in `also_copy`, so the whole Python mutation run died at stat collection. It stayed hidden because no PR touching mutable Python had run since that assertion landed — exactly the trap `[tool.mutmut]`'s own comment warns about.
- The convention is recorded in `.amazonq/rules/architecture-and-code.md`.
- Screenshots were regenerated in one pass against a freshly seeded container. Surfaces with no id row still differ because the run picked up a newer HA stable than the previous capture.

### Verified

Mutation **100%** on both sides (Python 90/90, TypeScript 13/13). Against a real HA container: name resolution, case/whitespace tolerance, raw ids unchanged, appliance+part by name, the ambiguity refusal, and the not-found message quoting the typed text.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v` 1381 passed, `bash ci/test-frontend.sh` 662 passed; `tsc --noEmit`, `ruff`, `vale`, both mutation gates, and the new e2e spec on **both** the local Chromium and CI's headless shell)
- [x] `CHANGELOG.md` updated — new `## [0.18.0b3]` section (`v0.18.0b2` is already published, so this opens the next beta)
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML `` tag
- [ ] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended to step through it — **n/a**: this adds a row to existing detail pages, not a new surface or flow

---
_Generated by [Claude Code](https://claude.ai/code/session_01QypSGC8Tov9Z5WYi9nD7VF)_


prestomation:
/q review Please be critical and skeptical, and lead with the most serious problems — but don't withhold the minor ones either.

This PR makes every `*_id` service field also accept an object's name, resolved by a new pure module `custom_components/home_keeper/resolve.py`, wired into ~25 handlers in `__init__.py`. Areas I'd most like torn apart:

**Correctness of the resolver.** `resolve.py` does three passes: exact id, exact name, then trimmed/case-folded name, with ambiguity raising per pass. Is the per-pass ambiguity rule actually right? Consider tasks named `"Filter"` and `"filter"` — the exact pass separates them, the folded pass collides. Are there inputs where a caller gets a *different* object than they'd reasonably expect, or where a rename silently redirects an automation to another task? Also: `_by_name` makes two passes over its argument (typed `Sequence` for that reason) — check no caller can hand it an iterator. And `_norm` returning `""` for a non-string is marked `# pragma: no mutate` as equivalent; please challenge that claim.

**The handler wiring.** This was a large mechanical edit and mechanical edits hide bugs. Every `call.data["task_id"]` became a resolved local. Please check for: a handler that resolves the wrong field, one that still reads raw `call.data` where it should use the resolved value, an error `translation_placeholders` now showing a resolved id where the user's typed text would be more useful (the intent is that a *failed* resolution passes the key through unchanged so the existing not-found message quotes what they typed — verify that actually holds on every path), and the ordering in `set_task_consumable` / `delete_archived_completion` / `adjust_part_stock`, where a part or document must resolve *within* an already-resolved asset. `handle_sign_document_url` and `handle_sign_part_file_url` gained a `_coordinator()` call they never had — is that safe there?

**Security / privilege.** Resolution now happens *before* some admin checks in reading order — does any path let a non-admin probe for the existence of an object (or enumerate ids via the ambiguity error, which lists candidate ids) that they couldn't see before? `_verify_admin` ordering relative to `_asset_ref` is worth a hard look.

**HA best practices.** Is one field accepting name-or-id the right call versus a separate field, given `todo.update_item`'s `item` ("Item name or UID") precedent? And is deviating from core by *raising* on an ambiguous name defensible, or should it match core and take the first match?

**Frontend.** `copyText` in `utils.ts` falls back to `execCommand` when `navigator.clipboard` is missing (non-secure context). Any leak, focus, or XSS concern with the temporary textarea? The ids are escaped into `data-copy` — check that's sound.

**Performance.** `_task_ref` calls `store.get_tasks()` and `_asset_ref` calls `store.get_assets()` per service call. Is that a real cost on a large store, and does any handler now build that map more than once?

---

prestomation:
/q review Follow-up push — prose only, no behaviour change. Please stay critical and lead with the most serious findings.

`docs/INTEGRATING.md` lost a colon-led sentence that tripped `ai-tells.VerbTricolon`, and the `strings.json` exceptions block is no longer re-sorted (sorting had moved a pre-existing line into the diff-scoped lint's added set). Worth re-checking that the reordering revert didn't drop or duplicate any exception key across `strings.json`, `translations/en.json` and the 15 other locales — the four new `*_ambiguous` keys should be present in all 17 files with identical placeholders (`{name}`, `{ids}`), and nothing else should have moved.

The earlier request still stands and is the more important one: the resolver's correctness in `resolve.py`, the mechanical handler rewiring in `__init__.py`, and the privilege ordering. On that last point I've since verified every `_verify_admin` handler calls it before any `_*_ref`, and that the ids the ambiguity error can reveal are ones `list_assets`'s non-admin `card_projection` already returns (asset id, document id/name, part id/name) — please challenge both claims rather than take them.

---

prestomation:
Thanks — triage of the review, one finding taken and two disputed.

### Taken: the `_norm` pragma (pushed in 2f5c301)

You were right and my "equivalent" claim was too confident. It only held while every *key* is a string. With a non-string key, the sentinel folds to the same value as every object whose name is also non-string, and they match:

```
orig, non-string key: None
mut,  non-string key: a1      # the mutant resolves to a nameless object
```

Rather than defend it, the sentinel is gone. `_norm` now takes a `str`; the type check moved to a `_folds_to` helper at the comparison site, and `_by_name` rejects a non-string key up front. No pragma left in the file, and the resolver scores **100% on 96 mutants** instead of 90 with one excused. Added `test_a_non_string_reference_matches_nothing` for the exact pair above.

### Disputed: "🔴 Critical: privilege leak via ambiguity errors"

This one doesn't hold, and the finding's own line numbers contradict it — it says `handle_update_asset` resolves "at line 1121 **before** the admin check at line 1117", but 1117 comes first. `_verify_admin` is the first statement in that handler and in every other admin-gated one. Checked mechanically across all twelve rather than by eye:

```
OK  handle_add_asset        OK  handle_adjust_part_stock     OK  handle_export_inventory
OK  handle_update_asset     OK  handle_remove_part_file      OK  handle_set_options
OK  handle_delete_asset     OK  handle_add_asset_document
OK  handle_archive_asset    OK  handle_remove_asset_document
OK  handle_restore_asset    OK  handle_update_asset_document
```

(each asserts the index of `_verify_admin` precedes the first `_*_ref`). A non-admin never reaches resolution on any of them.

### Disputed: `sign_document_url` / `sign_part_file_url`

Correct that these two aren't admin-gated — but that predates this PR (`api_surface.SERVICES` marks them non-admin), so there's no regression here, and adding `_verify_admin` would be an unrelated privilege change I shouldn't slip into this PR.

The leak concern doesn't apply either. A non-admin calling `list_assets` gets `assets.card_projection`, which already returns the asset `id`, the **full** `documents` list (ids and names included) and each part's `id`/`name`/`url`. So the ambiguity error can only name ids that caller can already enumerate directly — no new information.

### Minor notes

- **Performance:** `store.get_tasks()` / `get_assets()` `return self._tasks` / `self._assets` — the live dict by reference, not a copy. Two calls in `delete_archived_completion` is two attribute reads, so there's nothing to cache.
- **`AmbiguousName` placeholders:** the message carries both, not just ids — `"Several tasks are named {name}. Use an id instead: {ids}"`.

One clarification for anyone reading the screenshots: they show the seeded e2e fixture, whose ids are hand-written (`asset_water_heater`) so specs can target them. Real ids are `uuid4` — `build_task` and `build_asset` both mint `str(uuid.uuid4())` and accept no caller-supplied id. Creating a task against a live container returned `10561eb5-1ab5-4151-aa8e-5f6269a0c28a`.

---

prestomation:
/q review Second pass, please — stay critical and lead with the most serious findings.

Since your last review, `2f5c301` reworks `resolve.py`: the `# pragma: no mutate` sentinel in `_norm` is gone, `_norm` now takes a `str`, the type check moved to a new `_folds_to` helper at the comparison site, and `_by_name` rejects a non-string key up front. Please check that this actually closes the hole rather than moving it — in particular whether `_by_name`'s new guard and `_folds_to` agree on what counts as blank, and whether rejecting a whitespace-only key early changes behaviour for any object legitimately named with surrounding whitespace.

Also still open from the first pass, and I'd like these argued rather than accepted:

1. I disputed the "critical privilege leak" finding — `_verify_admin` is the first statement in all twelve admin-gated handlers, and the finding's own line numbers (resolve at 1121, admin check at 1117) put the check first. If you still believe a non-admin can reach resolution on an admin-gated service, name the handler and the call path.
2. I disputed the `sign_document_url` / `sign_part_file_url` concern on the grounds that `assets.card_projection` already hands a non-admin the asset id, the full `documents` list and each part's id/name — so an ambiguity error there reveals nothing new. Please verify that projection really does include document ids, and say so if it doesn't.

Beyond that, the earlier scope stands: the mechanical handler rewiring in `__init__.py` (a wrong field resolved, a raw `call.data` read left behind, the asset-before-part ordering), the `_with_parent_ref` path on `add_asset`/`update_asset` which you haven't commented on yet, and HA best practices.

---

prestomation:
Agreed on the conclusion, but one premise in that reasoning is wrong and shouldn't stand on the record.

> **However**, this is moot because all handlers that resolve parts/documents (`_part_ref`, `_document_ref`) are **admin-gated** — they're only reachable through handlers that call `_verify_admin` first.

Two of them aren't. `sign_document_url` and `sign_part_file_url` resolve a document/part and are **not** admin-gated — no `_verify_admin` in either handler, and neither carries `admin_only=True` in `api_surface.SERVICES`:

```
handle_sign_document_url:  _verify_admin=NO  resolves=['_document_ref(']
handle_sign_part_file_url: _verify_admin=NO  resolves=['_part_ref(']
```
```python
ServiceSpec("sign_document_url", response="only"),      # api_surface.py:193
ServiceSpec("sign_part_file_url", response="only"),     # api_surface.py:194
```

That's pre-existing (both were non-admin before this PR), but it means a non-admin *can* reach part and document resolution, so the admin gate isn't what makes this safe.

What makes it safe is the claim you challenged, and it does hold. `card_projection` passes the **whole** document dict through and takes `("id", "name", "url")` from each part, so a non-admin's `list_assets` already returns exactly the ids an ambiguity error could name:

```json
{"id": "a1",
 "documents": [{"id": "d1", "name": "Manual", "kind": "link", "url": "http://x"}],
 "parts": [{"id": "p1", "name": "Rinse aid", "url": "http://y"}]}
```

Note what the projection drops — `cost`, `serial_number` — and what it keeps. Part and document ids are deliberately exposed because the dashboard card resolves task→document links with them. So the ambiguity error reveals nothing a non-admin can't already list, on the two services they can actually reach; and on `adjust_part_stock` and the rest they never get as far as resolution.

Worth stating plainly because the two arguments fail in opposite directions: if someone later adds a field to `card_projection`'s part or document whitelist without thinking about it, the "admin gate protects it" reasoning would give false comfort on exactly the two services it doesn't cover.

---

prestomation:
### Fixture ids are real `uuid4`s now (`f427b8a`, `8c51856`)

The seeded e2e ids were readable — `task_fridge_filter`, `asset_water_heater` — which made them easy to grep but roughly a third the length of anything a real install holds: `models.build_task` and `assets.build_asset` both mint `str(uuid.uuid4())` and accept no caller-supplied id. That flattered the screenshots this PR is judged on, since the new id row has to carry a 36-character string, not a 14-character one. `asset_water_heater` is now `e8d76383-4067-415c-975f-2ee73f475fa1`.

It reached further than the JSON: the document blobs are *named* for the asset and document ids, so those files and their directory moved with them, along with the `.gitignore` allowlist naming that directory. Specs go through a new `tests/e2e/fixture-ids.ts` (`TASK.fridgeFilter`) rather than pasting a uuid.

Three things this surfaced that were otherwise invisible:

1. **A drift-guard would have gone silently blind.** `test_integration_fixture_clean.py` scrapes `data-detail-id="…"` out of the capture harnesses to check nothing the tours click has been dropped from the seed. Those now read `${TASK.nozzleUsage}`, matching nothing — it would have passed while asserting nothing, and what it protects is the *soft* walkthrough gate, so no check would have gone red. It resolves references through the module now; verified it still catches a dropped seed record, and additionally catches a mistyped constant.
2. **Constants inside a `page.evaluate` callback don't exist in the browser.** The capture died with `_fixtureIds is not defined`. Threaded through as evaluate arguments.
3. **The shopping-list shot was quietly order-dependent.** From a pristine seed it failed: the mirror syncs on a stock *crossing*, and enabling auto-buy for a part already at its reorder point creates the buy task without crossing anything. It had only ever worked because an earlier suite run happened to move that stock first. The harness now forces the crossing (up one, back one — the part ends on its seeded quantity), so the shot reproduces from a clean container.

Re-verified on the new seed: **1384 unit, 662 frontend, 70 e2e** (local Chromium and CI's headless shell), **162 integration**, and the screenshot capture passing end to end.

Screenshots below are the current ones — full uuids, which fit on one line on every surface:

Task detail page showing the ID row with a full uuid4 and a copy button

Appliance detail page: the appliance's uuid in the Details card, plus a compact uuid under each document and each part

The convention is recorded in `.amazonq/rules/testing-and-workflow.md`.

---

prestomation:
/q review Third pass on the two fixture commits (`f427b8a`, `8c51856`) — please stay critical and lead with anything serious.

These migrate the seeded e2e fixture from readable ids to real `uuid4`s, add `tests/e2e/fixture-ids.ts`, and rewrite ~120 references across 15 test files. Most of that rewrite was mechanical, which is exactly where I'd expect a mistake to hide. Specifically:

1. **Did any reference get missed or mis-mapped?** Every `task_*`/`asset_*`/`part_*` literal should now be a constant from `fixture-ids.ts`, and each constant's value should match the corresponding record in `.storage/home_keeper`. Worth checking the prefix-collision case in particular: `asset_water_heater` is a prefix of both `asset_water_heater_doc_manual_pdf` and the `asset_water_heater_meta_*` ids, and the migration relied on replacing longest-first. A silently truncated id would leave a valid-looking but wrong uuid.
2. **The on-disk blobs.** `ha_config/home_keeper/documents/<asset_id>/<doc_id>__name.pdf` embeds two ids in its path; the directory and both filenames moved. Please confirm the new paths match the ids in the seed exactly, that the `part_` prefix on the part-file blob is still correct, and that `ha_config/.gitignore`'s allowlist still names the directory that now exists (a mismatch means the blobs stop being tracked, and `documents.spec.ts` needs them on disk to serve a signed URL).
3. **The relaxed drift-guard.** `tests/unit/test_integration_fixture_clean.py` now resolves `${GROUP.name}` through `fixture-ids.ts` before comparing against the seed. Is the regex pair (`CONST_DECL`/`CONST_ENTRY`) robust enough, and — more importantly — can the change make the guard *weaker* in any case? The intent is that an unresolvable reference stays in the set under its own text so it still fails. I tested a dropped seed record and a typo'd constant; tell me if there's a third case that now slips through.
4. **`page.evaluate` threading.** Several capture blocks became `evaluate(async (IDS) => {…}, { TASK, PART })`. Please check none still closes over a module import inside the browser callback, and that each block passes exactly the groups it uses.
5. **The shopping-list capture fix in `8c51856`.** It nudges stock +1/-1 to force a reorder crossing. Is that sound, or does it risk a double-fire (two buy tasks, or a duplicate shopping-list line) on a re-run? And does `adjust_part_stock` clamping interact badly with it?

---

prestomation:
/q review Small follow-up push (`ebb5111`) — a docstring rewrap only, no behaviour change. `ruff` went red on `E501` in `tests/unit/test_integration_fixture_clean.py`: swapping the example id for its `${TASK.nozzleUsage}` constant reference pushed the line past 88 columns.

Nothing to scrutinise there, but the third-pass request on `f427b8a` / `8c51856` is still open and is the one that matters — the mechanical migration of ~120 fixture-id references, the on-disk blob paths, the relaxed drift-guard, the `page.evaluate` threading, and whether the +1/-1 stock nudge can double-fire.

I've since self-audited the first two of those and would like the results challenged rather than accepted:

- **All 28 constants** in `fixture-ids.ts` resolve to a real record in `.storage/home_keeper`, are well-formed `uuid4`s, and none is duplicated — so the longest-first replacement did not truncate anything across the `asset_water_heater` / `asset_water_heater_meta_*` prefix collision.
- **Both blobs** sit at `documents/<asset uuid>/<doc uuid>__manual.pdf` and `documents/<asset uuid>/part_<part uuid>__anode-rod-receipt.pdf`, matching the seed, with the `part_` prefix intact; `ha_config/.gitignore`'s allowlist names the directory that now exists, and `git ls-files` still tracks both.

Points 3–5 I have not independently proven, so those are where a fresh pair of eyes is most useful.

---

PR #265, opened 2026-08-29
fix: read the device registry through a version-tolerant shim (HA 2026.9)

### LINKED ISSUE #253: Home Keeper fails against the Home Assistant beta

prestomation:
Fixes #253

## What changed

The nightly `ha-beta` run went red on the **`typing`** job only (`upgrade`, `integration` and `e2e` all passed). Five mypy errors against HA `2026.9.0b1`:

```
devices.py:310: "Collection[DeviceEntry]" has no attribute "values"                    [attr-defined]
devices.py:503: Item "ChildDeviceEntry" of "DeviceEntry | ChildDeviceEntry"
                has no attribute "connections"                                          [union-attr]
coordinator.py:306: Incompatible return value type (got "DeviceEntry | ChildDeviceEntry | None",
                expected "DeviceEntry | None")                                          [return-value]
device_trigger.py:86,139: Argument 1 to "_hk_identifier" has incompatible type
                "DeviceEntry | ChildDeviceEntry"; expected "DeviceEntry"                [arg-type]
```

Both come from one upstream rework of `homeassistant.helpers.device_registry` in 2026.9:

- **Child devices.** A device can now be registered as a child of another, and a child is its own class (`ChildDeviceEntry`) rather than a `DeviceEntry` — it carries only the attributes both kinds share, so no `connections`, `manufacturer` or `model`. `DeviceRegistry.async_get` answers with either kind.
- **`DeviceRegistry.devices`** stopped being an id-keyed mapping and became a collection of entries. Iterating the old one yields **ids**, the new one yields **entries**; `.values()` answers on both today but is deprecated from 2026.9 and removed in 2027.9.

`lint.yml` type-checks against *stable* Home Assistant, where `ChildDeviceEntry` has no name at all, so the union cannot be spelled in an annotation that holds for both cores. Registry devices therefore keep their `dr.DeviceEntry` annotation, and every lookup that can now answer with a child goes through a new **`device_compat.py`**:

| helper | replaces | why |
| --- | --- | --- |
| `resolve_device(registry, device_id)` | `registry.async_get(...)` | one documented place where the `DeviceEntry` annotation approximates the 2026.9 union |
| `all_devices(registry)` | `registry.devices.values()` | asks which shape it was handed instead of using the deprecated spelling |
| `device_connections(device)` | `device.connections` | keys off `parent_device_id`, so it never reads through the child's compatibility shim |

A child device stays a **valid attach target** — Home Assistant links entities to either kind (`Entity.device_entry: AnyDeviceEntry`) — so nothing is filtered out. Reading `connections` off a real `ChildDeviceEntry` would otherwise go through a shim that logs a deprecation warning and stops answering in 2027.9; `device_connections` returns an empty set instead.

`device_compat.py` imports Home Assistant only under `TYPE_CHECKING` (the `options.py` precedent), which keeps it in the pure tier — that is what lets the two registry shapes be exercised with plain fakes rather than a container that can only ever be one core at a time.

### Testing

- `tests/unit/test_device_compat.py` (new, 11 tests) — both registry shapes, child-device resolution, the empty-id guard, and connections on main vs child devices. Confirmed the shape tests fail if `all_devices` iterates naively.
- `tests/unit/test_device_heal.py` — a `ModernFakeRegistry` with the 2026.9 collection shape, asserting the split repair still builds the same merge map from it.
- `pyproject.toml` — `device_compat.py` added to the mutmut `only_mutate` allowlist. **15/15 mutants killed** (100%).
- `tests/unit/test_release_issues.py` — `253` added to the pinned issue set.

Verified locally:

| check | result |
| --- | --- |
| `mypy` vs HA `2026.9.0b1` (py3.14) | ✅ 49 files, no issues (was 5 errors) |
| `mypy` vs HA `2026.8.3` (py3.14) | ✅ 49 files, no issues |
| `pytest tests/unit` | ✅ 1529 passed, 2 skipped |
| integration suite vs `HA_TAG=beta` (2026.9.0b2) | ✅ 167 passed |
| integration suite vs `HA_TAG=stable` (2026.8.3) | ✅ 167 passed |
| `ruff check` / `ruff format --check` | ✅ clean |
| `vale CHANGELOG.md` | ✅ no hits on added lines |

The 2026.9 container log also confirms the `device_registry.devices` deprecation warning is gone.

### Version

Bumped to `0.19.0b1` (`manifest.json` + `const.py`) with a matching CHANGELOG section — `main` was still on the just-shipped stable `0.18.0`, so the post-stable bump was due and the fix needs a version to ship on.

### Not fixed here — two *new* 2026.9 deprecation warnings

The 2026.9 container logs two warnings that HA `2026.8.3` does not emit. Neither fails anything today, both break in **HA 2027.8**, and both name Home Keeper and tell the user to file a bug report:

1. `devices.py:436` — `async_get_or_create(via_device=...)`; use `via_device_id`. Small: `via_device_id` is accepted on 2026.8 too, and `_reconcile_virtual` already corrects the parent link with it right after create.
2. `devices.py:572` — `async_get_device(identifiers=...)`, deprecated "because device identifiers and connections are no longer unique across config entries"; use `async_get_device_by_identifier` / `async_get_device_by_connection` / `async_get_devices`. Larger: those helpers do not exist on 2026.8, and `async_get_devices` returns *all* matches where the old call resolved to one — which is load-bearing for `_resolve_by_snapshot`'s foreign-device preference in the #183 split repair.

Left out to keep this PR to the failure the nightly reported; happy to take either on in a follow-up.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`; no frontend change, so `ci/test-frontend.sh` not applicable)
- [x] `CHANGELOG.md` updated — `(Fixes #253)` in the new `## [0.19.0b1]` section
- [x] Panel UI changed → n/a, no UI change
- [x] New user-facing feature → n/a (bug fix); version still bumped to `0.19.0b1` because `main` sat on the shipped stable

---
_Generated by [Claude Code](https://claude.ai/code/session_01QehZpzrrj2HyjJKNnz9HNo)_

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -89,6 +89,10 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
   tier assert on the option merge rules every write path shares. Keep it that way —
   reach for `hass.<something>` at import time and `tests/unit/test_options.py` stops
   running without the HA harness.
+- `device_compat.py` sits on the same boundary for the same reason: it only ever calls
+  methods on a `DeviceRegistry` it is handed, so its `homeassistant` import is
+  `TYPE_CHECKING`-only and `tests/unit/test_device_compat.py` can drive both registry
+  shapes with plain fakes. Both modules are in the mutmut `only_mutate` allowlist.
 
 ## Datetimes & timezones
 - All datetimes are timezone-aware. Use `homeassistant.util.dt` (`dt_util`) at the
@@ -258,6 +262,18 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
   `PROBLEM` `binary_sensor` (`assets.part_is_low`) for each `assets.part_has_reorder`
   part. Both enumerate via `coordinator.virtual_asset_parts(predicate)` and prune stale
   registry entries by unique-id shape (mirroring the asset-date-sensor cleanup).
+- **Read the device registry through `device_compat.py`, never `DeviceRegistry`
+  directly.** Home Assistant 2026.9 made `async_get` answer with a `ChildDeviceEntry`
+  (a separate class with no `connections`/`manufacturer`/`model`) as well as a
+  `DeviceEntry`, and turned `DeviceRegistry.devices` from an id-keyed mapping into a
+  collection of entries — so iterating it yields ids on one core and entries on the
+  next. `lint.yml` type-checks against *stable* HA, where `ChildDeviceEntry` has no
+  name, so the codebase keeps annotating registry devices as `dr.DeviceEntry` and
+  confines the approximation to that one module: `resolve_device` for a lookup by id,
+  `all_devices` to enumerate, `device_connections` for the one attribute a child
+  lacks. A child device is a valid attach target (HA links entities to either kind) —
+  don't filter them out. Reading any other `DeviceEntry`-only attribute off a resolved
+  device needs a helper there too, not a bare attribute access.
 - `diagnostics.py` provides **both** config-entry and **per-device**
   (`async_get_device_diagnostics`) downloads; the per-device one scopes to the device's
   appliance + its tasks. A device-level *action* was deliberately **not** added (the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.19.0b1]
+
+### Fixed
+
+- **Ready for Home Assistant 2026.9.** That release reshapes the device registry and
+  adds child devices, so Home Keeper reads both shapes and a task or appliance you
+  attached to a device still finds it after the upgrade. (Fixes #253)
+
 ## [0.18.0] - 2026-08-29
 
 ### Added
--- a/custom_components/home_keeper/const.py
+++ b/custom_components/home_keeper/const.py
@@ -8,7 +8,7 @@ PLATFORMS = ["todo", "calendar", "button", "sensor", "binary_sensor", "number"]
 # Frontend panel.
 # PANEL_VERSION is the single source of truth that release.yml validates against
 # manifest.json's "version" (mirrors Pawsistant's CARD_VERSION check).
-PANEL_VERSION = "0.18.0"
+PANEL_VERSION = "0.19.0b1"
 PANEL_URL_PATH = "home-keeper"  # sidebar route -> /home-keeper
 PANEL_STATIC_URL = "/home_keeper_panel"  # static path that serves the JS bundle
 PANEL_JS_FILENAME = "home-keeper-panel.js"
--- a/custom_components/home_keeper/coordinator.py
+++ b/custom_components/home_keeper/coordinator.py
@@ -31,6 +31,7 @@ from .const import (
     EVENT_TASK_OVERDUE,
     OPTION_ONE_OFF_RETENTION_DAYS,
 )
+from .device_compat import resolve_device
 from .options import current_options
 from .reconcile import buy_source
 from .store import HomeKeeperStore
@@ -312,9 +313,7 @@ class HomeKeeperCoordinator(DataUpdateCoordinator[dict[str, dict[str, Any]]]):
         Single source of truth for "is this an existing device we can merge onto?"
         so the DeviceInfo and name-prefix decisions can't drift apart.
         """
-        if not device_id:
-            return None
-        return dr.async_get(self.hass).async_get(device_id)
+        return resolve_device(dr.async_get(self.hass), device_id)
 
     def task_uses_existing_device(self, task: dict[str, Any]) -> bool:
         """True when the task's per-task entities merge onto an existing device.
--- a/custom_components/home_keeper/device_compat.py
+++ b/custom_components/home_keeper/device_compat.py
@@ -4,11 +4,17 @@ Home Assistant 2026.9 reshaped two parts of ``homeassistant.helpers.device_regis
 that Home Keeper leans on:
 
 * **Child devices.** A device can now be registered as a *child* of another, and a
-  child is its own class (``ChildDeviceEntry``) rather than a ``DeviceEntry`` — it
-  carries only the attributes both kinds share, so no ``connections``, no
-  ``manufacturer``, no ``model``. ``DeviceRegistry.async_get`` answers with either
-  kind, and ``Entity.device_entry`` accepts either, so a child device is a perfectly
-  good thing to attach a task or an appliance to and Home Keeper keeps taking one.
+  child is its own class (``ChildDeviceEntry``) rather than a ``DeviceEntry``. Both
+  derive from ``BaseDeviceEntry``, so a child still carries ``id``, ``identifiers``,
+  ``name``, ``name_by_user``, ``area_id``, ``labels``, ``config_entry_id`` and
+  ``disabled_by`` — every attribute Home Keeper reads off a device except one. What a
+  child drops is the hardware description: ``connections``, ``manufacturer``,
+  ``model``, ``model_id``, ``hw_version``, ``sw_version``, ``serial_number``,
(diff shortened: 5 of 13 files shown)


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first without withholding minor ones. This adds `custom_components/home_keeper/device_compat.py` as the single chokepoint for device-registry reads, because HA 2026.9 made `DeviceRegistry.async_get` return `DeviceEntry | ChildDeviceEntry` and turned `DeviceRegistry.devices` from an id-keyed mapping into a collection of entries — and the codebase must type-check against *stable* HA, where `ChildDeviceEntry` has no name.

Topics I want scrutinized:

- **Correctness across both cores.** `all_devices` discriminates the two `devices` shapes with `isinstance(devices, Mapping)`. Is that a safe discriminator for 2026.8's `ActiveDeviceRegistryItems` (a `UserDict`) and 2026.9's `_DeprecatedDeviceRegistryItemsView`, and could a future shape defeat it? `device_connections` keys off `parent_device_id` to avoid reading through `ChildDeviceEntry.__getattr__` — is there any main-device entry that could carry that attribute, or any child that could not?
- **The deliberate type lie.** `resolve_device` is annotated `-> dr.DeviceEntry | None` but may return a `ChildDeviceEntry` at runtime. Are there call sites downstream of `coordinator._existing_device` / `device_entry_for_device_id` / `virtual_asset_parts` that read a `DeviceEntry`-only attribute (`connections`, `manufacturer`, `model`, `entry_type`, `via_device_id`, `configuration_url`, `composite_device_id`) off a device that could now be a child, and would therefore trip the same 2027.9 removal this PR is trying to get ahead of? I checked `notifier.py` and `problem_sync.py` (only `labels`/`area_id`) but a second pass is welcome.
- **Behavior on 2026.9.** Accepting a child device as an attach target instead of filtering it out — right call? `devices.py:353` and `async_prune_orphaned_devices` iterate `async_entries_for_config_entry`, which is main-devices-only; is there a path where a child device we link entities to escapes cleanup?
- **Maintainability.** Is `device_compat.py` the right boundary, or should this live in `devices.py` next to `_supports_kwarg`? Are the three helper names and the module docstring enough for the next person to know when a new registry read needs a helper?
- **Performance.** `all_devices` materializes a list of every device in the registry on each setup; it runs once in `async_heal_split_device_ids`. Acceptable, or worth a generator?
- **Test quality.** `tests/unit/test_device_compat.py` drives fakes, not the real HA classes. Do the fakes model the two eras faithfully enough for the tests to mean anything, and is anything important asserted only by construction?
- **HA best practices.** These helpers are sync and call `@callback` registry methods but are not themselves `@callback`-decorated (the decorator would need a runtime `homeassistant` import and cost the module its pure/`TYPE_CHECKING`-only status). Is that trade-off defensible?

Also flagged in the PR body but deliberately not fixed here: two *new* 2026.9 deprecation warnings (`async_get_or_create(via_device=...)` and `async_get_device(identifiers=...)`), both breaking in HA 2027.8. Tell me if you think either belongs in this change rather than a follow-up.


---

prestomation:
Thanks — detailed review. Two of the three points are addressed in `0e944aa`; the blocking one is a false positive and I'm pushing back on it with evidence.

## ❌ Blocking finding — `_hk_identifier` reading `device.identifiers`

> **Problem:** This reads `device.identifiers`, which is a `DeviceEntry`-only attribute. A `ChildDeviceEntry` passed here will fail on HA 2027.9 when the compatibility shim is removed.

This is not correct, and the proposed guard would be a regression. `identifiers` is a **real attrs field on `ChildDeviceEntry`**, declared on the shared `BaseDeviceEntry` base — it is not served by the compatibility shim and there is nothing for 2027.9 to remove.

Checked against the installed `homeassistant==2026.9.0b1` wheel rather than by reading:

```
ChildDeviceEntry attrs fields: ['_cache', 'area_id', 'config_entry_id',
  'config_subentry_id', 'created_at', 'disabled_by', 'id', 'identifiers',
  'labels', 'modified_at', 'name', 'name_by_user', 'parent_device_id']

'identifiers' declared on BaseDeviceEntry? True
'identifiers' in the deprecation-shim set? False
shim set = ['configuration_url', 'connections', 'entry_type', 'hw_version',
            'manufacturer', 'model', 'model_id', 'serial_number',
            'sw_version', 'via_device_id']

child.identifiers -> {('home_keeper', 'task-abc')}
'identifiers' in BaseDeviceEntry.__slots__ -> True
child.connections -> AttributeError: 'ChildDeviceEntry' object has no attribute 'connections'
```

The shim is `ChildDeviceEntry.__getattr__`, and `__getattr__` only runs when normal attribute lookup **fails**. `identifiers` is in `__slots__`, so lookup succeeds and `__getattr__` is never reached. `_CHILD_DEVICE_COMPAT_ATTRS` (the exact set the shim serves, quoted above) is the definitive list of names at risk in 2027.9, and `connections` — the one this PR wraps — is the only member of it Home Keeper reads.

Applying the suggested guard would also change behavior for the worse: `_hk_identifier` returning `None` for a child device would discard a Home Keeper identifier the child genuinely carries, so a task on such a device would silently fall through to the `device_id` filter branch.

**No change made here.** If you have a source reference showing `identifiers` behind the shim, I'll happily re-open it.

Same reasoning retires the follow-on test suggestion — a `test_child_devices_have_no_identifiers` would assert something that isn't true of `ChildDeviceEntry`.

## ✅ Pointer comment in `devices.py`

Added above the import, since that file is the most likely place for a new registry read to appear:

```python
# Every device-registry *read* goes through device_compat, never straight at the
# registry: Home Assistant 2026.9 changed both what ``async_get`` can answer with and
# what iterating ``registry.devices`` yields. Read that module before adding one here.
from .device_compat import all_devices, device_connections, resolve_device
```

The durable form of the rule also lives in `.amazonq/rules/architecture-and-code.md` → *Entities & devices*, which is what both Amazon Q and Claude auto-load.

## ✅ Documentation fix prompted by the finding

The review misread my own wording, so that wording was the real defect. "Carries only the attributes both kinds share" invited exactly this inference. The `device_compat.py` docstring now names both halves explicitly — the shared fields a child keeps (`id`, `identifiers`, `name`, `name_by_user`, `area_id`, `labels`, `config_entry_id`, `disabled_by`) and the hardware-description names its `__getattr__` serves through the shim — so the next reader doesn't have to go check the wheel.

## Confirmations noted

Agreed and useful: `parent_device_id` as the child discriminator has no collision risk; `async_entries_for_config_entry` is main-devices-only, so a child we merely link entities to correctly never enters `async_prune_orphaned_devices` (we don't own it); a generator for `all_devices` would not avoid the materialization the merge-map builder needs; and both new 2026.9 deprecations belong in a follow-up.

CI on the previous head was green across lint (including mypy), test, mutation, HACS, e2e, coverage and docs preview.


---

prestomation:
/q review Follow-up review of `0e944aa`, which is documentation and comments only (no behaviour change) — please be critical and skeptical, most serious issues first, and don't withhold minor ones.

Two things I'd like you to check specifically:

1. **Re-examine your previous blocking finding.** I did not apply the `parent_device_id` guard you proposed for `device_trigger.py:_hk_identifier`, because `identifiers` is a real attrs field on `ChildDeviceEntry` (declared on `BaseDeviceEntry`, present in `__slots__`, and absent from `_CHILD_DEVICE_COMPAT_ATTRS`) rather than something the 2027.9 deprecation shim serves. The evidence is in my previous comment, taken from the installed `homeassistant==2026.9.0b1` wheel. If you still believe the guard is needed, please cite the specific line of HA source that puts `identifiers` behind `ChildDeviceEntry.__getattr__`; otherwise please confirm the finding is retracted.

2. **Is the new docstring accurate?** `device_compat.py` now enumerates which attributes a child device keeps and which it drops. Please verify that list against HA 2026.9's `BaseDeviceEntry` / `ChildDeviceEntry` / `_CHILD_DEVICE_COMPAT_ATTRS` — a docstring that lists the wrong names is worse than the vague sentence it replaced, so I'd rather it be checked than trusted.

Also still open from the first round if you want to weigh in: whether the two deferred 2026.9 deprecations (`async_get_or_create(via_device=...)` and `async_get_device(identifiers=...)`) belong in this PR or a follow-up.


---

PR #266, opened 2026-08-29
feat(notifications): channel and urgency, on Android and iOS

### LINKED ISSUE #255: Feature Request: Notification Channels

prestomation:
Fixes #255

## What changed

Two new fields on a saved notification — **Notification channel** and **Urgency** — plus the three worked automations the other half of #255 asked for.

The Settings → Notifications card showing a walk on the Chores channel at High urgency, with helper text under each new field

### Is this an Android-only thing?

Partly, and the docs now say so rather than leaving it implied. Channels are Android's model; iOS has none. So the two fields are stored **platform-neutral** and expanded into both vocabularies at payload-build time by a new `notifications.payload_data`. Each app ignores the keys it does not know, so there is no per-target branching in `notifier.py` and the panel never asks which phone the household owns.

| Urgency | Android | iOS |
| --- | --- | --- |
| `quiet` | `importance: low` | `push.interruption-level: passive` |
| `normal` (default) | *nothing emitted* | *nothing emitted* |
| `high` | `importance: high`, `ttl: 0`, `priority: high` | `push.interruption-level: time-sensitive` |
| `critical` | `importance: max`, `ttl: 0`, `priority: high` | `push.interruption-level: critical`, `push.sound: {name: default, critical: 1, volume: 1.0}` |

The channel name goes out as Android `data.channel` and iOS `data.push.thread-id`. `data.group` stays the literal `home_keeper` — the reporter said grouping already works, and `test_build_notification_walk_actions_and_tag` pins that string as a payload contract.

`normal` is deliberately absent from both mapping tables, so **an unconfigured notification sends byte-for-byte what it sent before these fields existed**. That is asserted, not assumed (`test_payload_data_unconfigured_is_what_it_always_was`).

Two caveats are surfaced in the field helper text and the README instead of being worked around, because both are platform behaviour: Android fixes a channel's settings when it creates the channel (so raising the urgency later moves nothing — rename it), and iOS demotes a critical alert unless Critical Alerts is allowed for Home Assistant.

### The other half of #255: re-sending until it's done

Left as an automation, per your call on the issue — the Notifications page now carries three worked examples so it's a recipe to point at rather than a paragraph retyped per issue: re-sending every couple of hours, holding reminders back until you're home, and a tight window at Critical urgency. `home_keeper.notify` sends nothing when nothing matches (`notifier.py` short-circuits an empty queue), which is what makes a `time_pattern` trigger safe, and the docs now say so.

### Notes for review

- **No new API surface.** `notifications` is already an opaque `list` in `SET_OPTIONS_SCHEMA`, the options flow never renders notifications, and `api_surface.OPTIONS` indexes the option key rather than its sub-fields. No `services.yaml` / `strings.json` / `translations/` change. Per your answer on scope, `home_keeper.notify` gained **no** override fields.
- **No migration.** `options.current_options` re-normalizes on every read, the same route that let `exclude_*` and `sync` ship without one.
- The screenshot step is now scoped to the `#hk-notifications` card rather than `fullPage` — after the Settings rail redesign a full-page shot renders these fields too small to read. Only that one PNG is committed; a local capture re-renders the whole set with different font rasterization, which would be noise.
- `.amazonq/rules/architecture-and-code.md` records the convention this establishes (platform-neutral storage, expanded at build time; the default must be indistinguishable from not having the feature).

## Checklist

      embedded above with an HTML image tag
      step through it

**On the walkthrough box:** left unchecked deliberately. This adds two fields to an existing form rather than a new surface, so the tour needs no edit. Worth knowing that `walkthrough.capture.ts` doesn't visit Settings → Notifications at all today — happy to add that step if you'd rather it did, but it felt like a separate change.

### Verification

`pytest tests/unit` (1544 passed), `ci/test-frontend.sh` (736 passed), the full Playwright suite (189 passed, including a new `settings.spec.ts` assertion on the notification form at all three viewports), `ruff check` + `format --check`, and `vale` clean on every line this branch adds. Mutation: **91.5%** Python and **100%** frontend against the 80% gate — the 25 Python survivors are all `dict.get(k, False)` → `get(k, None)` style equivalents on pre-existing lines that the one-line `data:` edit pulled into scope.

`mypy` could not run here (this container is on Python 3.11 and Home Assistant needs 3.14+, so every error was a missing import); the changed pure modules typecheck clean in isolation and CI runs the real thing.


prestomation:
/q review Please be critical and skeptical, and lead with the most serious problems — but don't withhold the minor ones either.

This adds two fields to a saved notification (`channel`, `urgency`) and expands them into both Android and iOS mobile-app payload keys in a new `notifications.payload_data`, which the three existing payload builders now share. Areas I'd most like scrutinised:

**Correctness of the payload contract.** Are the Home Assistant companion-app keys right for each platform and each urgency? Specifically: Android `channel` / `importance` (`low`/`high`/`max`) plus `ttl: 0` + `priority: high`, and iOS `push.thread-id` / `push.interruption-level` (`passive`/`time-sensitive`/`critical`) plus `push.sound: {name, critical, volume}`. Is emitting both platforms' keys on one payload actually safe, or does either app choke on the other's keys? Is there a key I should be setting and am not, or one that conflicts with the existing `tag` / `group` / `actions`?

**The "unchanged by default" claim.** `normal` is deliberately absent from both mapping tables so an unconfigured notification is byte-identical to before. Please try to break that — including via `normalize_notification` on stored data that predates the fields, junk values, and the `persistent_notification` target, which receives the same payload and ignores `data`.

**Error paths and normalization.** `str(raw.get("channel") or "").strip()` and the `urgency in URGENCIES` clamp: anything that gets through and reaches a phone as something unintended? Note Critical bypasses Do Not Disturb, so a value that shouldn't reach `critical` reaching it is the worst failure here.

**Maintainability.** Is `payload_data` the right seam, or should the mapping live in a table/dataclass? Is the module still honestly "pure" (no HA imports, no blocking I/O — `tests/unit/test_notifier_blocking.py` guards the latter)?

**HA best practices.** Anything off about how this rides `set_options` rather than a dedicated service, and whether `home_keeper.notify` should have gained per-call overrides (I deliberately left it without them).

**Test quality.** The mutation gate passes at 91.5% Python / 100% frontend, but please check whether the new assertions actually pin behaviour or just shape — particularly the exact-dict payload assertions and the new `settings.spec.ts` e2e check.

---

prestomation:
Pushed `b47c776`, a prose-only follow-up: the `vale` gate failed on the first push and is green now.

Worth recording why, since AGENTS.md already warns about it and this is a clean example. CI's pinned vale binary flagged a `VerbTricolon` that a local Vale 3.9.1 run over the same file did not report. The rule's pronoun arm is `(?:[Tt]hey|[Ww]e|[Yy]ou) \w+ [^,]+, \w+ [^,]+, and \w+ [^.!?]+` — only the *last* segment excludes sentence-ending punctuation, so `[^,]+` walks straight through a period. My paragraph read "…until **you** are home**,** belongs in … setting. That keeps the timing yours to change**, and** `home_keeper.notify` does nothing…", which is three verbs in series spanning two sentences. Rewriting it comma-free removes the whole tricolon family rather than shuffling one clause.

Since local vale had already missed one, I stopped trusting it and matched the rules' own regexes against the added blocks directly (AGENTS.md's advice for when a run matters). That turned up two more: a `MicDrop` on a sentence ending "…Do Not Disturb." — the rule reads it as a bare `Not X.` fragment — and `FormalTransitions` on "therefore". Both fixed. Same `MicDrop` shape had already bitten `notify.channel_help`, so that string is reworded in all 16 locales.

### On your two suggestions

**`"critical": 1` as an integer** — correct as written. The companion app's [critical-notifications docs](https://companion.home-assistant.io/docs/notifications/critical-notifications/) specify exactly `sound: {name: "default", critical: 1, volume: 1.0}` with the integer, not a boolean. I checked that against the docs while designing the mapping rather than inferring it.

**A warning log for a rejected urgency** — I'd rather not, and the asymmetry with the target rejection is deliberate. `normalize_notification` runs on *every* `current_options` read, so a persistently bad stored value logs on every panel load and every coordinator refresh, not once. The targets warning accepts that cost because a silently dropped notify target changes *who gets told* and is security-relevant. An urgency clamp is neither: the panel can only emit the four valid values, so a bad one means hand-edited storage, and the fallback direction is the safe one (toward quieter, never toward Critical). Happy to add it if you'd rather have the diagnostic.

---

/q review Second pass, on the same critical and skeptical footing — the code is unchanged since your last review, so please focus on what a first pass most often misses rather than re-confirming what you already checked.

Specifically: is there an ordering or precedence problem when a notification sets a channel *and* an urgency that disagree with a channel the phone already created? Does `payload_data`'s `push` sub-dict interact badly with anything the companion app expects at the top level? And on `tests/e2e/tests/settings.spec.ts` — the new assertion reads input values via `evaluateAll`; is that stable across the three viewports, or is there a race where the form has rendered but not yet populated?

---

prestomation:
Closed, because this work shipped in #277.

`notifications.py` on main carries the same design. It stores one platform-neutral `channel` and `urgency` pair, and `payload_data` expands that pair into the Android `importance` keys and the iOS `interruption-level` keys. The CHANGELOG records `(Fixes #255)`. The 29 merge conflicts on this branch are all against main's version of these lines.

---

PR #269, opened 2026-08-29
feat(panel): edit beside the page you pressed Edit on, and stop the Settings rail jumping

prestomation:
Two desktop panel changes.

## 1. Edit opens beside the page you pressed it on

Edit navigated: from a task's or appliance's own page it went back to the owning list
and opened the drawer there, so the schedule, the notes, the completion history and
the parts all went off screen at the moment those values were being changed.

The form now opens as a column beside that page, and the location does not move.
`_openEdit` / `_openEditAsset` open in place whenever the form belongs to the page
that is showing (`_editsThisPage`); a cross-view edit still navigates, because a form
mounts on its own view. Three consequences:

- `_mountDrawerForm` moves above `_hydrate`'s early return for a task detail. That
  return is why Edit on a task page would otherwise open an empty drawer.
- A detail page is exempt from the drawer's dimming (`.hk-wrap:not([data-detail])`):
  there the page is the subject of the form, not a list to recede behind it.
- An appliance is read beside its list, and a form beside both is a third column that
  does not fit, so above 1150px the list steps aside while the form is open — the move
  it already makes below 1000px. Below 1150px nothing changes: the drawer is already a
  sheet over the page.

The drawer's History button is dropped on the page it would lead to, and its Delete
replaces that page with the list before deleting, the way the page's own Delete
already does.

A task edited beside its own page: the form in a drawer on the right, the task's schedule, notes and completion history still readable on the left

An appliance edited beside its own page: the form on the right, the appliance's parts still on screen, the appliance list stepped aside

The committed shots of a form opened from a detail page were recaptured, since that is
what changed. Here is the managed-task one, where the locked fields and the footer are
both visible:

A managed task edited beside its page: the name field omitted because the integration owns it, Delete in the drawer footer, History absent

## 2. The Settings rail no longer jumps to the top

Clicking a section in the rail rebuilt the whole panel to move one mark. `_render`
replaces the entire shadow tree, including the card the rail's smooth scroll was aimed
at, and every `ha-form` in the replacement renders after the swap — so the scroll ran
against a page whose height was still settling and could be clamped to the top on the
way, read as a jump followed by a slow crawl back down.

A section change is a lateral step along one page, so it is now patched rather than
rebuilt: `_applyLocation` recognises a section-only move and hands it to
`_patchSettingsSection`, which updates the current rail entry, `.hk-sec-current`, the
layout's `data-section` and the narrow back bar (rebuilt and rewired). Everything else
still renders.

Two smaller things came with it: scrolls the panel starts itself now honour
`prefers-reduced-motion` (a `behavior` passed to `scrollIntoView` overrides the
stylesheet), and opening a section from the narrow index scrolls to the top now that
nothing else moves the scroll on that drill-in.

The Home Keeper Settings tab: the section rail on the left with its state marks, and the six sections beside it

**On reproducing the jump.** The seeded e2e container never shows it: its Settings page
is small and its `ha-form`s render inside the same task as the swap, so the height
never dips at layout time (measured — `scrollHeight` holds at 2078 across a rail
click). What does reproduce there is the mechanism, a rail click rebuilding the tree
the scroll was aimed at, and that is what the tests pin. A real instance, with many
more entities behind each selector, is where the settling window is wide enough to see.

## Tests

- `frontend/test/detail-edit.test.js` — the form mounts beside a task page and an
  appliance page without navigating, the drawer's History is dropped on the page it
  points to, closing returns to the page rather than a list, and a cross-view edit
  still navigates.
- `frontend/test/settings-nav.test.js` — node identity across a section change (a
  rebuild would detach every node held from before it), the marks moving, the back bar
  rebuilt and still wired, and a cross-view navigation still rendering. Two of its four
  cases fail without the fix.
- `tests/e2e/tests/detail-edit.spec.ts` — against a real Home Assistant: the page stays
  and the URL holds while the form is open, the appliance list steps aside and comes
  back, and a save updates the page in place.
- `tests/e2e/tests/settings.spec.ts` — a rail click leaves the rendered page standing
  (fails without the fix) and the scroll never falls below where it started.
- Three tests in `panel.test.js` stood in for HA's router after an edit navigation.
  There is no navigation now, so the stand-in is gone.

Run locally: `pytest tests/unit` (1531 passed), `bash ci/test-frontend.sh` (740
passed), the full Playwright suite across all three widths (187 passed), both
walkthrough tours, the screenshot capture, and `tsc --noEmit`.

## Checklist

- [x] `CHANGELOG.md` updated — a new `## [0.19.0b3]` section, because `0.19.0b2` is
      already a published tag and `changelog-release-gap` blocks edits to it. No issue
      to link.
      embedded above with HTML `img` tags
- [x] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` steps through
      Edit opening beside a task's page (both tours re-run locally)
- [x] New user-facing feature → version bumped to `0.19.0b3` (`manifest.json` +
      `const.py`), `preview-release` label requested


prestomation:
/q review Please be critical and skeptical, most serious issues first, and don't withhold minor ones.

This replaces a full `_render()` with an in-place DOM patch for one navigation case (moving between Settings sections). Scrutinise in particular:

- **Correctness of the fast path.** `_applyLocation`'s `sectionOnly` guard: is there any state a full render would have reset that the patch now leaves stale (open forms, `_noteEdit`, pending edits, `_liveHassEls`, markdown previews, focus, `inert`/drawer modality)? Is the pre-mutation evaluation of `sectionOnly` right, and can the guard let through a case it should not — or reject one it should allow?
- **Listener and node lifecycle.** `_patchSettingsSection` removes and re-inserts the narrow back bar and rewires `#settings-back`. Can that leak listeners, double-wire, or leave a dead button on any path (repeated moves, back to the index, a deep link, first paint, a `_render` that follows a patch)?
- **The diagnosis itself.** The claim is that the rebuild is what put a jump in front of the rail's smooth scroll. I could not reproduce the visible jump in the seeded e2e container (page height held steady across the swap) — tell me if a different mechanism is more likely (HA's router, scroll anchoring, focus restoration, sticky rail) and whether this fix would actually address what the user reported.
- **Test strength.** Do the jsdom and Playwright tests assert something a wrong implementation could not pass, or are they tautological? Anything untested that should not be (the `prefers-reduced-motion` branch, the index-row scroll-to-top)?
- **HA best practices and accessibility.** `aria-current` handling, focus after a patched navigation, and whether patching around the panel's "route is the single source of truth" rule weakens it.

---

prestomation:
/q review Please be critical and skeptical, most serious issues first, and don't withhold minor ones. Note the previous review described a change that isn't in this diff (a "Settings feature", a bump to 1.0.2, a `SETTING_TYPE` constant) — please review the actual diff on the current head.

The PR now carries two changes to `custom_components/home_keeper/frontend/src/panel.ts`. Scrutinise in particular:

**Editing beside the detail page** (`_openEdit` / `_openEditAsset` / `_editsThisPage` / `_mountDrawerForm`)
- **Correctness**: `_editsThisPage` returns true whenever no detail is open, and the caller also checks the view — is any call path left where a form mounts on a page it does not belong to, or where the drawer opens empty? Is the cross-view branch (`_pendingEdit` + `_navigate`) still reachable, and correct if it is?
- **Lifecycle**: `_mountDrawerForm` moved above `_hydrate`'s early return for a task detail. Does anything it relies on get wired later, or get wired twice now? Any listener or `_liveHassEls` leak?
- **State the old navigation used to reset**: `_applyLocation` cleared `_edit`/`_assetEdit`/`_noteEdit` on every edit before. Editing now happens without a location change — is there stale state (an open note editor, an in-flight upload, `_drawerOpener`, drawer `inert`/modality) that no longer gets cleared?
- **Interactions newly possible**: the detail page's own actions (Done, Archive, Delete, the note editor) are live beside an open form. Any way to lose data or write a stale record — e.g. archive then save?
- **Layout**: the appliance master pane is hidden above 1150px via `.hk-wrap[data-detail="asset"]`; the dimming rules now key off `.hk-wrap:not([data-detail])`. Any width or state where that leaves the page dimmed, the list wrongly hidden, or a third column?

**The Settings rail patch** (`_patchSettingsSection`)
- Can the in-place patch leave stale state a full render would have reset? Can the rebuilt back bar double-wire or leave a dead `#settings-back`?
- Is the diagnosis right? I could not reproduce the visible jump in the seeded e2e container (page height held steady across the swap) — say so if a different mechanism is more likely.

**Tests**: do the jsdom and Playwright tests assert something a wrong implementation could not pass? Anything untested that should not be — the `prefers-reduced-motion` branch, the index-row scroll-to-top, the drawer Delete's navigate-then-delete?

**HA best practices and accessibility**: focus handling when the form opens over a page that stays interactive, `aria-current`, and whether opening a form without a URL change weakens the panel's "route is the single source of truth" rule.

---

PR #270, opened 2026-08-30
fix(panel): UX polish — dialog titles, one set of button weights, and six smaller things

### LINKED ISSUE #262: Panel UX polish: dialogs lost their titles to an HA API change, plus chip contrast and inconsistent button weights
A UX review during the #261 redesign turned up a set of issues that **predate that branch**, so they were left out of it rather than widening a restyle. Collecting them here.

They are independent of each other — happy for this to be split up or cherry-picked.

## 1. Every dialog title silently stopped rendering (highest value here)

`ha-dialog` sets its title from a `slot="headerTitle"` in current Home Assistant. It no longer reads the `heading` attribute, which is what the panel sets — so the attribute lands on the element, nothing consumes it, and the dialog renders as a bare ✕ over its body.

Verified in a running container. The completion dialog's element carries `heading="Log: Replace fridge filter"` and its shadow root contains no title node at all:

```
lightChildren: ["DIV", "HA-DIALOG-FOOTER[slot=footer]"]
slotNames:     ["header", "headerNavigationIcon", "headerTitle",
                "headerSubtitle", "headerActionItems", "(default)", "footer"]
renderedTitle: null
```

The practical cost: completing a task opens a dialog over a scrimmed list with **no way to tell which task you are completing**. The move-date dialog has the same problem.

Two call sites, both on `main`:

- `panel.ts` — the completion dialog, `setAttribute('heading', …)`
- `panel.ts` — `dialog.setAttribute('heading', t('completion.moveDate'))`

Fix is a `<span slot="headerTitle">` in place of the attribute. Worth a test: this is exactly the class of breakage `AGENTS.md` describes under "Anything resting on an HA framework contract needs an integration-level assertion" — the strings are still correct, still translated, and still asserted by anything that reads the attribute, so nothing failed when they stopped being shown.

## 2. Two chips still put white text on a mid-tone fill

#261 fixed the two worst (`hk-orphaned` at 1.96:1, `hk-archived` at 1.88:1) by moving them onto the `--hk-*-soft` / `--hk-*-ink` pairing. Two remain, both under the 4.5:1 a label needs:

| chip | container | measured |
|---|---|---|
| `ha-assist-chip.hk-managed` — "Managed by Battery Notes" | `--info-color` | **3.08:1** |
| `ha-assist-chip.hk-comp-connected` — "Connected" | `--success-color` | **3.30:1** |

Same fix as the other two: a soft container with an ink label. `--hk-ok-soft` is already defined and unused.

## 3. The provenance chips are the loudest thing in a task row

"Managed by Battery Notes", "Managed by Pawsistant" and "Auto-synced" are painted in solid `--hk-accent` — the same fill as "Add task". They are not interactive, and in the task list the eye lands on the blue lozenge before the task name. The intended reading order (name → meta → chips → status → Done) is inverted on exactly the rows that have a provenance chip.

Suggestion: make them the same outlined-neutral chip as "Garage water heater", with the integration's icon carrying the identity, and keep solid fills for things you can press. (#261 established that convention for the row's status pill vs its Done button; this would extend it.)

## 4. Destructive and cancelling actions are drawn inconsistently

- Task detail: Done (filled), Edit (filled), **Delete (filled red)** — three equal-weight fills, the destructive one loudest.
- Settings → Profiles: each card's **Delete is solid blue**, ~20px from a solid blue "Add profile". Same verb, opposite colour from the detail page, in the colour that elsewhere means "confirm".
- Completion dialog footer: "Skip details" (tonal) / **"Cancel" (solid blue)** / "Mark done" (solid blue) — Cancel at confirm weight, between two affirmative actions.
- "Remove orphaned tasks" — a bulk delete — is the brightest button on the task list, inside a warning banner.

Suggestion: settle on three weights (filled primary / tonal secondary / text tertiary) plus one destructive treatment, and apply it once. Cancel is always tertiary; Delete is never the only filled button in its group.

## 5. A raw device UUID is shown as a chip

A task whose `device_id` is not in the device registry falls back to rendering the id: `5ff1f1bb41a19a763aa4ab750cd37c97`. At ~1000px it is cut mid-string by the chip's own border with no ellipsis. It should show the device's friendly name or not render.

## 6. Tree children look permanently selected

In Appliances → Tree, a child card gets a light-blue fill plus a blue left rail — the same treatment `#261` uses for *selection*, so every child reads as selected. Indentation plus a neutral rail would carry the hierarchy and leave blue meaning "selected".

Related: the collapse caret sits at the far right of the parent row, a long way from the name it collapses.

## 7. Smaller things

- **Back is drawn two ways.** A filled pill on the task detail, a plain text link on the appliance detail. Back is never a primary action.
- **A zero-count filter chip is still clickable**, landing on "No tasks match this filter" with no way out but the chip you came from. Dimming a zero chip, or putting a "Show all" action in the empty state, would close the loop.
- **Group by Area surfaces completed items inline.** A completed one-off sits mid-list with only a small "Completed" pill — no dimming, no strikethrough — where Group by Status tucks it into a collapsed section. Changing the grouping shouldn't change what appears to be in your active list.
- **The welcome card has two dismissals** (✕ and "Got it"), and "Got it" is a third solid blue button in the first screen.
- **History row controls are icon-only** (move-date / edit / delete) with no labels, and the destructive one is the same weight as the other two.
- **Seconds precision leaks into user-facing time.** "hh 12 : mm 00 : ss 00" in the completion dialog and "7/1/2026, 1:00:00 PM" on the detail page — machine formatting for "when did you change the filter".
- **Meta sentence case drifts**: "every 12 months after completion" everywhere, but "Every 300 h of use, or every 6 months" on a meter task.

## Not included

Two things the same review raised that are deliberate calls rather than bugs: the 268px appliance master pane, and the two rows of filter chips at phone width.

prestomation:
Fixes #262

## What changed

Six of the seven items from the issue. **Item 6 (the appliance tree) was reverted at
the maintainer's request** — the tree keeps its original look, and the rest of the
pass stands. It was the one self-contained item, so it came out cleanly.

Two of the remaining items turned out not to be style opinions at all, but Home
Assistant API changes that had gone by unnoticed — and the reason nothing noticed is
the same in both cases: **the panel was still setting an attribute the element had
stopped reading**, so the markup stayed correct, the strings stayed translated, and
every existing assertion kept passing.

- **`ha-dialog` no longer takes its title from `heading`.** It renders a
  `headerTitle` slot. Both dialogs opened as a bare ✕ over their body.
- **`ha-button` reads neither `raised` nor `destructive`.** It extends Web Awesome's
  `Button`, whose observed attributes are `appearance`, `variant` and `size` only —
  I checked `observedAttributes` on the live element in the e2e container. So all
  twelve `raised` buttons *and* every bare one rendered at the default accent fill.
  That single fact is the whole of item 4: Done, Edit, Cancel and "Remove orphaned
  tasks" weren't four separate mistakes, they were one. `--mdc-theme-primary` is dead
  for the same reason, which is why Settings' Delete was blue instead of red.

`utils.ts` now owns four button weights in HA's own vocabulary, measured against
rendered pixels in the container rather than eyeballed:

| weight | attributes | label vs its fill |
| --- | --- | --- |
| `primary` | *(none — HA's default)* | 3.26:1 † |
| `secondary` | `appearance="filled"` | 6.02:1 ‡ |
| `tertiary` | `appearance="plain" variant="neutral"` | 6.49:1 |
| `danger` | `appearance="plain" variant="danger"` | 7.04:1 |
| `danger-primary` | `variant="danger"` | 4.59:1 |

† HA's own filled-button pairing, unchanged across HA itself.
‡ Only with the ink override — HA's tonal label on its own tonal fill measures
**2.85:1**, which is what the `.done-btn::part(base)` rule had been working around one
button at a time. It is keyed off `[data-hk-weight="secondary"]` now, so every tonal
button gets it rather than the one that was fixed by hand.

Cancel is always tertiary. A destructive action takes the weight of the *surface's*
purpose, not of the thing it destroys — so a solid red fill appears only on the
confirm scrim, whose whole job is the deletion.

## Before / after

Both columns were captured from the same container on the same day; the only
difference is the code. The "before" column is
[`1c0213a`](https://github.com/prestomation/ha-home-keeper/commit/1c0213af9ed3fecbfaac19f44dbaeaf49be3a4d9),
a re-capture of an **unmodified** tree — 68 of the 104 committed shots already differed
from it, because the seeded tasks carry fixed dates and every relative age ("120 days
overdue") drifts with the clock. Comparing against the old committed images would have
shown that drift as if it were my change.

### 1 — Dialog titles (items 1, 4)

Before: no title at all, and "Skip details" / "Cancel" / "Mark done" as three
identical fills. After: the task's name, and three distinct weights.

| Before | After |
| --- | --- |
| Completion dialog with no title and three identical blue footer buttons | Completion dialog titled Log: Replace fridge filter, with tonal Skip details, plain Cancel and solid Mark done |

### 2 — The task list (items 3, 4, 5, 7)

Before: solid blue provenance chips louder than the task names, a raw device id
(`5ff1f1bb…`) where a device name belongs, "Remove orphaned tasks" as the brightest
button on the page, full-weight zero-count scopes, and "every 12 months after
completion" beside "Monitored". After: outlined provenance chips carrying the
integration's own icon, no id, a destructive action drawn as one, dimmed empty scopes,
and sentence case throughout.

| Before | After |
| --- | --- |
| Task list with solid blue Managed by chips, a raw device id chip and a bright Remove orphaned tasks button | Task list with outlined Managed by chips, no device id, and Remove orphaned tasks as red text |

### 3 — A task page (items 4, 7)

Before: Done and Edit at the same weight, "Edit in Battery Notes" as a third fill, and
"Jun 11, 2026, 11:30:00 AM". After: solid Done, tonal Edit, plain "Edit in Battery
Notes", and a time to the minute.

| Before | After |
| --- | --- |
| Task page with Done, Edit and Edit in Battery Notes all filled, and a next-due time showing seconds | Task page with solid Done, tonal Edit, plain Edit in Battery Notes, and a next-due time without seconds |

### 4 — Settings (items 2, 4)

Before: "Connected" as white on mid-tone green (3.30:1), and "Add profile" / "Add
notification" as solid blue. After: the soft/ink pairing the other chips already use,
and tonal Adds.

| Before | After |
| --- | --- |
| Settings with solid green Connected chips and solid blue Add profile and Add notification buttons | Settings with soft green Connected chips and tonal Add profile and Add notification buttons |

### 5 — An empty scope, and a delete confirm (items 4, 7)

The empty state had no way out but the pill you came from. The confirm scrim drew
Cancel at confirm weight beside the red Delete.

| Before | After |
| --- | --- |
| Empty shopping scope with no way back to the full list | Empty shopping scope with a Show all tasks action in the empty state |
| Delete confirmation with a solid blue Cancel beside the red Delete | Delete confirmation with a plain Cancel beside the red Delete |

## Two calls worth flagging

- **The zero-count scope pill is dimmed but still pressable.** Disabling it is the
  tidier reading, but "is my shopping list really empty?" is a fair question, and
  `screenshots.capture.ts` deliberately photographs that empty scope. The issue offered
  dimming *or* an escape hatch; this does both, and the escape hatch is the half that
  actually closes the loop.
- **The seconds in the completion dialog's own time field are not ours.** HA's
  `ha-selector-datetime` hard-codes `enable-second` on its `ha-time-input` and exposes
  no option to turn it off. Everything the panel formats itself now reads to the
  minute; replacing HA's control would mean hand-building `ha-date-input` +
  `ha-time-input`, which the "don't build on lazily-loaded HA components" rule warns
  against. Happy to open an upstream issue if you'd like.

Also: Edit ends up **primary** on an appliance page and **secondary** on a task page.
That is deliberate — a task page exists for Done, an appliance page has no completion
action — but it reads like a miss if you don't know.

## Guarding it

`raised` survived a full hi-fi redesign because no test can see an attribute nobody
reads. Both new specs assert on **rendered pixels**, not markup:

- `button-weights.spec.ts` — Done, Edit and Delete are three different weights; the
  dialog doesn't draw Cancel like Confirm; every tonal label clears 4.5:1 on its own
  fill.
- `list-honesty.spec.ts` — an empty scope is dimmed and escapable, a completed one-off
  reads as done under *every* grouping, and no surface renders a bare device id.
- `completion-dialog.spec.ts` gains a `headerTitle` guard beside the existing #144
  footer-slot one.

`docs/images/44-panel-shopping-filter.png` had been documenting the empty-scope dead
end for months without anything failing — a capture photographs a surface, it doesn't
cover it.

A later self-review also caught a regression this PR introduced: the registry-presence
check meant to keep a departed device out of a section heading was true for a device
that is present but *nameless*, which headed the bucket with an empty string instead.
It keys on a resolvable name now (`groupableDeviceId`).

## Checklist

- [x] Tests run locally — 758 vitest, 1532 pytest, 196 Playwright (on the headless
      shell CI uses, not `CHROMIUM_EXEC`), ruff clean, frontend mutation **98.88%**
- [x] `CHANGELOG.md` updated with `(Fixes #262)` in the bullet's first paragraph
- [x] Panel UI changed → screenshots re-captured and embedded above as HTML image
      tags, SHA-pinned
- [ ] New user-facing UI surface → no new surface; the tour needs no new steps
- [x] Version bumped to `0.19.0b4` (`manifest.json` + `const.py`) — `v0.19.0b3` is
      already a published tag, so `changelog-release-gap` requires it
- [x] `preview-release` label applied — installable as `0.19.0.dev270` via HACS

`mypy` isn't run here: this box is Python 3.11 and HA's floor is 3.14.2, so installing
HA would silently resolve a stale version — the #199 trap. The Python diff is two
lines (the version bump and the issue ledger), so it can't move mypy either way.

The button vocabulary, the missing `--hk-ok-ink`, and the note that `ha-dialog` has now
broken the panel three times the same way are written into
`.amazonq/rules/architecture-and-code.md`.


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones. This PR rewrites how every button in the panel states its weight and changes two shared helpers, so the blast radius is wide.

**Correctness — the load-bearing claims.** The whole change rests on two assertions about Home Assistant's current frontend: that `ha-button` reads neither `raised` nor `destructive` (its `observedAttributes` are `appearance`/`variant`/`size`), and that `ha-dialog` ignores `heading` in favour of a `headerTitle` slot. I verified both against the live elements in the e2e container. Please check whether anything in `panel.ts` still depends on the old behaviour, and whether `_makeDialog`'s "set the title both ways" trick can double-render a title on any HA version — my reasoning is that a light-DOM child whose slot name matches no slot is never rendered, so an older frontend drops the span and uses the attribute.

**`deviceName` now returns `''` instead of the device id.** This is a behaviour change in a shared helper with nine call sites across `panel.ts`, `card.ts` and `card-filter.ts`. Please hunt for a caller I have not re-checked, especially anywhere an empty string is truthy-tested, concatenated, or used as a Map key. I made group-by-device registry-aware in two places so an unresolvable device falls into the existing "No device" bucket — is there a third grouping path I missed?

**`setBtnWeight` idempotence.** It clears the attributes the new weight does not set. Is the `BTN_ATTR_NAMES` list exhaustive against `BTN_ATTRS`, and can a re-weighted button retain stale state? A drift here is silent.

**Maintainability.** Is `data-hk-weight` pulling its weight, or is it duplication of the attributes it sits beside? Is `_makeDialog`'s deferred `mount()` callback worth the indirection over just appending in order? And is `sentenceCase` in `recurrenceSummary` too clever — should the sixteen locale files have been edited instead?

**Performance.** `btnAttrs` builds a string per button per render and the task list re-renders wholesale. Does that matter at a few hundred tasks, or is it noise next to the `innerHTML` assignment?

**HA best practices.** Reaching into `ha-button` via `::part(base)` to restate the tonal label colour — is there a supported route I have missed? And is keying a style rule off a `data-*` attribute rather than a class a smell here?

**Accessibility.** The tree caret became a real `<button>` with `aria-expanded` kept in sync on toggle; the zero-count filter pill is dimmed to 0.5 opacity but deliberately still pressable. Is the dimmed-yet-pressable state defensible, or does it need `aria-*` to say the scope is empty?

---

prestomation:
/q review Review this PR critically and skeptically. Most serious issues first, but don't withhold minor ones.

Focus on `custom_components/home_keeper/frontend/src/utils.ts`, `panel.ts` and `card-filter.ts`.

1. Correctness: `deviceName()` now returns `''` instead of the device id. Find any caller across `panel.ts`, `card.ts` and `card-filter.ts` that breaks on an empty string — truthiness checks, concatenation, use as a Map key, or a grouping path I did not make registry-aware.
2. Correctness: `setBtnWeight()` clears attributes the new weight does not set. Is `BTN_ATTR_NAMES` exhaustive against `BTN_ATTRS`, so a re-weighted button cannot keep stale state?
3. Correctness: `_makeDialog()` sets the dialog title both as a `heading` attribute and as a `slot="headerTitle"` child. Can any Home Assistant version render both and show the title twice?
4. Maintainability: is `data-hk-weight` redundant with the attributes beside it? Is `sentenceCase` in `recurrenceSummary` too clever versus editing the sixteen locale files?
5. Performance: `btnAttrs()` builds a string per button per render, and the task list re-renders wholesale.
6. Security and HA best practices: reaching into `ha-button` through `::part(base)`, and styling off a `data-*` attribute rather than a class.
7. Accessibility: the tree caret is now a real `button` with `aria-expanded`; the zero-count filter pill is dimmed to 0.5 opacity but deliberately still pressable.

---

prestomation:
Pushed `3789e0e` after self-reviewing the diff. The previous review found nothing, so I went back through the three risks I had flagged myself — and one was a real regression I had introduced in this PR:

**The registry-presence check was the wrong question.** To keep a lost device out of a section heading I added `deviceKnown(devices, id)`, which is true for a device that is *present but has no name at all* (`name` and `name_by_user` both unset — some integrations do this). `deviceName` resolves that to `''`, so Group by Device headed the bucket with **nothing** instead of with an id. A different wrong answer, not a fix. The test is now whether the device can be *named*, which is what a heading needs; `deviceKnown` is gone, replaced by `groupableDeviceId` shared by both grouping call sites, and a nameless device joins the departed ones in "No device".

**`BTN_ATTR_NAMES` is now derived from `BTN_ATTRS`** instead of restated beside it. A hand-written list silently stops clearing an attribute the moment a weight adds one it does not name, and the symptom is a button keeping a colour from the weight it used to have.

/q review Review commit `3789e0e` critically and skeptically, on top of the earlier changes. Most serious issues first; don't withhold minor ones. Files: `custom_components/home_keeper/frontend/src/utils.ts`, `panel.ts`, `card-filter.ts`.

1. `groupableDeviceId(devices, deviceId)` returns the id only when `deviceName` resolves non-empty. Both grouping call sites now call `deviceName` twice per task (once to decide the key, once to build the label). Is there a correctness problem where a task's device is named at key time and not at label time, and is the double lookup worth caching?
2. `BTN_ATTR_NAMES` is computed as the union of keys across `BTN_ATTRS` at module load. Is the module-init ordering safe given `BTN_ATTRS` is declared above it, and does anything depend on the array's order?
3. Did removing the exported `deviceKnown` leave a dangling import or reference anywhere?
4. The `// Stryker disable next-line ConditionalExpression` on the `!deviceId` guard in `deviceName` claims the mutant is equivalent because a falsy id looks up `undefined` and hits the `!dev` guard. Is that claim actually true for every falsy id, including `''` and `0`?

---

prestomation:
Reverted the appliance-tree restyle (issue item 6) at the maintainer's request — `1a78a5f`. The tree keeps its original look: child cards back to the accent tint and accent rail, caret back in the corner of the parent card. `hk-tree-child` and `.hk-chevron` are byte-identical to their pre-change form, `_assetCard` takes `isLast` again, and `tree.toggle` is removed from all sixteen locales along with its CHANGELOG bullet. The other six items stand.

/q review Review commit `1a78a5f` critically and skeptically. It is a partial revert, so the risk is leftovers rather than new logic. Most serious issues first; don't withhold minor ones. Files: `custom_components/home_keeper/frontend/src/panel.ts`, `src/locales/*.json`, `CHANGELOG.md`.

1. Is the revert complete? Look for anything from the tree restyle still present — a `.hk-chevron-space` rule, an `inTree` reference, `aria-expanded` handling in the tree toggle handler, or CSS that no longer has a consumer.
2. Is anything from the *other* six items collaterally damaged? The revert touched a shared CSS region and `_assetCard`, which also renders the device chip and the archived chip.
3. `tree.toggle` was removed from all sixteen locale files. Confirm no dangling reference, and that no other key was disturbed — the files are hand-ordered, not sorted.
4. `_assetCard`'s third parameter is `isLast` again: declared, passed, and never read. It was dead before this PR and is dead again. Worth deleting outright, or is restoring the prior signature the right call for a revert?

---

PR #271, opened 2026-08-30
refactor: one word for keeping in step with the outside — sync, not mirror

prestomation:
## What changed

Home Keeper used **mirror** for three different things, and the vocabulary had already drifted half-way to "sync" on its own:

| Concept | Called in code | Called in the UI |
|---|---|---|
| Profile-filtered tasks pushed to an external `todo.*` list | `task_mirror.py` / `task_mirror_sync.py` | "Sync to a to-do list", "Two-way sync" |
| Auto-buy reminders pushed to a shopping list | `shopping.py` / `shopping_sync.py` | "Shopping list" |
| `device_class: problem` sensors turned into tasks | `problem_tasks.py` / `problem_sync.py` | "Problem sensor sync" |

So the panel said *"Sync to a to-do list"* over i18n keys named `mirror.*`, the profile field was literally `sync`, the CSS was `.hk-sync-group`, the docs-site slug was already `todo-sync` — while the Python underneath was `TaskMirrorSync`. Meanwhile `settings.sync_off` (problem sensors) read *"Problem sensors are not mirrored"*: the two features had swapped each other's word.

**Sync** is now the single word for "kept in step with something outside Home Keeper", on every surface and in all sixteen languages.

### The feature rename

- `task_mirror.py` / `task_mirror_sync.py` → **`todo_list.py` / `todo_list_sync.py`**, preserving the repo's `NAME.py` (pure planner) / `NAME_sync.py` (HA-aware driver) convention. A literal `mirror`→`sync` swap would have produced `task_sync_sync.py` and `class TaskSyncSync`, because `_sync` already means "the driver half".
- `TaskMirrorSync` → `TodoListSync` (beside `ShoppingListSync`, `ProblemSensorSync`); `TaskMirrorPlan` → `TodoListPlan`; `mirror_key` → `sync_key`; `desired_by_mirror` → `desired_by_sync`. The record a plan is built from *is* a profile, so `mirrors`/`mirror` became `synced`/`profile` rather than colliding with the `sync` block a profile already carries.
- `ORIGIN_TODO_MIRROR` → `ORIGIN_TODO_SYNC`, value `home_keeper_todo_sync`.
- Storage key `task_mirror_items` → `todo_list_items`, **no migration**. The feature shipped in 0.18.0, and an emptied bookkeeping self-heals without duplicating: `plan_sync`'s phase 2 looks for an open line with the same summary and adopts it instead of adding a second one — a path that exists precisely for bookkeeping that was lost.
- i18n namespace `mirror.*` → `todo_sync.*` across all 16 locales, including the string-concatenated `t('todo_sync.' + s.name)` lookup in `panel.ts` that builds three field labels and that a key-by-key grep would have missed.

### User-visible prose

Every string that said "mirror" now says "sync" — for the to-do list sync, the shopping list and problem sensors alike — in `services.yaml`, `strings.json`, `translations/*.json` and the frontend locales, English **and** the fifteen translations (each using its own language's sync verb rather than its mirror verb).

### Left alone on purpose

Ordinary English "mirrors"/"mirroring" in internal comments about unrelated things (`limits.ts` mirrors `const.py`, the Settings tab mirrors the options flow), the shipped CHANGELOG sections, `docs/*_PLAN.md` scratch docs, and the two historical DOM ids `#hk-task-mirrors` / `#hk-mirror-add` that a test asserts are *absent* — renaming those would stop the guard guarding anything.

`ha-integration-template` needs no changes: all ten of its "mirror" hits are ordinary English in comments.

## Screenshots

The Settings tab, with the reworded copy on all three cards:

Home Keeper Settings tab: Buy reminders synced to Shopping List, Problem sensors synced as tasks

A profile's **Sync to a to-do list** group — the three field labels here come from the concatenated `todo_sync.*` lookup, so this shot is what proves that rename landed:

Profile edit form with the Sync to a to-do list group: To-do list picker, Two-way sync, Treat removed items as completed

The synced chores on the household's own list, proving the renamed storage key and origin marker work end to end:

Family chores to-do list carrying the synced Home Keeper chores with their due dates

Settings on a narrow screen:

Mobile Settings index: Buy reminders synced to Shopping List, Problem sensors synced as tasks

## Verification

Run locally, all green:

| Tier | Result |
|---|---|
| `pytest tests/unit` | 1533 passed |
| `npx vitest run` | 758 passed (26 files) — incl. i18n parity + untranslated-leak guard |
| `bash ci/e2e-up.sh` (Playwright) | 196 passed |
| `bash ci/test-python-integration.sh` (Docker) | 167 passed |
| `bash ci/test-mutation-python.sh` | **98.97%** (574/580), threshold 80% |
| `bash ci/test-mutation-frontend.sh` | no mutable TypeScript changed |
| `vale` | 0 hits on lines this PR changes (40 pre-existing, all on untouched lines) |
| `ci/check-changelog-release-gap.py` | clean |

`mypy` could not run here — Home Assistant won't `pip install` in this sandbox (a native wheel fails to build). `mypy --follow-imports=skip` over the two renamed modules is clean; CI runs the real check.

The Docker tier is the meaningful one for the persisted rename: `test_todo_list_sync.py` exercises the renamed `todo_list_items` key and `origin: home_keeper_todo_sync` against a real HA container.

## Checklist

- [x] `CHANGELOG.md` updated — no issue references, so no `(Fixes #N)`
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML img tag
- [ ] New user-facing UI surface → n/a, no new surface; the tour needs no new step
- [x] Version bumped to `0.19.0b5` (`manifest.json` + `const.py`) — required because `v0.19.0b4` is already tagged. No `preview-release` label: terminology change, not a new feature.


prestomation:
/q review This is a terminology rename: the "task mirror" feature becomes "to-do list sync", and every user-visible string that said "mirror" now says "sync". Please be critical and skeptical, most-serious issues first, and don't withhold minor ones.

Specifically scrutinise:

**Correctness / completeness of the rename.** Is there any call site, string, or persisted name I missed? The riskiest spots: `panel.ts` builds three form labels by string concatenation (`t('todo_sync.' + s.name)`), so a key-by-key grep would not have caught it; `tests/unit/test_todo_list_sync.py` loads the driver by *file path string*; `pyproject.toml`'s `only_mutate` and `tests/conftest.py`'s pure-module list both name the module by path. Are there other indirection points of that shape?

**The un-migrated persisted rename.** `task_mirror_items` -> `todo_list_items` in `.storage/home_keeper`, and `ORIGIN_TODO_MIRROR` -> `ORIGIN_TODO_SYNC` (wire value `home_keeper_todo_sync`) in the event stream, both with **no** migration or back-compat shim. The reasoning is that the feature shipped in 0.18.0 and the bookkeeping self-heals on the next pass. Is that self-healing actually total? On upgrade the store reads no bookkeeping, so every synced task looks untracked — does the planner re-add duplicates beside items already on the list, or does `_identity`'s summary fallback re-attach them? If it duplicates, that argues for reading the old key once.

**Naming.** `_sync` is the repo's suffix for "HA-aware driver" (`shopping_sync.py`, `problem_sync.py`), which is why the pair became `todo_list.py` / `todo_list_sync.py` rather than `task_sync_sync.py`. Does `todo_list.py` read confusingly next to the `todo.py` entity platform? And inside the planner the record is now `profile` (the collection `synced`) because a local `sync` already existed — is that clearer than the `mirror` it replaced, or have I traded one ambiguity for another?

**Copy quality.** Three separate surfaces now say "sync" in the Settings tab ("Buy reminders synced to …", "Problem sensors synced as tasks", "Sync to a to-do list"). Are they still tellable apart by a user, or does the shared verb blur them? Same question for the 15 non-English locales, where I switched each language's mirror verb to its sync verb.

**HA best practices** for the options/translation surfaces, and anything that would break an integrator relying on the documented event contract.

---

prestomation:
Correcting the record on the comment above about `max_sync_retries` — none of the code it describes exists, in this PR or on `main`. I checked every symbol it names:

| Cited | Actually in the repo |
|---|---|
| `max_sync_retries` | no match anywhere |
| `_sync_state` dict, `todo_list_sync.py:121` | no match |
| `_async_sync_one`, `todo_list_sync.py:260-299` | no match (the file is 521 lines and has no such method) |
| `test_sync_retries_failed_syncs` | no match |
| `test_sync_stops_retrying_after_max_retries` | no match |
| `test_sync_resets_retry_count_on_success` | no match |

There is no retry-count mechanism to review: the sync's failure model is *retry-not-compensate* — a failed operation puts its bookkeeping entry back so the next pass tries again, with no counter and no give-up state. The only bounded loop is `_MAX_PASSES = 4`, which caps re-entrant passes within a single sync, not retries of a failed one.

The comment also describes "extensive additions to `panel.ts` for task management UI, filtering, and localization features". This PR changes `panel.ts` by 5 lines — four `t('mirror.…')` lookups renamed to `t('todo_sync.…')` and one docstring word. It adds no UI.

Flagging it so nobody takes the claimed retry tests as existing coverage. Amazon Q's actual review on this PR (the one above it) is accurate and found no blocking defects.

---

PR #272, opened 2026-08-30
feat(panel): reach snooze and skip from the panel, log skips, and fix skip on meter tasks

### LINKED ISSUE #268: Skip and Snooze
I'm confused on what exactly the Skip and Snooze actions do. I know that they are notification actions, but there's no documentation on how exactly they work.

- Does Skip act as a "phantom" completion and generate the next schedule due date? If so, what does the completion history look like? Does it show "Skipped" with the skip date, or does it just push the date of the current task out to the next scheduled date for fixed/floating tasks? Why is it only available as a notification action, but not actually in the task details UI?
- How long is the Snooze duration? Is it 1 hr, 12 hrs, 24 hrs, 2 days, etc.? Is it just a notification snooze without changing the task due date, or does it actually push out the due date? If it is the latter, why is it not available in the task details UI?

Without knowing the exact intent behind these features, here would be my ideal design for each:

### Skip
Acts as a "phantom" completion and generates the next scheduled due date. In the completion history UI for the task, it would show a log for the Skip with the usual details (skip date and buttons to adjust the date, details (notes, who did it, etc.), and delete). This should work for floating, fixed, and meter/usage based tasks. The Skip button should be available in the task details UI.

A few examples:
- Change the water filter every 1 month after the last completion. Let's say it's currently due 8/29/26, but I've been on vacation for the last month, and I don't think it needs changing. I'd like to "Skip" the current due date. It would be logged in the completion history as a "Skip", and it would generate the next due date as 9/29/26.
- Change the car oil every 5,000 miles after the last completion. Let's say it's currently due, and I have 30,000 miles on the car. Maybe for some reason I want to "Skip" this current instance. In that case, the "Skip" would be logged in the completion history (noting it was Skipped at 30,000 miles, and it would generate the next due usage at 35,000 miles).

### Snooze
Pushes out completion dates/usages by a fixed amount. This could be user selectable with a few predefined defaults (1 hr, 1 day, 1 week, 1 month, plus a custom option where the user could choose the date). If this could work with meter/usage tasks, that would be even better. The predefined defaults could be percentages of the usage interval (for a task due every 10,000 miles, the snooze options, for example, could be 100 miles (1%), 500 miles (5%), and 1,000 miles (10%)), plus a custom snooze duration. Snoozing would actually push out the due date by the chosen date/usage. Not sure if this should be logged in the completion history UI. The Snooze button should be available in the task details UI.

A few examples:
- Change the water filter every 1 month after the last completion. Let's say it's currently due 8/29/26, but I've been on vacation for the last week, and I don't think it needs changing right away. I'd like to "Snooze" the current due date by 1 week, and the due date would be changed to 9/5/26.
- Change the car oil every 5,000 miles after the last completion. Let's say it's currently due, and I have 30,000 miles on the car. Maybe for some reason I want to "Snooze" this current task by 500 miles. In that case, the due usage would be changed to 30,500 miles.

prestomation:
Fixes #268

> **Merged `main` (`a6f8c5f`) and re-landed the panel wiring.** This branch was cut before `main` split `panel.ts` into 15 `panel-*.ts` region modules, so its 430 lines of panel wiring had no file left to apply to. The feature's own modules merged untouched and the backend merged with 3 one-line conflicts; what needed redoing was the wiring. Version moves to **0.21.0b4**, because 0.21.0b3 is already released. Details in "What the merge cost", below.

## What changed

Snooze and Skip shipped as `home_keeper.*` services and notification buttons, but neither had a websocket command — and the panel only talks `callWS`. That is the structural reason they have only ever been notification buttons, and it is what the issue was really reporting: two verbs nobody could find, with no documentation of what they did to the schedule.

They are now reachable from both UI surfaces, shaped to each.

### On the task page and list rows: a caret inside Done

**Done** grows a caret that opens both. They sit behind it rather than beside it: they are the exceptions to the one action a task page is really for, and three peers would read as three equal choices. Done keeps its own hit target, so it never costs an extra tap or changes meaning. Each menu entry states what it does to the schedule.

<p>A task detail page with the Done caret open, showing a Snooze entry described as Push the due date out and a Skip entry described as Move to the next occurrence</p>

The caret is an `ha-button` taking **the weight Done itself has**, which is the only way the two halves are guaranteed to paint alike. Home Assistant fills a button from its appearance and the weights differ by surface — Done on the task page is solid accent, on a list row it is a pale tonal — so anything naming a colour is wrong on one of them, and wrong again under someone else's theme. Both halves square off and the wrapper rounds the pair by clipping, since `ha-button` takes a single-value radius override and rejects a four-value one.

**Snooze** opens a dialog with 1 hour / 1 day / 1 week / 1 month / Custom, and a live line resolving the choice to a real date so the user reads the answer rather than doing the arithmetic. The presets live in one exported constant, so a future editable-presets option has somewhere to plug in.

<p>The Snooze task dialog: a Snooze until dropdown set to 1 week, and the line Due date moves to Sep 13, 2026</p>

### On the dashboard card: both verbs on the row

The card takes a different shape because it is answering a different question. A task page is *about* one task, so it can make one action primary and tuck the exceptions behind a caret. A card row is a list you scan, and a chevron beside a dense icon button has no container to lean on and reads as decoration. So both verbs sit on the row, muted against the accent on Done and placed ahead of it, leaving the rightmost target the one people mean.

<p>A dashboard card whose rows have a snooze and a skip button ahead of the accent Done button</p>

That screenshot is also the eligibility rules in one frame: the completion-blocked Sump pump row offers snooze but not skip, and the dormant Monitored batteries, the completed one-off and the metered nozzle offer neither.

Press and hold either button and it says what it does, using the same line the panel's menu shows. Two icons on a row say nothing about themselves, and the only other way to learn them was to try one and watch the schedule move. The hold deliberately does not then act.

Both surfaces run the same code: `defer.ts` holds the decisions and `defer-dialogs.ts` the menu controller and the two dialogs, so the card opens the panel's dialogs rather than an imitation of them.

### A skip is recorded, and is not a completion

A skip now writes an entry to its own `skips` list beside `completions`, with a note, a person and (for a meter task) a reading. It renders in history interleaved with completions but with its own chip and a muted date, and can be amended, re-dated or undone through `update_skip` / `move_skip` / `delete_skip`.

<p>Task history with a skipped occurrence at the top, marked Skipped with a note, above six completions, while the tally still reads 6 completions</p>

Keeping the two lists apart is the load-bearing decision. "A skip never counts as work done" is then true by construction rather than by filtering at the ~18 places that read `completions` — including two that would have been outright bugs: `remove_completion` and `move_completion` re-derive `last_completed` as `max(history)`, so a shared list would let a later skip become the floating anchor when an earlier completion was undone, jumping the schedule a full interval. The completion tally, the cadence average, `last_completed`, the sensor attributes and the appliance archive all keep their meaning, and the raw task dicts published by `list_tasks` and `get_tasks` keep theirs for integrators.

It is also additive: no `STORAGE_VERSION` bump, every read is `.get("skips", [])`, and a document written before this reads back as having none.

### Skipping a metered task actually skips it

The oil-change example in the issue did not work. Skip cleared `next_due` but left `sensor.baseline` alone, so `evaluate_usage` still saw the target exceeded and re-armed the task on the next watcher tick.

The fix has two halves, and only doing the first leaves the bug half-alive: with the default `combinator: "any"` an elapsed `also_every` backstop re-arms the task however recently it was skipped. So skip resets the meter exactly as a completion does, and `backstop_due` now measures from the last *decision* about the task — the later of a completion and a skip, via the new `sensor_tasks.latest_decision_ts`, which also tells the undo paths which entry currently anchors the meter. Deleting a skip restores the baseline it replaced, recorded on it as `meter_start`.

### A global Skip &amp; snooze setting

Two options, `allow_snooze` and `allow_skip`, both defaulting **on**, with a Settings section of their own. Turning one off withdraws it from the panel, the card and notification button sets, but never from the services — those are the interoperability contract, and silently breaking an automation someone already wrote is not a setting.

<p>The Settings tab with a Skip and snooze section holding an Allow snooze and an Allow skip switch, both on, and a matching entry in the section rail</p>

Defaulting *on* is the same promise every other option makes by defaulting off: nothing changes for someone who never opens the setting. Both verbs predate the switch, so the frontend reads them through `skipSnoozeFlags` rather than `!!v` — a missing key means "not configured", not "off", and that is every existing install.

One documented exception survives: `actions_for` still forces Snooze onto a completion-blocked task even when the switch is off. A notification walk only advances on a successful action and such a task can be neither completed nor skipped, so the alternative is a reminder that can never be got past (#248). The help text says so.

### Services and events

- `snooze_task` gains `until`, mutually exclusive with `hours`. Whole hours cannot name "next Tuesday 09:00" across a DST boundary, which the dialog needs. `hours` keeps its default, so every existing caller is untouched.
- `skip_task` gains `note` / `who` / `reading`.
- New `update_skip`, `delete_skip`, `move_skip`, mirroring the completion trio.
- New `home_keeper_task_skip_updated` / `_skip_removed` events; `_skipped` gains `ts`.
- Five websocket commands, each delegating to the same store method as its service twin.

Also fixes a contract that was documented but never true: `origin` is now actually echoed in `home_keeper_task_snoozed` and `_skipped`, as `services.yaml`, `strings.json` and the README have all promised.

## What the merge cost

The feature's own modules — `defer.ts`, `defer-dialogs.ts`, `dialogs.ts` — merged untouched, so the decisions and the dialogs are the code that was already reviewed. The panel wiring was re-landed in the architecture it now lives in: a new **`panel-defer.ts`** holds the panel's half as free functions over `PanelHost`, in the same shape as `panel-dialogs.ts`, and each region took its own piece (styles to `panel-styles.ts`, skip rows to `panel-history.ts`, the Done caret to `panel-detail.ts`, row carets to `panel-lists.ts`, the Settings card to `panel-settings.ts`). `PanelHost` gains `_skip`, `_snooze` and `_wireDeferMenus`.

Four merges needed a decision rather than a keep-both, and each is worth a reviewer's eye:

1. **`main` replaced the card's private `_toast` with the shared `toast()` helper.** Git converted the three call sites both sides knew about and left this branch's new one calling a method that no longer exists — a clean textual merge that would not compile. `tsc` caught it.
2. **`main` added `_store_errors`**, a context manager doing exactly what the four skip handlers here hand-rolled. They now use it, which is how they would have been written against this base — about 60 lines lighter.
3. **The 33 locale and translation files are merged structurally**, from the three git stages by key, not textually. Both sides appended distinct keys at the same spot and concatenating the sides dropped a brace; a by-key merge cannot.
4. **`notifier_harness.py`'s fake `current_options` returned `entry.options` verbatim**, so `notifier`'s subscript of the two new switches raised `KeyError` in 10 tests. It now defaults from the real `_empty_options()` table, so the fake cannot drift from what a running install sees.

The README section is re-inserted into `main`'s copy rather than merged: this branch predates a section reorder, so the textual merge silently duplicated six `##` sections.

Since the merge, three follow-ups landed: the `HomeKeeper.Personification` vale rule arrived on `main` mid-branch and the seven prose uses it rejects are reworded, the two deferral-menu lifecycle invariants an automated review queried are now pinned by tests rather than by argument, and the Settings capture is refreshed for the reworded help text.

## Two things worth a reviewer's eye

**Scan-locked tasks can be skipped and snoozed.** `_reject_scan_required` gates `complete_task` only. The tag proves you were physically present to *do* the work; skipping asserts you did not do it and snoozing asserts nothing, so neither seemed to warrant a scan. Deliberate, but a household using tag locks to enforce presence might disagree.

**Snooze-by-usage (`+500 miles`, from the issue) is out of scope.** Snooze writes a due *date*, and a meter offset has no home in the data model yet.

## One-way doors

| Surface | Shape | Who relies on it |
|---|---|---|
| `task["skips"]` | list of `{ts, note, who, reading, meter_start}` | `list_tasks` / `get_tasks` consumers; storage document |
| `snooze_task.until` | ISO datetime, exclusive with `hours` | service callers, automations |
| `skip_task.note` / `.who` / `.reading` | strings / number | service callers |
| `update_skip` / `delete_skip` / `move_skip` | `task_id` + `ts` (+ `new_ts`) | service callers |
| `home_keeper_task_skip_updated` / `_skip_removed` | event payloads | automations |
| `home_keeper_task_skipped.ts` | added field | automations |
| `allow_snooze` / `allow_skip` | bool options, default **true** | stored options |

`skips` being its own list beside `completions` is the big one: moving it later would change every integrator reading a raw task dict.

## Verification

Run locally on the merged tree, against a container rebuilt from the committed fixture.

| Tier | Result |
|---|---|
| `pytest tests/unit` | **1920 passed**, 2 skipped |
| `bash ci/test-frontend.sh` | **1132 passed** (36 files) |
| Playwright e2e (desktop + tablet + phone) | **271 passed**, including 9 `skip-snooze.spec.ts` cases |
| `npx tsc --noEmit` / `bash ci/build-panel.sh` | clean |
| `doc-anchors.test.js` | 13 passed |

CI is green on `6b23d43` across every check, the mutation gate and the Docker tiers included.

Every committed screenshot was read back and visually checked; the other 79 the capture run rewrote are reverted, since they differ only in render nondeterminism.

## Checklist

- [x] `CHANGELOG.md` updated with `(Fixes #268)`, as `0.21.0b4`
- [x] Panel UI changed → screenshots committed under `docs/images/` and embedded above
- [x] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended
- [x] New major feature → documented in `README.md` and given a User Guide page
- [x] Version bumped to `0.21.0b4` (`manifest.json` + `const.py`); `preview-release` label applied

https://claude.ai/code/session_012Riitqtw1CZ5rnoXbSazh5


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first — but don't withhold minor ones. Focus on:

**Correctness.** The load-bearing claim is that a skip is never a completion: `skips` is a separate list, and `remove_completion` / `move_completion` re-derive `last_completed` as `max(history)` from `completions` alone. Look for any path where a skip could still leak into a completion count, the cadence average, `last_completed`, the sensor attributes, the appliance archive, or the raw task dicts published by `list_tasks` / `get_tasks`.

**The meter fix specifically.** `store.skip_task` resets `sensor.baseline` and stamps `meter_start`; `sensor_tasks.latest_decision_ts` is the new anchor for both `backstop_due` and the undo paths (`baseline_after_delete`, `_reanchor_from_completion`, `_reanchor_from_skip`). Check the interaction with `also_every` + `combinator: "all"` (not just the `any` default), a meter that resets or rolls over, a skip and a completion at the same instant, and the ordering of `_stamp_meter_start` before `_reset_usage_baseline`. Is there a sequence where undoing leaves the baseline stranded or double-restored?

**Timezones and DST.** `snooze_task` now accepts an absolute `until`; naive datetimes are qualified in `__init__.py` and `websocket_api.py`. `resolveSnoozePreset` in `utils.ts` clamps month arithmetic client-side and is meant to agree with `recurrence.add_months`. Do they actually agree across a DST boundary and on Jan 31 + 1 month?

**Options and defaults.** `allow_snooze` / `allow_skip` default **True**, which is new — every other option defaults falsy. Check `_empty_options`, the generalised `_BOOL_OPTIONS` branch in `_normalize`, `merge_flow_input` (these are `in_flow=False`), and the frontend's `skipSnoozeFlags`. Is there any path where an existing install reads them as off?

**HA best practices.** Five new websocket commands and three new services; admin gating and the service/websocket privilege parity described in `docs/SECURITY.md`. `skip_task` deliberately does *not* call `_reject_scan_required` — is that defensible? Also the `vol.Exclusive` pair on `snooze_task` and whether the default-when-neither-given is handled correctly.

**Maintainability.** The deferral menu binds document-level listeners on open and removes them on close, and `_render` clears them. Look for leaks or a path that opens a second menu without closing the first. Also whether `_deferMenu` / `_deferVerbs` belong where they are.

**Performance.** `latest_decision_ts` parses every skip timestamp on each call and is invoked from `backstop_due`, which the watcher hits per evaluation. Is that hot enough to matter with a 500-entry log?

---

prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first without withholding the minor ones. Since your last pass this branch has only gained a walkthrough beat and a docs note, so please concentrate on the substance you have not yet been able to falsify:

**Correctness of the skip/completion split.** `skips[]` is a second history list beside `completions[]`. Every reader of `completions` was audited on the claim that a skip must never reach it — please try to break that. In particular: does any path let a skip entry become `last_completed`, or a completion become the meter anchor when a later skip exists? `remove_completion` and `move_completion` re-derive `last_completed` as `max(history)`; `sensor_tasks.latest_decision_ts` deliberately spans both lists. Those two rules have to coexist without one leaking into the other.

**The meter-baseline undo path.** Skip now resets `sensor.baseline` and stamps `meter_start`, and deleting a skip restores it. Please check the orderings: deleting a *non-latest* skip, deleting a skip that is older than the newest completion, editing a skip's `reading` (which re-anchors via `_reanchor_from_skip`), and moving a skip across a completion so the two swap order. Is the baseline always left where a user would expect, and is any of it silently lossy?

**Default-true options.** `allow_snooze` / `allow_skip` are the first options in this codebase that default on, so a missing key must read as enabled everywhere — backend `options.py` and frontend `skipSnoozeFlags`. Are there readers that still see a missing key as false? And is the `actions_for` forced-Snooze exception for completion-blocked tasks correctly scoped, or does it leak Snooze onto tasks that are not completion-blocked?

**Concurrency and staleness.** The panel's deferral menu binds handlers on `document` on open and removes them on close. Can a background refresh, a route change or a dialog closing leave a handler attached or a menu orphaned?

**HA best practices / performance** on the five new websocket commands and their service twins — admin gating parity in particular, since `call_service` walks around a websocket-only gate.

---

prestomation:
/q review Critical review of `7ce6473` please, most serious first, and don't withhold the minor ones.

**What it fixes.** `integration` was red on three consecutive heads: `test_todo_entity_exists_with_seeded_tasks` could not find "Renew passport" on the to-do list. Root cause was not the feature code — the committed `tests/integration/ha_config` fixture had a local walkthrough run baked into it. The load-bearing part was a `walkthrough_family_chores` profile carrying `sync.two_way: true` together with the `todo_list_items` bookkeeping it had written. Together those make Home Keeper reconcile against a stale external to-do list *during platform setup*, completing nine seeded tasks eight seconds into startup, before the first test ran. "Renew passport" was one; a completed one-off is dormant, so it correctly left the to-do list and the assertion failed. It passed locally only because my container had been started before the fixture was restored. Verified by running the same test on a clean container against `origin/main` (passes) and this branch (fails), then again after the revert (167 passed).

Please scrutinise:

1. **Is reverting the whole `ha_config` tree to `main` right, or did this PR legitimately need a fixture change?** I concluded every hunk was runtime capture — device_id churn, `require_tag_scan` null→false and `completion_detail` null→"none" normalization writebacks, a moved `last_completed`, two uploaded PDFs. If any seeded state is genuinely required by the new skip/snooze e2e spec or the walkthrough beat, I have removed it and the specs are relying on something that will not exist on a pristine checkout.

2. **Are the two new guards in `test_integration_fixture_clean.py` the right invariants, or too narrow / too strict?** `todo_list_items` must be empty, and no seeded profile may carry a `sync` block. I verified both fail against the fixture as it was committed and pass against the restored one. The second one forecloses ever seeding a sync profile deliberately — is that a reasonable trade for catching this class, or should it assert something weaker?

3. **Does anything else in the seed reconcile at startup** the way two-way sync does — shopping-list sync, problem-sensor sync, asset part reconciliation — such that a future capture run could bake in a different self-mutating fixture that these two guards would not catch?

4. Anything in the earlier substance you have not yet been able to falsify: the skip/completion separation invariants, the meter-baseline undo orderings, and default-true option reads.

---

prestomation:
Thanks — triaged all four points.

**3. Shopping-list sync — taken, shipped in `944bf60`.** You were right that `shopping_items` is the same shape of state behind a reconciler that also runs at startup, and that neither guard covered it. Confirmed it is written only by `shopping_sync.py` (plus a read in `diagnostics.py`) and never authored, so I generalised the guard over a `SYNC_BOOKKEEPING` tuple rather than adding a second near-identical test. Verified non-vacuous by planting an entry under each key in turn — both fail. I did not add a profile-level guard for it, since as you note there is no profile switch to key on; the bookkeeping assertion is the whole coverage there.

**2. Weakening the sync-profile guard — declining, for now.** `sync.two_way` is the flag that makes reconciliation *write back*, but a one-way profile still writes items onto an external list at startup and still leaves `todo_list_items` behind, which is the state the other guard exists to keep out. Narrowing to `two_way` would let exactly that through. If someone later needs a seeded one-way profile, the guard is one line to relax and the failure message will point them at this decision — better than the reverse, where the pollution is silent again.

**4. The omission you flagged — already verified, and it is what I based the guards on.** I checked out the fixture as committed at `415d0f0` and ran the guards against it: both fail, naming the `walkthrough_family_chores` sync profile and the populated `todo_list_items`. Then against the restored fixture: pass. That retroactive check is exactly why I trust them; a guard I had not seen fail would not have been worth adding.

**1.** Agreed, and independently confirmed: the same test on a freshly built container passes against `origin/main` with zero startup completions and fails on this branch with nine.

One correction to your write-up for the record: the `sync.two_way: true` came from the walkthrough profile stored in the fixture's `core.config_entries` options, not from a line in `walkthrough.capture.ts` — the harness creates it through the running panel, and the bind-mounted config is what captured it.

---

prestomation:
/q review This branch was just rebased onto `main` at `f294909` (which merged #273). Please be critical and skeptical, and lead with the most serious findings — but do not withhold the minor ones.

Focus especially on what a rebase can quietly break, since three conflicts were resolved by hand:

1. **Rebase integrity.** `custom_components/home_keeper/frontend/test/panel.test.js` had both sides append a `describe` block and both were kept — check nothing was dropped, duplicated or left inside the wrong closure. Every `docs/images/*.png` that #273 recaptured was resolved in favour of this branch's captures; flag any screenshot that no longer matches the UI it claims to show. `CHANGELOG.md` was split into a new `## [0.19.0b7]` section above the released `## [0.19.0b6]`, with `manifest.json` and `PANEL_VERSION` moved to `0.19.0b7` — check that against the `changelog-release-gap` gate and `ci/release-issues.py`, and that `(Fixes #268)` still sits in the first paragraph of its bullet.

2. **Semantic conflicts the merge could not see.** #273 changed `build_task` (`add_task` now infers recurrence from the schedule fields passed) and the edit-drawer focus/Escape handling in `panel.ts`. This branch adds `skips` to the task dict and a deferral menu to the same surfaces. Does a bare `{name}` `add_task` still produce a task the skip and snooze paths handle? Does the deferral menu interact correctly with the drawer's focus restore, its Escape handler and `_syncDrawerModality`?

3. **Correctness.** `sensor_tasks.latest_decision_ts` and the meter reset on skip — off-by-one, timezone/DST, the `also_every` backstop, and the undo paths (`delete_skip` restoring `meter_start`, `remove_completion`/`move_completion` re-deriving `last_completed`) when completions and skips interleave at the same timestamp. Error paths on `snooze_task` with `until` and `hours` both set.

4. **Privilege model.** Each of the five new websocket commands and its service twin: is the admin gate applied in both, per `docs/SECURITY.md`?

5. **Maintainability and performance.** The `defer.ts` / `defer-dialogs.ts` split, duplication between the panel and card surfaces, and any redundant full reloads on a skip or snooze.

6. **HA best practices** for the new services, events and options, including options-flow merge semantics for `allow_snooze` / `allow_skip`.

---

prestomation:
Triage of the Cue review. CI is green on `b7dcf83` — all 19 checks, including `mutation`, `integration`, `e2e`, `walkthrough`, `mypy`, `vale` and `changelog-release-gap`. Nothing below changes the diff; each item is either answered by code already on the branch or rests on a premise the repo contradicts.

## The CRITICAL finding is a false positive

> non-admins can skip/snooze/edit any task via websocket while the UI correctly blocks them

Snoozing and skipping are **deliberately not admin-only**. `docs/SECURITY.md`, under "Open to any signed-in user", lists exactly this:

> - Completing, snoozing and skipping tasks, and creating tasks.

The review's premise — that "the services properly gate" these — is also not so. There is no service-side admin gate on `snooze_task` or `skip_task`; `_verify_admin` is called only by the handlers `api_surface` models as `admin_only`, and `tests/unit/test_admin_only_services_verify_admin` fails the build if that set and the source ever disagree. The five new websocket commands match their service twins exactly, which is the property SECURITY.md asks for.

The comparison to `ws_set_options` and `ws_update_asset` picks the wrong neighbours. The right ones are the commands these five mirror: `ws_complete_task`, `ws_add_task`, `ws_update_completion`, `ws_move_completion` and `ws_remove_completion` — none carry `require_admin` either, and `test_websocket_admin_flags_match_source` holds every command to its modelled flag. Adding the decorator would take a completion trio and a skip trio that behave alike and make them diverge, and it would lock non-admins out of the reminders the panel's own privilege split exists to keep open to them.

## The rest, in order

**Meter reading read before `skip_occurrence`.** Intentional, and `store.skip_task` says so at the call site: "Same resolution order as `complete_task`". The caller's number wins, otherwise the entity is read once and that same figure anchors both the history row and the baseline, so the two cannot disagree. Re-reading after the advance, as suggested, would make skip differ from completion for no gain.

**Snooze on a dormant task.** `store.snooze_task` rejects `next_due is None` outright with a `TaskValidationError` — a dormant task can't be silently re-armed from any surface. The UI agrees ahead of it: `defer.deferVerbs` withholds both verbs from a dormant task, which is what the card screenshot in the body shows for the completed one-off and the monitored batteries.

**`hours` and `until` both passed.** `SNOOZE_TASK_SCHEMA` declares both as `vol.Exclusive(..., "deferral")`, so voluptuous rejects the call. There is no resolution order to document because both together never reach the handler.

**`latest_decision_ts` ties.** It sorts `(parsed_instant, raw_string)` tuples and takes the `max`, so an exact-microsecond tie between a completion and a skip falls back to the raw ISO string — deterministic, not corrupting. Reaching it needs two deliberately back-dated entries at the identical microsecond.

**Circular import between `defer.ts` and `defer-dialogs.ts`.** There isn't one. `defer-dialogs` imports from `defer`; `defer` imports nothing from `defer-dialogs`. `tsc --noEmit` and the Rollup build are both clean.

**Card duplicating the defer code.** It doesn't — `card.ts` imports `deferRowActions`, `deferVerbs`, `emptySkipState`, `emptySnoozeState` from `defer`, and `renderSkipDialog` / `renderSnoozeDialog` from `defer-dialogs`. That sharing is the point of the split.

**Options default when the key is absent.** Handled on both sides: `options._empty_options` sets `allow_snooze` / `allow_skip` to `True`, and the frontend reads them through `forms.skipSnoozeFlags`, which defaults each to `true` rather than using `!!v`. An install predating the switches sees both verbs, unchanged.

**Entity reload after a skip.** `ws_skip_task` uses `async_request_refresh`, the same as every completion path; a skip changes neither `device_id` nor `enabled`. If a refresh were insufficient for a reconcile-driven entity change, completions would have the same defect today — that would be a pre-existing issue, not one this branch introduces.

**Service YAML descriptions.** `services.yaml` and `strings.json` parity is enforced by `tests/unit/test_api_surface.py`, which parses the component's source and fails on drift; the Developer Guide's API reference is generated from the same declarations. `until` carries a `datetime` selector.

**Screenshots.** One real caveat, unrelated to the review: the Settings-rail footer in the committed captures still reads `v0.19.0b6`, because b7 came from the rebase rather than a fresh capture. Recapturing is a one-command rerun of the harness if it is worth the binary churn.

---

prestomation:
/q review Be critical and skeptical, and put the most serious findings first. Do not withhold minor ones.

This push merged `main` into a branch cut before `panel.ts` was split into the `panel-*.ts` region modules, so the panel wiring was re-applied by hand. Please give that re-application the hardest look:

- **Correctness of the re-landed wiring.** `panel-defer.ts` is new and binds `defer-dialogs.ts` to `PanelHost`. Check the deferral menu's lifecycle for leaks: `_closeDeferMenu` runs at the top of `_render` and in `disconnectedCallback`, and `DeferMenus` holds document-level `keydown` and `click` handlers. Is there a path that replaces the markup without closing the menu, or that binds a second pair of handlers over the first?
- **Semantic merge gaps.** Two clean textual merges were wrong: a call to a `_toast` method `main` had deleted, and four handlers hand-rolling what `main`'s new `_store_errors` does. Look for others I missed — places where both sides changed behaviour and git took both without conflicting.
- **The skip log.** `skips` is a separate list from `completions`. Verify nothing that reads `completions` needs to read skips too, and nothing that reads both double-counts — the completion tally, the cadence average, `last_completed`, sensor attributes, the appliance archive.
- **The meter fix.** `sensor_tasks.latest_decision_ts` is the later of a completion and a skip, and `backstop_due` measures from it. Check the undo paths: does deleting a skip restore the right baseline when a completion and a skip interleave, and is `meter_start` always the value it replaced?
- **The three-way JSON merge.** The 33 locale and translation files were merged by key from the git stages rather than textually. Please spot-check for keys dropped, duplicated, or taking the wrong side.
- **Options and defaults.** `allow_snooze` / `allow_skip` default **true**, which is unlike every other option. Confirm an install with neither key stored behaves exactly as it did before, on every reader: the panel, the card, `actions_for`, and the options flow's merge.
- **HA best practices and security.** The five new websocket commands and their service twins — is each gated the same way, and does `call_service` reach anything the websocket command gates?

---

prestomation:
Pushed `e334240` for the red `vale` gate.

`HomeKeeper.Personification` landed on main while this branch was open, and the gate is diff-scoped, so the seven `carry`/`carries` uses in the prose this PR adds went red. Reworded rather than silenced — the rule is a house style decision, and each of the seven was reaching for "carry" where a plainer verb says the same thing:

| File | Was | Now |
| --- | --- | --- |
| `README.md` | **Done** carries a caret | **Done** has a caret |
| `README.md` | carries a note and whoever decided | records a note and whoever decided |
| `README.md` (alt text) | whose rows carry a snooze and a skip button | whose rows show a snooze and a skip button |
| `CHANGELOG.md` | Done carries a caret | Done has a caret |
| `docs/EVENTS.md` | `_task_skipped` carries the new entry's `ts` | `_task_skipped` includes the new entry's `ts` |
| `locales/en.json` ×2 | let a notification carry a Snooze/Skip button | include a Snooze/Skip button on notifications |

Checked the way AGENTS.md prescribes for a rule that reports per *block* rather than per line: rather than filtering hits by added line number — which is what let a red `vale` ship on this branch once already — I ran vale over every scoped file on this head and on a worktree of `origin/main`, and compared the hit *sets*. **79 on each, zero difference.** The `Personification` hits that remain in `CHANGELOG.md` are all in released sections from line 280 down, untouched here and equally present on main.

Also re-run locally: `pytest tests/unit` (1920 passed, 2 skipped), `ci/test-frontend.sh` (1130 passed, 36 files), `tsc --noEmit`, `ruff check` / `ruff format --check`.

**One thing left, and I could not do it here:** the two reworded strings are Settings help text, so `docs/images/17-panel-settings.png` now shows the previous wording. Recapturing needs the Docker HA harness (`ci/e2e-up.sh`), and this session's container has no Docker daemon, so `docker info` fails and the harness cannot start. The stale text is two help lines under **Skip &amp; snooze**; everything else in that capture is current. Happy to recapture in a session that has Docker, or it can ride along with the next capture pass.

---

prestomation:
/q review Please review `e334240` critically and skeptically, most serious findings first, and do not withhold minor ones. It is a prose-only change clearing the `HomeKeeper.Personification` vale findings.

Scrutinise in particular:

1. **Meaning drift.** Each reworded sentence should say exactly what it said before. "carries a note and whoever decided" became "records a note and whoever decided" — for a skip entry, is "records" accurate given the note is supplied by the caller rather than derived? Likewise "Done has a caret" versus "Done carries a caret": does "has" still convey that the caret is part of the Done control rather than a separate button beside it?

2. **New rule violations introduced by the rewording.** The `ai-tells` rules apply per block, so check whether any reworded sentence trips `EmDashUsage`, `VerbTricolon`, `StackedAnaphora`, `ContrastiveFormulas` or `OverusedVocabulary` in the paragraph it sits in, not just on its own line.

3. **String changes.** `locales/en.json` `settings.allow_snooze_help` and `settings.allow_skip_help` changed text. Is there anywhere that asserts on or duplicates that copy — `strings.json`, a frontend test, a docs page, or the other locale files — that should have moved with it?

4. **Consistency.** README, CHANGELOG and EVENTS describe the same behaviour. After the rewording, do the three still agree with each other and with the code?

---

prestomation:
Triage of the two "must fix before merge" findings. **Both are false**, and `d365b5c` pins each as a test rather than leaving the answer in a comment.

### 1. "Menu leak: `_closeDeferMenu` is not called in `disconnectedCallback`"

It is — `panel.ts:447`, three lines into the method, with a comment saying why:

```ts
disconnectedCallback(): void {
  teardownOverlay(this);
  this._closeDeferMenu();
```

### 2. "Menu double-binding on re-render"

`_render()` ends at `this.shadowRoot.innerHTML = …` (`panel.ts:1421`), which discards every node in the tree. The `wire()` on the next render binds carets that did not exist a moment earlier, so nothing accumulates — that is what the "Safe to call on each render" note on `DeferMenus.wire` refers to. The predicted symptom ("a click would fire N times after N renders") does not occur; the doubled-handler case would in fact make one press *toggle twice*, opening and closing the menu in the same tick.

A file/line note for future runs: `DeferMenus` is in `defer-dialogs.ts`, not `panel-defer.ts`, and the line numbers cited for it do not correspond to that file.

### Why this still produced a change

Neither property was asserted anywhere, and an invariant no test holds is one the next reader re-derives from the source — which is how the claim arose. Two tests in the deferral `describe` block now hold them:

- **`takes its document handlers back off when the panel unmounts`** — wraps `document.addEventListener` / `removeEventListener`, opens the menu, unmounts, and asserts every type added was removed the same number of times.
- **`still opens on one press after repeated re-renders`** — three `_render()` calls, then one caret press must leave the menu open.

Both were checked for vacuity by breaking the code each one guards:

| Break | Result |
| --- | --- |
| Comment out `_closeDeferMenu()` in `disconnectedCallback` | `expected +0 to be 1` — every document 'keydown' handler the menu added is taken back off |
| Call `_wireDeferMenus` a second time on an already-wired tree | `expected true to be false` — the menu is hidden again after one press |

The second reproduces exactly the double-binding the review described, so it would catch that regression if someone ever introduced it.

Full local run on `d365b5c`: `ci/test-frontend.sh` 1132 passed (36 files), `pytest tests/unit` 1920 passed / 2 skipped.

The verified-correct sections of the review (skip-log separation, meter baseline restore, options defaults, websocket/service parity) match what I found independently, so nothing to add there.

---

prestomation:
`vale` is green on `d365b5c`, along with `ruff`, `mypy`, `hacs` and `publish`; the longer tiers are still running. Triage of the review of `e334240`:

### 1. "Restore `carries` in both places" — cannot, and that is the whole point of the commit

`HomeKeeper.Personification` matches `\bcarr(?:y|ies|ied|ying)\b` at **error** level, and the gate is diff-scoped onto exactly these lines. Restoring either word puts the branch straight back to the red `vale` this commit was fixing — the same failure reported on this PR earlier today at README:371 and README:389, the two lines the suggestion names. The rule is a deliberate house style decision that landed on main while this branch was open, so "carries is more precise" is an argument for changing the rule, not for reverting past it here.

On the substance: for the caret, "intrinsic part of the Done control" is what `has` says, and the DOM agrees — the caret is a child of the `.hk-split` wrapper, not a sibling. `logs a note` instead of `records a note` is a fair alternative and rule-safe, but the sentence already ends "…and can be edited, re-dated or undone exactly like a completion row", which settles the provenance question, so I have left it.

### 2. No new rule violations — agreed, and independently confirmed

Matches my own check, which compared the full vale hit *set* for every scoped file against a worktree of `origin/main`: 79 on each, zero difference.

### 3. String changes — both claims here are wrong

> ✅ `custom_components/home_keeper/strings.json` has matching changes

`strings.json` contains no `allow_snooze` or `allow_skip` key at all — `grep -c` returns 0. These two options are panel-only: they are absent from `options.FLOW_OPTIONS`, so the Configure dialog never renders them and there is no HA-side label to keep in step. There was nothing to change there and nothing was changed.

> ⚠️ The 15 other locale files all still have the old text and need updating

They hold **translations**, not the old English. The behaviour being described did not change, so none of them is now wrong:

| | |
| --- | --- |
| `de.json` | *Verschieben an einer Aufgabe anbieten und in Benachrichtigungen als Schaltfläche zulassen.* |
| `ru.json` | *Показывать «Отложить» у задачи и разрешать кнопку «Отложить» в уведомлении.* |
| `fi.json` | *Tarjoa tehtävällä Lykkää ja salli ilmoituksen Lykkää-painike.* |

German, Russian, Finnish and Czech never used the carry metaphor to begin with. Others (`fr`, `it`, `nl`, `pl`, `es`, `sv`, `nb`, `da`, `pt-BR`, `ca`, `zh-Hans`) do calque it — *porter*, *portare*, *dragen*, *nieść* — and a translator may well want to revisit those, but the English house rule does not make a French sentence incorrect, and hand-editing eleven languages I cannot verify would trade a style preference for a real risk of mistranslation. `i18n-parity.test.js` covers key parity and passes.

✅ on the other two: no frontend test asserts this copy, and README/CHANGELOG/EVENTS agree with each other and with the code.

### 4. Consistency — agreed.

No change from this review. The one item still open is the stale Settings screenshot noted above, which needs a session with a Docker daemon.

---

PR #274, opened 2026-08-31
refactor: factoring and dedup pass — split panel.ts, unify twin code paths

prestomation:
## What changed

A factoring and deduplication pass across the whole codebase: find code that does the same thing twice, keep one copy, and split the panel monolith into modules that each carry one idea. Behavior is preserved except three deliberate alignments where the twin code paths had drifted and a user would expect them to match.

**Behavior alignments** (the reason for the `0.19.0b7` bump; CHANGELOG has the user-facing story):
- The panel's websocket `complete_task` now settles buy tasks like the service, button, and to-do paths always did — a completed auto-buy task no longer leaves its reminder behind until the next reload.
- `export_inventory` localizes its CSV like the websocket twin; `ws_delete_asset` now actually shares `_delete_asset` (its docstring claimed it already did).
- Three websocket completion-editing errors report `invalid_task` instead of `not_allowed` (machine-readable code only).

**Backend dedup** — the same shape, kept once:
- `__init__.py`: a `_store_errors` context manager replaces ~37 pasted exception-translation blocks. `websocket_api.py`: a `@_with_coordinator` decorator replaces 28 coordinator preambles and most of 34 `_err(); return` blocks (−137 statements in that file alone).
- One `find_coordinator`, one `assets.find_part` (was 9 hand-rolled loops), one `store._mutate_asset` scaffold behind the five document/part-file mutations (`remove_part_file`'s idempotent no-op preserved), a shared `HomeKeeperPartEntity` base and registry-prune helper for the four platforms, shared internals for `manuals.py`'s two HTTP views (their genuinely different error-recovery tails kept apart).
- New pure `todo_items.py` (the item-matching helpers both to-do syncs had byte-identical copies of; on the mutation surface at 100%) and `todo_sync_driver.py` (the HA-facing base under both sync drivers; `plan_sync` semantics deliberately stay per-sync, and `problem_sync.py` is deliberately not a subclass).

**Test-suite dedup**: seven unit files each hand-rolled the same HA-stub bootstrap; now one `ha_stubs.py` + `fakes.py`, verified identical pass counts per-suite, in aggregate, and across load orders, in both the HA-present and HA-free lanes.

**Frontend dedup**: the panel adopts `card-filter.ts`'s bucketing (the per-surface differences ride on `statusBucket` options — panel has no `today` bucket and adds `completed`; defaults reproduce the card); pure helpers moved to their real homes (`utils`, `documents`, `forms`) with mutation-killing tests — 100% Stryker score on every touched surface.

**The panel split**: `panel.ts` goes **8,559 → 1,706 lines**. The element (fields, routing, `_render`, `_hydrate`, form lifecycles) stays in `panel.ts`; everything else is a flat `panel-*.ts` region module of free functions over an explicit `PanelHost` interface — the coupling surface is now a single, reviewable list. Along the way the settings scaffolds (6 hand-rolled copies), the `createElement('ha-form')` sites (5→1), the confirm-scrim teardown (3→1), the file-card/file-picker/document-mutation envelopes all collapsed to one implementation each. The conventions are recorded in `.amazonq/rules/architecture-and-code.md` ("The panel is one element plus flat region modules").

**Proof of behavior preservation**: all 205 e2e browser tests pass against the split panel; the region moves were verified by parsing the built bundle and diffing every HTML template literal (zero differences beyond the two deduplicated copies); 1,581 Python unit tests and 815 vitest tests green; mutation score 100% on every diff-scoped surface (98.6% on the one Python run that pulled in pre-existing planner debt).

Net: product logic shrank by ~270 executable statements while gaining ~20 documented modules; `custom_components` Python is net −157 lines with all the new docstrings.

One note for reviewers: `tests/e2e/screenshots.capture.ts` fails at its step 1d (the managed task's edit drawer: the form opens, then closes before `#f-cancel` is clicked) **identically on `main` and on this branch** in this environment — verified by running the capture against a `main`-built bundle on a pristine container, and with both the full Chromium and the Playwright headless shell. Direct interaction with that drawer works (probes pass, e2e green), so it looks like a sequence-dependent race in the capture harness, out of this PR's scope. The screenshots below are from the harness run up to that step plus targeted captures of the Settings and Appliance surfaces.

### Screenshots (current, from this branch)

Task list after the split

Task detail with the edit drawer

Settings tab

Appliance detail with parts

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`, `bash ci/test-frontend.sh`, full e2e suite, diff-scoped mutation runs)
- [x] `CHANGELOG.md` updated — user-facing changes only (the behavior alignments, under `0.19.0b7`)
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with HTML `` tags
- [x] New user-facing UI surface → n/a (no new surface; the walkthrough tour needs no edit)
- [x] New user-facing feature → n/a (bug-fix alignments only); version bumped to `0.19.0b7` for the alignments

https://claude.ai/code/session_0199CFHtdRnJcJMJ4Z5qXzJa


prestomation:
/q review This is the opening slice of a staged factoring pass — only the behavior alignments have landed so far (websocket `complete_task` now settles buy tasks; `export_inventory` localizes its CSV; `ws_delete_asset` delegates to `_delete_asset`; version bump to 0.19.0b7). Please review these critically and skeptically, most serious issues first, without withholding minor ones: **correctness** of the settle-vs-refresh change (can `async_settle_buy_tasks` from the websocket path race a reload or double-settle when the panel fires several completions quickly?), **error paths** (does the shared `_delete_asset` change any error surface the old inline copy had?), and **HA best practices** for calling `hass.config.language` at service-call time. The larger dedup commits will follow on this branch — flag anything here that would make those riskier.

---

prestomation:
/q review The full factoring pass has now landed and the PR is ready for review. Please review critically and skeptically, most serious issues first, without withholding minor ones. Priorities: **correctness of behavior preservation** — the `_with_coordinator` decorator and `_store_errors` context manager must reproduce every handler's old error code/key/placeholder exactly, `store._mutate_asset` must keep `remove_part_file`'s idempotent no-op, and the `panel-*.ts` extraction (free functions over `PanelHost`) must not have changed any emitted HTML, wiring order in `_hydrate`, or the `_liveHassEls`/`_assetEdit.asset` mutation contracts. **Maintainability** — is `PanelHost` a reasonable coupling surface or is anything on it that should be an argument; are the `TodoSyncDriver` base's shared/per-sync boundaries right. **HA best practices** — the settle-vs-refresh change in `ws_complete_task`, the shared `find_coordinator`, and the entity platform prune helper. Also sanity-check the `statusBucket` option flags against the panel's old bucketing (no `today` bucket, `completed` for done one-offs).

---

PR #276, opened 2026-09-01
fix(panel): theme the native dropdown popup, not just the closed control

prestomation:
Fixes #

## What changed

The panel's dropdown menus (`.hk-menu-select` / `.hk-profile-select` — the Profile
filter, Group by, and other segmented pickers) rendered their open popup list in the
browser's default grey-on-white, even though the closed control itself is themed.
Chromium and Firefox style a native `<select>`'s open-state popup from the color and
background-color set on its `<option>` elements, not the select's own (often
transparent) styling, so without an explicit value on the options the popup fell back
to the OS default — unreadable against a dark Home Assistant theme, as reported.

Added `color`/`background-color` on the `<option>` elements for both select classes in
`panel-styles.ts`, sourced from the same `--hk-surface` / `--hk-ink` tokens the rest of
the panel uses, so the popup matches the panel's theme instead of the OS default.

Also extended `tests/e2e/screenshots.capture.ts` to open the Profile dropdown's native
popup before selecting a value, so the screenshot below is reproducible via the normal
capture harness (`npx playwright test --config=screenshots.config.ts`) rather than a
one-off script.

Profile dropdown open, showing themed dark popup with white text

## Checklist

      step through it — not applicable, this is a styling fix to an existing surface
- [ ] New user-facing feature → version bumped to the next beta (`manifest.json` +
      `const.py`), `preview-release` label applied — not applicable, bug fix only

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qnga84CsggQyqvtH35FrV4)_


prestomation:
/q review Please review critically and skeptically, focusing on:

- **Correctness**: does styling `<option>` via `color`/`background-color` reliably theme the native popup across the browsers HA users actually run (Chromium-based, Firefox, Safari/WebKit)? Any known gaps where this silently no-ops and the popup reverts to OS default?
- **Maintainability**: is scoping the fix to `.hk-menu-select`/`.hk-profile-select` sufficient, or are there other native `<select>` elements in the panel (e.g. inside `ha-form`-rendered fields) that share this problem but aren't covered by this selector?
- **HA best practices**: is deriving the option colors from `--hk-surface`/`--hk-ink` (rather than HA's own `--card-background-color`/`--primary-text-color` directly) the right call for a component meant to track the active HA theme, including high-contrast/custom themes?

Surface the most serious issues first; don't hold back minor nits either.

---

prestomation:
Thanks for digging in — responding to each:

**1. Safari/WebKit.** Confirmed: this is a documented cross-browser limitation, not something this CSS can work around. MDN's own `<option>` reference notes Safari largely ignores styling on `<option>` elements. It's not a regression, though — Safari's native `<select>` popup is OS-rendered (not a lightweight in-page popup the way Chromium's is), so it already follows the system's light/dark appearance without any CSS from us. The bug as reported (and the screenshot in the PR) is Chromium's flat grey-on-white, which this fix addresses. Getting pixel-identical theming on Safari too would mean replacing the native `<select>` with a custom-built listbox component — a much bigger change or than this contrast fix calls for. Happy to open a follow-up if it turns out Safari users are actually seeing a problem in practice, but there's no minimal-diff fix for it here.

**2. Other `<select>` elements.** Checked — `.hk-menu-select` and `.hk-profile-select` are the only two native `<select>` elements in the panel's own markup (`panel-controls.ts`, used for the Profile filter and the segmented Group-by/filter menus). Everything else that looks like a dropdown in the panel is either `ha-form`-rendered (HA's own `ha-select`/`mwc-select` Material components, not a raw `<select>`, and already themed by HA itself) or the custom `.hk-seg` button group. So there's no uncovered instance of this bug today.

**3. `--hk-ink` on `--hk-surface` contrast.** This is the panel's existing default text/background pairing — the same tokens used for ordinary body text everywhere else in the panel (task names, form labels, etc.), not something new introduced here. If there's a contrast gap in a specific HA theme it'd be a pre-existing issue affecting the whole panel, not something scoped to this fix — out of scope here, but open to a dedicated audit if you're seeing a concrete case.

---

PR #277, opened 2026-09-01
feat(notifications): channel, urgency, and a Test button

### LINKED ISSUE #255: Feature Request: Notification Channels

prestomation:
Fixes #255

Supersedes prestomation/ha-home-keeper#266. Rebased onto current `main` (no merge commit), with the notification prose re-drafted into the ASD-STE100 voice `main` adopted in #285, and a **Test** button added.

## What changed

Three things on a saved notification — **Notification channel**, **Urgency**, and a **Test** button — plus the three worked automations the other half of #255 asked for.

The Settings, Notifications card showing a walk on the Chores channel at High urgency, with helper text under each new field and a Test button beside Delete

### Channel and urgency: is this an Android-only thing?

Partly, and the docs now say so. Channels are Android's model and iOS has none, so the two fields are stored **platform-neutral** and expanded into both vocabularies at payload-build time by a new `notifications.payload_data`. Each app ignores the keys it does not know, so there is no per-target branching in `notifier.py` and the panel never asks which phone the household owns.

| Urgency | Android | iOS |
| --- | --- | --- |
| `quiet` | `importance: low` | `push.interruption-level: passive` |
| `normal` (default) | *nothing emitted* | *nothing emitted* |
| `high` | `importance: high`, `ttl: 0`, `priority: high` | `push.interruption-level: time-sensitive` |
| `critical` | `importance: max`, `ttl: 0`, `priority: high` | `push.interruption-level: critical`, `push.sound: {name: default, critical: 1, volume: 1.0}` |

The channel name goes out as Android `data.channel` and iOS `data.push.thread-id`. `data.group` stays the literal `home_keeper` — the reporter said grouping already works, and `test_build_notification_walk_actions_and_tag` pins that string as a payload contract.

`normal` is deliberately absent from both mapping tables, so **an unconfigured notification sends byte-for-byte what it sent before these fields existed** — asserted, not assumed (`test_payload_data_unconfigured_is_what_it_always_was`).

### The Test button

Delivery is the one part of a notification the panel cannot show: the channel, the urgency and the target only mean something on the phone, so the only way to check a change was to wait for a task to come due. **Test** sends the notification now.

Three decisions worth review:

- **It goes through `home_keeper.notify`, not a websocket command of its own.** That is a deliberate exception to services-first, and the reason is the feature: a test that took a different path from the real delivery would prove nothing about the real delivery. `api.runNotification` is the panel's only service call. `.amazonq/rules/architecture-and-code.md` records the exception *and its limit* — a panel action a service does not already cover still lands as a service first.
- **The row is saved before the send.** `notify` resolves the notification out of stored options, so an edit still sitting in the form's debounce would test the *previous* channel and report success. A test asserts the ordering, not just the outcome.
- **`sent: 0` gets its own message.** It means the filter matched nothing, which is a successful run — reporting it as an error would send somebody hunting for a delivery problem that is not there.

No new service, event, or websocket command, so `api_surface.py` is unchanged.

### The other half of #255: re-sending until it's done

Left as an automation, per your call on the issue. The Notifications page now carries three worked examples: re-sending every 2 hours, holding notifications back until somebody is home, and a tight window at Critical urgency. `home_keeper.notify` sends nothing when nothing matches, which is what makes a `time_pattern` trigger safe, and the docs now say so.

## What the rebase changed

- **`README.md` was reset to `main`'s and the additions re-drafted.** #285 rewrote the README wholesale into STE *and reordered its sections*, so git could not align the old block at all. The new prose is `### Channels and urgency` and `### Automation examples` written in the new voice, not the old text replayed.
- **The two `notify.*_help` strings were rewritten in STE, and so were their 15 translations.** `locales/en.json` is inside the vale gate's scope, and the originals were pre-STE ("reminders" for "notification", "how loudly this lands"). Leaving the translations describing the older English would have been the worse half of the job.
- **Dropped `7de08c8`** ("rework the notification prose the vale gate flagged"). It only patched vale hits in prose the re-draft replaced; nothing factual in it is missing.
- **The version bump targets `0.20.0b3`.** `v0.20.0b1` (#278) and `v0.20.0b2` (#282) are both published tags.
- **The screenshot was re-captured twice.** The one on #266 predated the Settings rail polish (#270), the button-weight pass (#273) and the dropdown theming (#276) — and the committed one no longer matched this branch's own `en.json`, so it was stale before the rebase started.

## Notes for review

- **No new API surface.** `notifications` is already an opaque `list` in `SET_OPTIONS_SCHEMA`, the options flow never renders notifications, and `api_surface.OPTIONS` indexes the option key rather than its sub-fields. No `services.yaml` / `strings.json` / `translations/` change.
- **No migration.** `options.current_options` re-normalizes on every read, the same route that let `exclude_*` and `sync` ship without one.
- `itemCard` grew a footer row (`.hk-item-actions`) so a notification keeps Delete where it has always been and puts the safe action to its left. A profile row still has Delete alone, and a test asserts that.

## Checklist

- [x] Tests run locally (`pytest tests/unit`, `bash ci/test-frontend.sh`)
- [x] `CHANGELOG.md` updated, with `(Fixes #255)` in the bullet's first paragraph
- [x] Panel UI changed → current screenshot committed under `docs/images/` and embedded above with an HTML image tag
- [x] New user-facing feature → version bumped to `0.20.0b3` (`manifest.json` + `const.py`), `preview-release` label applied

**On the walkthrough box, now checked:** the tour used to walk *past* Notifications on the settings rail without opening a row. It now seeds a notification, opens it, and frames the delivery fields and Test.

### Verification

- `pytest tests/unit` — 1606 passed, 2 skipped
- `bash ci/test-frontend.sh` — 858 passed (28 files), including 6 new tests for the Test button
- `npx tsc --noEmit` — clean
- `ruff check` + `ruff format --check` — clean, 157 files
- `python3 ci/check-changelog-release-gap.py` — clean
- Playwright, real HA container, **headless shell (no `CHROMIUM_EXEC` override)**: `settings.spec.ts`, 11 passed across all three viewports
- Screenshot re-captured on a clean container and visually inspected before committing
- `vale` — 0 hits on the lines this branch adds, checked two ways: `--output=JSON` intersected with the added-line set, and the rule regexes in `styles/*/` matched by hand against reconstructed markdown *blocks* (the diff-scoped run is what CI gates on; the file-wide hits are pre-existing)

Two things worth flagging rather than burying:

**CI's vale caught a `VerbTricolon` that a local Vale 3.9.1 run did not** — the exact discrepancy `AGENTS.md` warns about. The modal arm matches `[^,]+` across sentence ends, so two commas after one `must`, three sentences apart, read as a tricolon. Fixed in `7026d7c`, and I added a by-hand block-level regex check to the loop because the local binary cannot be trusted on this rule.

**A second screenshot capture in one session fails on the first-run intro.** `frontend.user_data_*` is gitignored, so neither `docker compose down -v` nor `git clean -fd` clears it, and it carries `home_keeper_intro_dismissed` from the previous run. `git clean -fdx tests/integration/ha_config/` is the reset that works. CI never hits this because it starts from a fresh checkout.

`mypy` could not run here — this container is on Python 3.11 and Home Assistant needs 3.14+, so every error was a missing import. CI runs the real thing.

https://claude.ai/code/session_01SPgLqKAJ1zaDUQ1a35zcfE


prestomation:
/q review Be critical and skeptical, and surface the most serious issues first — but do not withhold the minor ones. This is #266 replayed onto a `main` that has since shipped 0.19.0 and split `panel.ts` into region modules, so please look hardest at what the rebase could have quietly broken.

**Correctness.** `notifications.payload_data` is the new single source of the mobile-app `data` block for `build_notification`, `build_digest` and `build_all_clear`. Does an unconfigured notification (`channel: ''`, `urgency: 'normal'`) really emit the exact same dict those three built before, key for key, including when `actions` is `None` versus an empty list? Check the Android/iOS key mapping against the companion app's documented schema (`channel`, `importance`, `ttl`, `priority`, `push.thread-id`, `push.interruption-level`, `push.sound`) — wrong key names fail silently on a phone and no test can see it. Also check `normalize_notification`: a stored notification predating these fields has neither key, and the panel round-trips whatever it reads back through `set_options`.

**Rebase-specific.** The form's helper text moved from `panel.ts` to `panel-settings.ts`, passed through `_makeForm`'s `labelling.computeHelper` instead of being assigned onto the element. Is that path equivalent — does supplying `computeLabel` + `computeHelper` there lose the `help.<field>` default for any field in `notificationSchema` that previously had one? And is `addNotification`'s blank object still complete against the `Notification` type after both new fields became required?

**Maintainability.** Is `payload_data` the right seam, or does the `push` sub-dict assembly belong in its own helper? Are `_ANDROID_IMPORTANCE` / `_IOS_INTERRUPTION` / `_WAKE_URGENCIES` / `_IOS_CRITICAL_SOUND` overlapping in a way that will drift when a fifth urgency is added?

**HA best practices.** `notifications` stays an opaque `list` in `SET_OPTIONS_SCHEMA` with no `services.yaml` / `strings.json` / `api_surface.py` change, on the reasoning that the option key rather than its sub-fields is the declared surface. Is that right, or does adding user-configurable sub-fields cross the line into something the API surface should declare? Separately: is emitting both platforms' keys on every payload safe, or can an unknown key make either companion app drop the notification?

**Security.** The channel name is user-authored free text that reaches a notification payload. Anything to worry about there?

---

prestomation:
/q review `main` moved under this PR (#278 and #282 landed), so this pushes a merge of `main` into the branch. Be critical and skeptical, surface the most serious issues first, and don't withhold the minor ones. Focus on what the merge could have quietly broken rather than re-reviewing the feature you already passed.

**Correctness of the merge itself.** Three files conflicted — `CHANGELOG.md`, `manifest.json`, `const.py` — and everything else auto-merged. The riskiest auto-merges are `forms.ts` (#282 added ~56 lines of duplicate-task schema right beside the notification schema this PR edits) and `tests/e2e/screenshots.capture.ts` (#282 added a duplicate-drawer capture step next to the notifications one). Please check that neither lost a hunk from either side, and that `notificationSchema`'s field order and `notifyFormData`/`notifyFormToNotification` round-trip are still intact after git's three-way merge.

**Version and release bookkeeping.** `main` had already taken `0.20.0b1` (#278) and `0.20.0b2` (#282), so this moves to `0.20.0b3` in `manifest.json` + `const.py` with a matching CHANGELOG section. Is the resulting CHANGELOG ordering correct — `0.20.0b3` above `0.20.0b2` above `0.20.0b1` above the released `0.19.0` — and does `(Fixes #255)` sit in the first paragraph of its bullet where `ci/release-issues.py` will find it? Also: does `tests/unit/test_release_issues.py`'s `_KNOWN_ISSUES` correctly carry `255`, `262` and `279` after the merge, rather than one side clobbering the other?

**Regression risk from #282.** #282 touched `panel-styles.ts` and `panel-host.ts`. Does anything there change how `_makeForm`'s `computeHelper` renders, or how a Settings item card lays out, in a way that would affect the notification editor's two new fields?

**Anything else.** If the merge introduced duplicated logic or a now-redundant construct in the files both sides touched, name it.

---

prestomation:
/q review Be critical and skeptical, surface the most serious issues first, and do not withhold the minor ones. This head is a true rebase onto current `main` (the merge commit is gone) plus a new **Test** button, so look hardest at the two things that changed shape rather than re-reviewing the channel/urgency payload work you already passed.

**The Test button — correctness.** `api.runNotification` calls `home_keeper.notify` over `hass.callWS({type: 'call_service', …, return_response: true})`. Is that the right websocket envelope, and is `res.response` the right place to read `{matched, sent}` from? `testNotification` awaits `api.setOptions` *before* the send, to beat the form's debounce — but `persistDebounced` still has a pending timer holding the *same* list, so does that later write race the save, and can it clobber anything? `notifyFormToNotification` runs on every `value-changed`, so `current` is rebuilt from form data rather than merged onto the stored row: does any field of `Notification` fail to survive that round-trip and get silently dropped when Test saves? And `handle_notify` has no `_verify_admin` while the panel is admin-only — is calling it from the panel a privilege question at all, or is it correctly a read-only-ish operation any user can invoke?

**The Test button — architecture.** This is the panel's only service call, against a rule that says services are the contract and websocket commands are the optimization. I argued a websocket twin would be worse *for a test button specifically*, and wrote that exception plus its limit into `.amazonq/rules/architecture-and-code.md`. Push back if that reasoning is wrong, or if the rule as written is too loose and invites a future action to skip the service.

**The Test button — UX edge cases.** A notification with no target fails with a localized `ServiceValidationError`; the button surfaces the message and does nothing else. Should it instead be disabled, the way **Add notification** is disabled with no profiles? Is `sent: 0` → "No task is due" the right split, or does it hide a real misconfiguration (a profile that filters everything out looks identical to a genuinely clear list)? And the button has no in-flight state, so a double-press sends twice.

**The rebase.** `README.md` was reset to `main`'s and the additions rewritten rather than replayed, because #285 rewrote *and reordered* the file. Please check nothing from the original #266 prose was lost in that: the Android channel-immutability caveat, the iOS Critical Alerts permission, the empty-channel default, and all three worked automations should still be there and still accurate. I also dropped commit `7de08c8` as obsolete — confirm nothing factual went with it.

**Translations.** I rewrote the two `notify.*_help` strings in STE and hand-updated all 15 translations. Check the placeholder-free strings for meaning drift, especially `ru`/`pl`/`cs`/`fi` where I am least confident, and check `notify.test` / `notify.test_sent` / `notify.test_none` read naturally rather than as calques of the English.

**Anything else.** `itemCard` grew a `.hk-item-actions` footer that both profiles and notifications now render through — did that change any layout or a11y affordance beyond the two rows I checked?

---

prestomation:
Thanks — triaged. One taken, one pushed back on, and chasing the second one found a flake worth reporting.

**In-flight state — taken, already pushed.** Fixed in `09ab097`, which landed before the review arrived, so you were reading the older head. The button disables itself for the round trip and drops presses arriving while one is in flight, released in a `finally` so a failed send stays retryable. Two tests cover it, one gating the stub's response on a deferred promise so the in-flight state is observable rather than raced.

**Disable Test when no profiles exist — I don't think so, and I'd rather not.** The analogy to **Add** does not hold. Add is disabled because with no profiles a new notification *cannot be constructed*: `addNotification` reads `profiles[0].id` for the blank row's `profile_id`. That is a real precondition. An existing row already carries a `profile_id`, so pressing Test on it is always a well-formed request.

What happens instead is better than a disabled button. Every failure path returns a localized `ServiceValidationError` that the panel toasts verbatim:

- no target → *"This notification has no target to send to. Add a 'Send to' device in Settings → Notifications, or pass target: when calling the service."*
- dangling `profile_id` → *"No notification profile named {profile}."*

The first one tells the user exactly what to do. A disabled button replaces that with silence, and silence is the failure mode people file issues about. So the button stays enabled, and `b81e01f` adds an e2e assertion that the message actually reaches the user — the reasoning is now pinned by a test rather than living in this thread.

**A flake I introduced, since it is worth naming.** Writing that assertion made the spec fail on the first run and pass on the retry. Two separate cold-boot hazards, both already known in this repo and both new to this spec:

1. Home Assistant raises its own *"Home Assistant has started!"* toast into the same slot on a cold boot, so asserting straight after the click reads that toast's text. `shots.ts` already has `settleToasts` for exactly this.
2. Settling *after* the row is open gives Home Assistant time to replace the custom-panel element, and a fresh panel folds every row — so the click then finds nothing. That is the same hazard `walkthrough.capture.ts` guards its row-opens against.

Fixed by settling before the row is open and guarding the open. Verified on two consecutive cold containers, 11 passed with no retries; a warm container never reproduced either failure, which is why the first "green" run proved nothing.

**On the two you confirmed** — the `call_service` envelope / `res.response` path, and `handle_notify` correctly having no admin gate because sending is usage rather than administration — thank you, those were the two I was least sure of.

**Still open for a human:** the `ru` / `pl` / `cs` / `fi` translations of `notify.test*` and the two rewritten `notify.*_help` strings. You flagged that you did not examine them, and I wrote them, so neither of us has actually checked them.

---

prestomation:
/q review Second pass on the two commits pushed since your last review: `09ab097` (the in-flight guard you asked for) and `b81e01f` (an e2e assertion plus a flake fix). Be critical and skeptical, and focus on these rather than re-reviewing what you already passed.

**The in-flight guard.** `itemCard` now wraps `onTest` in a click handler that returns early if the button carries `disabled`, sets it, and clears it in a `finally`. Is the `disabled` attribute the right in-flight flag on an `ha-button`, or can the component's own rendering drop or ignore it and let a press through? The guard is per-button-element rather than per-notification: if the panel re-renders mid-flight (a config-entry reload from the `set_options` this very handler triggers), a fresh button is built without the attribute — can a re-render land inside the round trip and hand the user an enabled button over an in-flight send? And `onTest` is now typed `() => Promise<void>` while `onDelete` stays `() => void`; is the asymmetry justified or should both return promises?

**The flake fix.** The spec now calls `settleToasts` before opening the row and re-opens the row through a guarded `openRow()` before the click. I convinced myself the second failure was Home Assistant replacing the custom-panel element during the settle. Is that actually the mechanism, or did I paper over a slower race that will come back on a loaded CI runner? Specifically: is `openRow()`'s read of `aria-expanded` followed by a conditional click itself racy against a swap landing between the read and the click? Two consecutive cold-boot runs passed with no retries, but that is a weak sample.

**The error-path assertion.** The test asserts on `.message` for the service's localized text. Is that selector stable, or is it matching something incidental? And the test now depends on `notify_no_targets`'s exact wording in `strings.json` via a regex — is that too tight a coupling to a user-facing string that a copy pass could reword?

**Anything the two commits broke.** `.hk-item-actions` and the `openRow` helper are the shared surfaces; say if either has a consequence I have not covered.

---

PR #278, opened 2026-09-01
fix(todo-sync): hold an unconfirmed add instead of adding it twice

### LINKED ISSUE #267: [FEAT] Caldav syncing
Wondering if there are any plans to support syncing tasks to/from a caldav server. My understanding is caldav supports todo/tasks like this and having it synced to the same server other calendar information is on would be ideal

prestomation:
Related to #267 — but deliberately **not** `Fixes #267`. The reporter's symptom was never reproduced; this fixes a different, real bug found while investigating theirs. A false reference would make `notify-issues` close their issue on release.

## What changed

Exploratory testing against a **real Nextcloud 34.0.3** (HA 2026.8.3, built-in `caldav`) turned up two bugs on one precondition, both now fixed. Findings are written up in `docs/CALDAV_SYNC_FINDINGS.md`.

### The precondition

`todo.add_item` answers with nothing, so a fresh entry carries no uid and is matched by summary on a later pass. HA's CalDAV entity refreshes its cached items in a **fire-and-forget** task, where `local_todo` and `todoist` both `await` theirs:

```python
# caldav/todo.py
await self.hass.async_add_executor_job(partial(self._calendar.save_todo, **item_data))
# refreshing async otherwise it would take too much time
self.hass.async_create_task(self.async_update_ha_state(force_refresh=True))
```

Measured locally, `todo.get_items` did not show the item on **any** of five consecutive adds.

### Bug A — permanent duplicate (reproduced 4/4)

The next pass failed to resolve the item and, having no uid, read that as a failed add and added it again. The bookkeeping only ever points at one copy, so later edits moved one and deleting the task orphaned the other.

### Bug B — phantom completion (found in review, verified in source)

`resolve_tracked` falls back to the first summary match **even when it is ticked off** — which, after a recurring task completes, is the predecessor Home Keeper ticked off itself. `completed_since` is a strict `>`, so the pass read that as the household's tick and planned a second completion, skipping an occurrence and stranding the new item. Strictly worse than A.

### The fix

Both come from one unsound step: letting a read of a stale cache overturn what the write already established. An unconfirmed entry is now **held** — carried forward verbatim, since for a uid-less entry the summary is its only handle — and stamped `added_at`, until `UNCONFIRMED_GRACE` (20 min) expires and a genuinely lost add is repaired rather than leaving the task with no item. The same stamp gates the ticked-item arm.

Two decisions worth review:

- **Wall clock, not a pass count.** The driver runs four passes back to back with no delay, so "unseen for two passes" can elapse in milliseconds — inside the window being waited out.
- **20 minutes is a staleness budget**, calibrated to clear CalDAV's 15-minute `SCAN_INTERVAL`, not a formula. The mechanism depends on nothing upstream; only the number does.
- The stamp is **rewritten** rather than kept, because a value that is unparsable or in the future (a clock that jumped back before NTP corrected it) would never age out — turning "hold, never duplicate" into "hold, never deliver".

### Also in here

- **README**: says plainly that a to-do item you add on the list yourself never becomes a Home Keeper task, what does travel inbound, and the two CalDAV gotchas (the 15-minute poll; Nextcloud's default *Personal* calendar holds only events so it never appears as a to-do list).
- **New workflow rule** in `AGENTS.md` + `.amazonq/rules/`: user-facing text is drafted by a Sonnet 4.5 subagent, then reviewed. Dogfooded for the README and CHANGELOG here — I tightened both drafts for vale `ai-tells` risks.
- **Version bump to `0.20.0b1`** — required, not optional: `v0.19.0` is already tagged, so `check-changelog-release-gap.py` rejects editing that section and `release.yml` would skip a tag it already cut.

## Verification

- `pytest tests/unit` — **1591 passed**
- `ruff check` / `ruff format --check` — clean
- **Mutation gate — 97.29%** (395/406 mutants killed, threshold 80%), scoped to the four changed functions
- `mypy --strict` on `todo_list.py` — clean (the full mypy job needs HA, whose Python floor this container is below; CI covers it)
- **Against the real Nextcloud that produced the bug:**
  - Bug A: **0 duplicates across 4 rounds**, where every round previously produced 1–3. Server VTODO count matched the wanted-task count exactly, nothing missing.
  - Bug B: a completed recurring task recorded **exactly one** completion, with its next occurrence written and intact.

## Checklist

- [x] Tests run locally (`pytest tests/unit`, mutation gate, ruff) — plus a real CalDAV run
- [x] `CHANGELOG.md` updated — `## [0.20.0b1]`, no `(Fixes #N)` by design (see top)
- [x] Panel UI changed → n/a, no UI change
- [x] New user-facing UI surface → n/a
- [x] New user-facing feature → n/a; bug fix, so no `preview-release` label

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -991,10 +991,22 @@ The appliance/asset feature lives in `assets.py` (pure model — no HA imports, 
   - **Vanish semantics deliberately diverge from the shopping-list sync.** With `two_way`
     and `vanish_as_completed` on, a tracked open item that disappeared completes the
     task — required for providers (Todoist) whose `todo` entity drops completed items —
-    but only when the entry captured a `uid` (an add that never confirmably landed is
-    re-added, never completed). Otherwise a vanish means deleted → recreate (strict
-    self-healing). With `two_way` off the inbound direction is inert and a ticked item
-    freezes its entry so phase 2 doesn't re-add against the user's wishes.
+    but only when the entry captured a `uid`. Otherwise a vanish means deleted → recreate
+    (strict self-healing). With `two_way` off the inbound direction is inert and a ticked
+    item freezes its entry so phase 2 doesn't re-add against the user's wishes.
+  - **A write's own outcome outranks a later read of the list.** `todo.add_item` returns
+    nothing, so a fresh entry has no `uid` and is matched by summary next pass — but some
+    lists don't make an added item readable straight away (HA's CalDAV entity refreshes
+    its cache in a fire-and-forget task where `local_todo` and Todoist await theirs). So
+    "I can't see it" is **not** proof the add failed: reading it that way added the item
+    twice, permanently, since the bookkeeping only ever points at one copy. An
+    unconfirmed entry is therefore *held* — stamped with `added_at`, carried forward
+    **verbatim** because the summary is its only handle — until `UNCONFIRMED_GRACE`
+    (20 min, chosen to clear CalDAV's 15-minute poll) expires, then re-added so a
+    genuinely lost add still repairs. The same stamp gates the ticked-item arm, or a
+    summary match onto the predecessor we just ticked off would complete the task twice
+    and strand its replacement. Wall clock, not a pass count: the driver runs up to four
+    passes back to back with no delay between them.
   - **Content is capability-gated in the planner**, not the driver: due dates
     (`SET_DUE_DATE_ON_ITEM`, written date-only like our own `todo.py`) and
     descriptions (`SET_DESCRIPTION_ON_ITEM`, carrying the task's notes) are neither
--- a/.amazonq/rules/testing-and-workflow.md
+++ b/.amazonq/rules/testing-and-workflow.md
@@ -4,6 +4,12 @@
 - Never push directly to `main`. Work on a feature branch and open a PR; squash
   merge.
 - Update `CHANGELOG.md` for every user-facing change before a release.
+- **User-facing text is drafted by a Sonnet 4.5 subagent** (`model: sonnet`), not written
+  inline: `CHANGELOG.md` bullets, `README.md`, the canonical `docs/*.md`, `strings.json`,
+  `services.yaml` descriptions, the frontend locale. Hand it the diff, the surrounding
+  section for voice, and the house rules it must satisfy; review and edit the draft before
+  committing. Commit messages, PR bodies and code comments stay inline — they are not
+  user-facing.
 - **Keep every CHANGELOG bullet to three sentences at most.** A bold lead naming the
   change, then what a user notices, then a caveat or `(Fixes #N)` if one is needed.
   Cut the worked example, the before-and-after story, the list of every surface the
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -63,6 +63,15 @@
   try the feature via HACS *before* merge. The build is ephemeral and auto-deletes
   when the PR closes (see RELEASE.md → "Preview releases"). Bug-fix-only /
   developer-only PRs don't need it.
+- **Always have a Sonnet 4.5 subagent write user-facing text.** Any prose a *user* reads —
+  `CHANGELOG.md` bullets, `README.md`, the canonical `docs/*.md`, `strings.json`,
+  `services.yaml` descriptions, the frontend locale — is drafted by a subagent spawned
+  with `model: sonnet`, not written inline. Give it the diff, the surrounding section for
+  voice, and the house rules it has to satisfy (the three-sentence CHANGELOG budget, the
+  `(Fixes #N)` placement, the vale AI-tells style), then review what comes back and edit
+  it yourself before committing — the subagent drafts, you are still responsible for what
+  ships. Commit messages, PR bodies and code comments are *not* user-facing text and stay
+  inline.
 - **Always run tests locally before pushing.** Never use CI as the test runner.
   - Pure-logic unit tests need only `pip install pytest PyYAML`: `pytest tests/unit -v`.
     (`PyYAML` is for the API-surface gate below, which reads `services.yaml`; without
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.20.0b1]
+
+### Fixed
+
+- **Sending a task to a CalDAV to-do list (Nextcloud, Baikal, Radicale) no longer
+  creates a duplicate.** The duplicate was permanent: editing the task afterward
+  updated only one copy and deleting it left the other behind on the server. The same
+  timing could also complete a recurring task twice and skip its next occurrence.
+
 ## [0.19.0] - 2026-09-01
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -788,12 +788,11 @@ remotely completes nothing and the item returns on the next pass. That refusal i
 the point of requiring the scan.
 
 **A to-do item you add on the list yourself never becomes a Home Keeper task.** The
-sync is a delivery surface: it puts Home Keeper's tasks onto the list, and anything
-you write there stays exactly where you put it. Home Keeper leaves it alone — no
-import, no rename, no delete. A Home Keeper task carries recurrence, a device, labels
-and completion history, and a plain to-do line holds none of that. The one thing that
-travels inward is a checkmark: ticking off one of Home Keeper's own items completes
-the task behind it, and that is the whole inbound direction.
+sync is a delivery surface. It puts Home Keeper's tasks onto the list. Anything else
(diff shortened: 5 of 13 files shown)


prestomation:
/q review Be critical and skeptical, and do not withhold minor findings. This is a findings document, so the thing to attack is whether the **evidence actually supports the claims** — the first draft over-claimed a causal link between the duplicate bug and the reported symptom, and that has since been walked back, so please look hard for any remaining place where a measurement is presented as a diagnosis.

Order the most serious first, and scrutinise specifically:

1. **Correctness of the technical claims about Home Assistant internals.** The doc asserts that `caldav`'s `async_create_todo_item` fires its refresh without awaiting it while `local_todo` and `todoist` await theirs; that `todo.get_items` reads the entity cache without forcing a refresh; that `todo.update_item` merges over the existing item rather than replacing it; and that `_api_items_factory` serialises `due` with `isoformat()` so a `[:10]` prefix reads the right date for both `DATE` and `DATE-TIME`. Check each against the real source and flag anything version-dependent or wrong.

2. **The causal reasoning about Home Keeper's own code.** The claimed mechanism is: `todo.add_item` returns no uid, the entry is tracked with `uid=None`, the next pass cannot resolve it against a not-yet-refreshed list, the `item is None` branch in `plan_sync` drops the entry because there is no uid, and the second loop then re-adds. Is that the actual path, and is there a *different* path that would also produce a duplicate that the doc misses?

3. **Whether the proposed fixes are sound.** Option 2 is a grace period for uid-less entries in `todo_list.py`. Does that break the genuine "the add failed" case, given `_apply` already drops the key on a raising call? Does it interact badly with `vanish_as_completed`, or with two profiles syncing onto one list and the `claimed`/`settled` sets?

4. **Under-explored risk.** With `vanish_as_completed` on by default, a tracked entry that *does* have a uid and resolves to nothing produces a `CompleteOp` — completing a task nobody finished. The doc does not claim this is reachable on CalDAV because it was not reproduced. Is there a read path (a partial or failed `calendar.search`, the Nextcloud trashbin, a racing refresh returning an older snapshot) that could make it reachable? If so that is more serious than the duplicate and should be called out.

5. Anything in the "what is expected" / "CalDAV limitations" section that is overstated, or that contradicts the code.

---

prestomation:
Thanks — three of the five landed, pushed in e7aed5c. Triage:

**#5 (summary binding) — correct, fixed.** "Two tasks with the same name on one list are only ever distinguished by that binding" was false. `resolve_tracked` tries the uid first and only falls back to summary, so the ambiguity is confined to the pre-binding window, not permanent. Rewritten.

**#3 (Option 2 has a gap) — correct, fixed.** And the line was wrong twice over: it also said "Option 2 alone closes **the reported bug**", which contradicts this PR's own conclusion that the reporter's symptom was never reproduced. Replaced with the three questions that have to be settled before it is written — where a retry counter lives given `plan.tracked` is rebuilt each pass, the uid-binds-while-still-invisible transition, and not claiming an identity that was never resolved when two profiles share a list — and an explicit "neither option is validated here; both are directions."

**#1 (version-dependence) — the substance is right, the premise is not.** These were not asserted from memory: every one was read out of the running container. But you are right that the doc never said which version, and right that they are version-dependent — the `caldav` fire-and-forget refresh is an implementation detail upstream does not owe us. Now pinned to **HA 2026.8.3** with a note to re-check before relying on any of it, and to prefer a fix that does not depend on it holding.

**#2 (other races) — answerable, and the answer is no.** A refresh cannot land between `plan_sync`'s two loops: it is pure and synchronous over an `items_by_entity` snapshot captured before the call, with no `await` inside, so both loops see identical input. The window is strictly *between* passes. Added to the doc rather than left open.

**#4 (vanish risk) — reviewing the wrong text.** "More serious than the duplicate" is from my review request, not the document; the doc makes no such claim, which is why it is not overstated there. Your underlying question is a good one and is now answered from source: `WebDavTodoListEntity.async_update` assigns `_attr_todo_items` once, wholesale, after the executor job returns, so **no caller ever sees a partial snapshot**. An older complete snapshot overwriting a newer one is possible with two refreshes in flight, but that is the same missing-recent-item shape as Finding 1, not a separate defect. Agreed the vanish path stays theoretical and is not claimed otherwise.

---

prestomation:
/q review This push turns the findings doc into an actual fix in `todo_list.py`. Be critical and skeptical, most serious first, and do not withhold minor findings. The change is a concurrency/consistency fix, so please attack the reasoning rather than the style.

**Correctness — the core claim.** The fix asserts that a successful `todo.add_item` is proof the item landed, and that a later read of a stale cache must not overturn it. An unconfirmed (uid-less) entry is now held rather than re-added, bounded by a 20-minute wall-clock grace. Attack this:

1. Is there a state where an entry is held **forever**? I believe the two escapes are covered — `_added_stamp` rewrites a stamp that is unparsable or in the future, and the coordinator's 5-minute sweep is unconditional so it always drives the eventual repair — but the fix now *depends* on that sweep staying unconditional. Is there a path (no listeners on the coordinator, a profile deleted mid-hold, a target list permanently unavailable) that strands one?
2. Does holding the key in `plan.tracked` interact badly with the `claimed` and `settled` sets? A held entry resolved no item so it claims nothing. Can a different key's `find_open` adoption then take the very item the held entry is waiting to match, and does that matter given loop one runs to completion before loop two?
3. `UNCONFIRMED_GRACE` is 20 minutes against CalDAV's 15-minute `SCAN_INTERVAL`. Is ~33% headroom enough over a constant upstream owns, on a poll that itself takes a WAN round trip? Argue for a different number if you think it is wrong.
4. **Bug B specifically** — the ticked-item arm is now gated on `entry.get("added_at")` being *present* rather than on the grace rule, so an entry with no stamp still completes as before. Is that gate correct, or does it leave a reachable phantom completion?

**Regression risk.** `_entry` gained a sixth key with no storage migration (the map is opaque in `store.py` and documented as additive). Does anything break on a document written before this key existed, or on a provider that returns no uids at all — where `_entry` at the rebind site writes `uid=None` and drops the stamp on every successful pass?

**Also worth your scrutiny:** whether the two new helpers should be one; whether the module docstring and `.amazonq/rules/architecture-and-code.md` now describe the real behaviour; and anything in the README addition or CHANGELOG bullet that overstates what shipped.

Context: mutation gate 98.77% (401/406) scoped to the four changed functions, 1591 unit tests pass, and both bugs were verified fixed against a real Nextcloud (0 duplicates across 4 rounds, previously 1–3 every round).

---

prestomation:
Thorough review — thanks. Two points taken, pushed in e06141b; one I think is not reachable; two corrections.

**Taken: the sweep dependency (your headline recommendation).** Right, and it was undocumented. Written at both ends now — `async_schedule_sweep`'s docstring says why it must stay unconditional, and `UNCONFIRMED_GRACE` points back at the sweep as the thing that repairs it. The specific trap worth naming is that `needs_pass` answers **False** for a held entry, and correctly so — a pending write is not drift — so gating the sweep on it later would look like a harmless optimisation and would silently make every hold permanent.

**Taken: minor 2 and 3.** `_added_stamp` now says why it coerces with `str()`: the store hands back whatever the document holds, so a number or some other write's leftover has to read as "cannot be trusted" rather than raise. The constant now cross-references the sweep interval.

**Not reachable: the `find_open` race in Q2.** You suggest pass 2 could adopt the item a held entry is waiting for. It cannot, and the reason is the one you already identified — loop 1 runs to completion before loop 2, over **one snapshot** captured before `plan_sync` is called:

- If the item is visible in that snapshot, loop 1 resolves the held entry against it first (uid-less entries fall back to summary in `resolve_tracked`), so the entry binds and is never held.
- If it is not visible, `find_open` in loop 2 cannot see it either.

There is no state where loop 2 sees an item loop 1 did not. The case that *does* exist is a second entry claiming the only visible line, leaving the first to hold — and adding a fresh line for the other task is the correct outcome there, already pinned by `test_two_tasks_reading_the_same_never_share_a_line`. So no doc change; documenting a race that cannot occur would mislead the next reader.

**Correction: minor 4.** `grace` is already annotated — `grace: timedelta = UNCONFIRMED_GRACE`.

**Correction: your Q4 reasoning reaches the right verdict by the wrong route.** You say the post-grace phantom arm "is not reachable" because grace expiry is handled earlier by `continue`. That earlier `continue` is in the `item is None` branch, which only runs when nothing resolved. A ticked-off predecessor *does* resolve, so control reaches the ticked-item arm with the grace expired and the task completes. That is intended — after 20 minutes the cache is fresh and an open item would have won over a ticked one — but it is a deliberate choice, not an unreachable path, and it is covered by `test_a_tick_still_lands_once_the_unconfirmed_add_has_waited_its_grace`.

**On 25 minutes (Q3):** your WAN/jitter argument is reasonable, but 20 was chosen deliberately with the 15-minute floor already in view, and the practical path is much shorter than the poll — Home Keeper's own write forces a refresh within ~1s, so the 15-minute poll is the fallback, not the norm. Leaving it at 20; happy to revisit if anyone reports a hold expiring early in the wild.

**Diffs you asked for:** README and CHANGELOG are in this PR (`08ffd12`, `22b7cf5`). Both were rewritten after the vale job caught three AI-tells in the README paragraph, including a `VerbTricolon` that a local vale 3.9.1 run reported clean — the token spans sentence boundaries via `[^,]+`, so one pronoun swept up every comma in the paragraph. Verified afterwards both with the vale CLI and by running every `styles/ai-tells/*.yml` regex against the added text directly: 0 hits.

---

PR #282, opened 2026-09-01
feat(panel): duplicate a task from its detail page

### LINKED ISSUE #279: Support / Feature request - Task duplication
Hi,

I'm looking for a solution to easily duplicate tasks.
For example I created a task to water flowers depending on humidity level of the ground based on the nearby sensor. I have ten sensors like that so is there a way for me to easily duplicate the task configuration and only adjust the sensors and values ?

prestomation:
Fixes #279

## What changed

A **Duplicate** button on a task's detail page, beside Edit. It opens the *create* form already filled in with a copy of that task; nothing is saved until **Create** is pressed.

The reporter has ten "water the flowers" tasks, one per soil-moisture sensor, differing only by a name, a sensor and a device. Today the eleventh means retyping the whole form.

A task's page with Duplicate beside Edit, and the create drawer open prefilled with a copy — titled New task, subtitled Replace fridge filter (copy), with the notes and cadence carried across and a Create button

**No new backend surface.** `_submitForm` already routes an id-less task to `add_task`, so duplication is a *seeding* problem, not a persistence one. `duplicateTaskSeed` (in `forms.ts`, beside `consumableLinkToken`) builds the copy as an **allowlist** — the same discipline `buildTaskPayload` follows — so a field added to `Task` later cannot leak into a copy just by existing.

### Three omissions are load-bearing

The copy has no `id`, and the marshalling treats an id-less task as a creation. That makes these behavioural, not tidiness:

| Field | Left in, what happens |
|---|---|
| `last_completed` | `buildTaskPayload` emits it **only** when `!task.id` — never true on the edit path, always true here. Every copy is born back-dated to the original's last completion, and `next_due` is derived from it. |
| `sensor.baseline` | The meter anchor. The copy counts from the original machine's reading instead of stamping its own. The fixture nozzle task is anchored at 660 h against a 300 h target, so an inherited baseline starts a copy most of the way through an interval it never measured. |
| `tag_id` / `require_tag_scan` | One sticker completes one task. A second task claiming the same tag is a collision, not a copied setting. |

`source` is omitted too, which is why `consumable_link` rides along as the flat token: `taskFormData` renders the picker from `source`, but `_submitForm` reads only the flat key, so seeding `source` would display a link the save never applies.

### Refusal, and why triggered tasks are in it

A task Home Keeper doesn't own keeps a **greyed** Duplicate that explains itself when pressed, rather than silently missing one.

A Pawsistant-managed task's page showing a live tonal Edit beside a greyed-out Duplicate pill, with Delete replaced by delete-from-Pawsistant guidance

Source-owned (wear parts, synced problem sensors) and managed tasks are refused on ownership. **Triggered tasks are refused on the recurrence type instead**, which is a correctness fix rather than a policy: `buildTaskPayload`'s triggered branch emits only `name`/`notes`/`device_id`, so a copy would reach `build_task` with no schedule at all and `infer_recurrence_type` would read it as a **one-off due today**. Guarding on the type keeps that true for a triggered task an integration pushed without declaring ownership.

### Button weight

Edit and Duplicate are drawn at the same (secondary) weight, because they are peer actions on the same task. At tertiary the live button was bare text and the blocked one bare *grey* text, so the only thing separating "you can press this" from "you can't" was the shade of the label. The row still draws at the three weights #262 exists to protect: Done fills, Edit and Duplicate are tonal, Delete recedes to red text — `button-weights.spec.ts` passes unchanged.

### One more thing worth a reviewer's eye

`_blockedDone`'s span-wraps-a-disabled-button trick is generalized to `_blockedButton` under a neutral `hk-blocked-wrap` class. It deliberately does **not** reuse `done-blocked-wrap`: `panel-lists.ts:420` wires every element carrying that class to the *Done* reason, which would misfire the moment a blocked action of another kind appeared somewhere `wireLists` runs. The greyed Duplicate is a `role="button"` span, so Enter and Space are wired by hand — a disabled `ha-button` swallows events.

### Open question

**`consumable_link` is carried over.** Only a `manual` link can reach this path (reconciler-owned ones are refused), so the affected case is "a task the user linked to a consumable by hand". The copy becomes a proper manual link, but **both tasks then draw down the same part and both fire the low-stock reorder**. For two identical filters sharing a spares bin that's right; for an unrelated copy it's a surprise. Dropping it is one line and one flipped assertion — say the word.

## Verification

- `npx vitest run` — 843 pass, including 13 new `duplicateTaskSeed` cases and 12 new DOM cases (`task-duplicate.test.js`)
- `pytest tests/unit` — 1592 pass
- `bash ci/test-mutation-frontend.sh` — **100%** on the changed surface (16/16 mutants killed), threshold 80%
- `bash ci/e2e-up.sh` — 208 pass, including `button-weights.spec.ts` and the new `task-duplicate.spec.ts`; the new spec re-run with `CHROMIUM_EXEC` unset on CI's headless shell
- `vale` — clean; I also ran all 24 `styles/ai-tells/*.yml` regexes directly against every added line, which is what caught the second tricolon the first CI run had not yet reported
- `python3 ci/check-changelog-release-gap.py`, `ruff check`, `ruff format --check`, `tsc --noEmit`, `ci/build-panel.sh` — clean

Two notes on what I did **not** commit:

- **Only the task-detail screenshots are refreshed.** Re-running the capture rewrites all 74; a pixel diff shows the untouched surfaces (appliance lists, settings, mobile) differ only by font-rendering drift in this environment, so committing them would bury the change under noise.
- **The planned blocked-Duplicate toast shot is dropped rather than committed broken** — the capture caught HA's loading screen. Shot 9 above already shows the greyed button, and the e2e spec, not the picture, holds the behaviour.

The screenshot capture's timeout goes to 240s: adding the Duplicate drawer to the walk pushed a cold container past the 180s budget partway through, starving the later shots — the same reason the previous bump exists.

## Checklist

      embedded above with an HTML `img` tag
      step through it


prestomation:
/q review Be critical and skeptical. Put the most serious issues first, and don't withhold the minor ones.

Focus on:

**Correctness of the field strip.** `duplicateTaskSeed` in `forms.ts` builds a copy as an allowlist. Two omissions are subtle and I want them attacked: `buildTaskPayload` emits `last_completed` *only* when `!task.id`, which is never true on the edit path and always true on a duplicate; and `sensor.baseline` is read from `task.sensor?.baseline`. Can either still reach a created copy by any path? Is the shallow copy of the `sensor` binding actually enough to stop the source task in `_tasks` being mutated, given nested objects like `also_every`?

**Completeness of the refusal gate.** `_canDuplicate` blocks source-owned, `managed_by`, and `recurrence_type === 'triggered'` tasks. Is there any task shape whose payload would be rebuilt *wrongly* by a duplicate that this gate still lets through — an orphaned managed task, a task with `completion_required_fields`, a one-off already completed, a task whose `managed_by.locked_fields` covers a field the copy carries?

**The consumable link.** The copy carries `consumable_link` deliberately, so two tasks can draw down the same part and both fire the low-stock reorder. Is that defensible, or should it be dropped?

**Maintainability.** Does the seed belong in `forms.ts` (it is inside the Stryker mutate allowlist there) rather than `panel.ts`? Is `_blockedButton` the right generalization of `_blockedDone`, and is widening `PanelHost` with `_blockedDuplicate` / `_canDuplicate` / `_notifyNoDuplicate` / `_openDuplicate` the narrowest seam? Is lifting `sourceOwnedTask` into `panel-chips.ts` the right home?

**Accessibility and HA practice.** The greyed Duplicate is a `role="button"` span wrapping a disabled `ha-button`, with Enter/Space wired by hand because the disabled button swallows events. Is that sound across HA button versions, and is the `hk-blocked-wrap` / `done-blocked-wrap` class split correct given `panel-lists.ts` wires the latter to the Done reason specifically?

**Security.** A copy must never carry `managed_by` or `source` into an unowned task, and `store.add_task` rejects the reserved `part` / `problem_sensor` / `buy` source namespaces. Is that boundary actually airtight from the panel?

---

PR #283, opened 2026-09-01
feat: close the remaining #220 asks — the amount on the shopping line, and buy reminders out of Overdue

### LINKED ISSUE #220: A few feature requests and questions

prestomation:
Fixes #220

## What changed

The 2 asks that were still open from crag364&#39;s follow-up on #220, and 2 problems found while the work was done.

### 1. The shopping-list line shows the amount to buy

A mirrored buy task read `Buy fabric softener` and did not show the amount, although the part already has a restock quantity and a unit. The line now shows it:

| part | line |
| --- | --- |
| `stock_unit: ml`, restock 500 | `Buy fabric softener (500 ml)` |
| no unit, restock 2 | `Buy air filter (×2)` |
| no unit, restock 1 | `Buy anode rod`, not changed |

The amount is added to the **mirrored summary only**. The task keeps its own name in the panel and the calendar and the notifications and the Home Keeper to-do entity. The suffix is the same in every language, so there is no new translated template.

2 points for a reviewer:

- The amount uses the existing `name` key of `buy_tasks_by_part` and not a second key. `plan_sync` and `needs_pass` both compare that value against the last mirrored value. If the key meant the task name in one function and the line text in the other, the mirror would never settle. It would read the list and rename the same item on every pass.
- **No migration is needed.** `resolve_tracked` matches on the summary that was stored, and not on the summary that is wanted now. A line that was mirrored before this change is found by its old text and is renamed in place. There are tests for an entry with a uid and for an entry without one, because a new line has no uid to match on.

### 2. A buy task is no longer shown as overdue work

A buy task is created as a one-off task with no due date, and a task with no due date is due immediately. It was therefore overdue from the moment a part went low until the part was restocked. It was shown in the Overdue section with the overdue maintenance tasks.

It now has a **Shopping** section of its own, in the panel and on the dashboard card. Its status reads **Low stock**.

The panel task list grouped by status: an Overdue section of ten rows, then a Shopping section holding Buy Anode rod with a Low stock status, then Later

**No count changes.** The task is still overdue for the scope pills and the per-task binary sensors and every saved Profile. If the panel showed 1 overdue task while the sensors showed 2, that would be a second problem. In the image above the Overdue pill reads 11 and the Overdue section holds 10. The eleventh task is the buy task. To make the pill match the section, change `scopeMatches`. The cost is that the panel then disagrees with the sensors.

The section and the status use `filter.shopping` and `chip.lowStock`. Both are already translated in all 16 languages, so there are no new strings.

### 3. The Shopping filter is kept after a reload

`_setFilter` always saved the value, but the restore accepted only `all` and `overdue` and `soon`. The pill therefore changed back to All on every reload. The fix is 1 line. There is a test that fails without it.

The Shopping filter selected, listing the single Buy Anode rod task under a Shopping heading

### 4. A Profile can exclude the shopping tasks

A saved Profile could remove tasks by label or area or device. A buy task has none of its own, because it takes the label and the area of its appliance. There was therefore no way to select every overdue task except the shopping. The reporter kept buy tasks out of a spoken digest with a script that filtered on `source.buy`.

The new **Exclude shopping** switch does this instead. It removes tasks with the exclude lists, but by kind and not by id. It is off by default. `normalize_filter` rebuilds the block from a fixed set of keys on every read, so an existing Profile is not changed and no migration is needed.

A profile editor showing the new Exclude shopping switch, off, below Exclude companions and above the to-do list sync section

The conformance cases for the switch found a **real difference between the 2 languages**. The Python function `buy_source` requires both `asset_id` and `part_id`. The frontend read `source.buy` as a plain truthy check. A source with only one of the 2 ids was therefore shopping on one side and not on the other. The frontend now has 1 `isBuyTask` function that matches the backend. The status bucket and the Shopping filter and the panel scope and the profile matcher all use it.

### Also in this pull request

`docs/images/44-panel-shopping-filter.png` showed nothing. The capture ran before the step that puts a buy task in the store, so the image showed only &#34;No tasks match this filter&#34;. The README has shown that image since the filter was released. The capture now runs after that fixture, which is also what makes the new Shopping section visible. AGENTS.md warns about this failure with #221: a captured surface that no test checks.

The status colour was wrong in the first capture, and it was found only by looking at the PNG. The rule `.hk-status ha-assist-chip:not(.hk-overdue)` sets the page colour on every status chip, which replaced the warn fill. The rule now excludes `.hk-shopping` in the same way that it already excludes `.hk-overdue`.

The 2 formatters that render a quantity did not agree. Python `round` rounds a tie to the even digit and JavaScript `toFixed` rounds a tie away from zero, so 0.0625 read &#34;0.062&#34; on the shopping list and &#34;0.063&#34; in the panel. `assets.format_quantity` now quantizes the exact binary value half-up. This was checked against Node over 8000 values, 2000 of them exact halfway values, with no differences. `tests/fixtures/quantity_format_cases.json` holds the 2 implementations to the same output, in the same way that `profile_filter_cases.json` does for the Profile filter.

## Checklist

- [x] `CHANGELOG.md` updated, user-facing changes only, with `(Fixes #N)` for each
      issue that is fixed
- [x] Panel UI changed, so current screenshots are committed under `docs/images/` and
      shown above with an HTML image tag
- [x] New user-facing UI surface, so `tests/e2e/walkthrough.capture.ts` steps through it
- [x] New user-facing feature, so the version is at the next beta (`manifest.json` and
      `const.py`), and the `preview-release` label is applied

## Verification

Run on the rebased branch, at daad708.

- `pytest tests/unit` — 1662 passed, 2 skipped
- `bash ci/test-frontend.sh` — 950 passed
- `bash ci/e2e-up.sh` — 217 passed
- `python3 ci/check-changelog-release-gap.py` — clean
- `ruff check` and `ruff format --check` — clean
- `tsc --noEmit` — clean
- The screenshot capture — the 5 shots above are re-captured on the rebased tree

Not run again after the rebase: the 2 mutation jobs and vale. CI runs both.

## Rebase note

This branch is now rebased onto `main` at 927e0ba, which adds 3 pull requests that landed while it was open: #288 (credit outside contributors), #290 (a profile filters by companion) and #289 (the Test button reports its delivery, and a Settings row keeps its edit).

**The beta is 0.20.0b6.** #290 shipped `0.20.0b4` and #289 shipped `0.20.0b5`, so this work opens its own section. `manifest.json`, `const.py` and the top `CHANGELOG.md` heading all read `0.20.0b6`, and `check-changelog-release-gap.py` passes.

**#290 is the one that really touched this branch.** It adds `companions` and `exclude_companions` to the same profile filter that this branch adds `exclude_shopping` to, so both sides changed the same 6 places. Each is resolved by keeping both, with the shopping check placed after the device check and before the companion check on both sides:

- `profiles.py` — `normalize_filter` carries both new keys, and `matches_filter` runs both new exclusions.
- `card-filter.ts` — the same order, so the 2 languages still agree.
- `forms.ts`, `types.ts`, `panel-settings.ts` — the field, the type and the default for a new profile.
- The 16 locale files — `notify.exclude_shopping` from this branch beside `notify.exclude_companions` and `notify.filters_help` from #290.
- `tests/fixtures/profile_filter_cases.json` — 46 cases from #290 plus the 6 from this branch, 52 in all, no name collision.

**The fixture is back in the file's own format.** This branch had reflowed all 198 lines of `profile_filter_cases.json` to fully-expanded JSON, which is why the diff read +685. The merge writes the 6 new cases in the one-line-per-field style the other 46 use, so the file is 318 lines rather than 755 and the diff for it is +38. That is most of why this pull request is now +1129 rather than +1713.

**5 screenshots are re-captured, not 3.** #290 puts the Companions and Exclude companions pickers in the profile editor, so **Exclude shopping** now sits below those rather than below Exclude devices. `47-panel-profile-sync.png` and `profiles-card.png` were byte-identical through the previous rebase and are not any more. Every shot also read `v0.20.0b3` in the panel footer. Only the shots this branch owns are committed; a full capture run rewrites every PNG in `docs/images/`, because the seeded dates move each day.

`panel-styles.ts` and the 2 capture harnesses merged without a conflict this time, and were re-read to confirm both sides survived.

Not done here: the code comments and docstrings that this pull request adds are still in the old voice. The writing-style rules include them, but no lint checks them, and a rewrite would hide the changes to the code. Say if you want that in this pull request.

https://claude.ai/code/session_014kKLwSDBBTXV4WEHxCYNC2


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones. This PR touches the shopping-list mirror, the panel/card status bucketing, and the Profile filter contract that Python and TypeScript both implement. Areas I most want scrutinised:

**Correctness of the mirror (`shopping.py`, `shopping_sync.py`, `assets.py`)**
- The restock amount now rides on `buy_tasks_by_part`'s existing `name` key, which `plan_sync` compares against `tracked["summary"]` and `needs_pass` uses as its drift gate. Is there any path where those two can now disagree and leave the mirror permanently dirty — re-reading the list and re-renaming the same item on every pass?
- I claim no migration is needed because `resolve_tracked` matches on the *stored* summary, not the desired one, so an upgrade renames a line in place. Is that true for every branch — uid-less entries, a list that hands out no uids, a target that changed between passes, two parts whose lines read identically?
- `_restock_label` is deliberately total (a missing asset or part yields `""`). Does that silently mask a real bug anywhere, or produce a line that flip-flops between two summaries across passes as the reconciler retires the task?
- `format_quantity` is a hand-written twin of the frontend's `formatQuantity`. Any float case where they disagree?

**The Python↔TypeScript contract (`profiles.py`, `card-filter.ts`, `tests/fixtures/profile_filter_cases.json`)**
- `exclude_shopping` is new on both sides. Is the absent-key default genuinely off everywhere it is read, including a profile loaded from storage that predates it?
- I unified the frontend on one `isBuyTask` predicate requiring both ids, to match `reconcile.buy_source`. Did I miss a `source.buy` reader, and does tightening the *Shopping filter* to require both ids change behaviour for any task that could exist in the wild?

**HA best practices**
- `profiles.py` now imports `buy_source` from `.reconcile`. Is that a sensible dependency direction for the pure core, or does it couple two modules that should stay independent?
- Any concern about `shopping_sync` calling `store.get_assets()` on every pass?

**Panel/card correctness**
- `statusBucket` gains a `shopping` branch placed after the `completed` check and before the `!next_due` guard. Is that ordering right for every task shape — a bought one-off, a dateless buy reminder, a disabled one?
- Buy reminders stay in the Overdue *count* but move out of the Overdue *section*, so the pill reads 11 while the section holds 10. Is that defensible, or should the pill follow the section?

**Performance / maintainability**: anything in the above that adds redundant work per sync pass or per render, and anything where I have duplicated logic that should be one function.

---

prestomation:
Thanks — the `format_quantity` parity concern was the right thing to press on, and it was a real divergence rather than a theoretical one.

**Confirmed and fixed in ded66c4.** The two formatters disagreed on exactly-halfway values: Python's `round()` breaks a tie to the even digit, JavaScript's `toFixed` breaks it away from zero. So `0.0625` rendered `0.062` on the shopping list and `0.063` in the panel it came from.

`format_quantity` now quantizes the exact binary value half-up, which reproduces `toFixed`. I checked that empirically against Node over 8000 values — 2000 of them deliberate halfway cases at the 4th decimal — with zero mismatches. `_round_stock` is deliberately untouched: it rounds stock on the way *into storage*, so changing it would move stored numbers rather than just displayed ones.

The two are now pinned by `tests/fixtures/quantity_format_cases.json`, following the `profile_filter_cases.json` precedent you'd expect here, run by both `tests/unit/test_assets.py` and `frontend/test/utils.test.js`. I verified the fixture actually bites: reverting the Python fix fails it on the `0.0625` case specifically.

Worth recording that in practice no household would have hit this — `_normalize_stock` already rounds to the same three places on the way in, so both sides were seeing a settled number. Fixed anyway, because the function's docstring claims the two render the same numbers and that should be true without a caveat.

On your second point, `store.get_assets()` per pass: agreed it's acceptable as-is. It returns the in-memory dict by reference with no copy, and the `needs_pass` gate short-circuits before it on the overwhelming majority of pokes. Left alone rather than adding a cache that would need invalidating.

Everything else in the review matched my own reading, so no other changes.

---

prestomation:
/q review Please re-review after ded66c4, which addresses the one gap you asked to close before merge. Be critical and skeptical again, and look hardest at the new work rather than re-confirming what you already cleared:

**The rounding change (`assets.format_quantity`)**
- It now quantizes `Decimal(float(value))` half-up at three places instead of using `_round_stock`. Is `Decimal(float(v)).quantize(..., ROUND_HALF_UP)` genuinely equivalent to JavaScript's `Number.prototype.toFixed(3)` for every input that can reach it, or did my 8000-value differential sample miss a class — very large magnitudes, subnormals, values where `toFixed` switches to exponential notation, negative inputs?
- I deliberately did **not** change `_round_stock`, so storage rounding and display rounding now use different tie-breaking. Is that split defensible, or does it create a case where a stored value and its rendering disagree in a way that matters?
- `int(quantized)` truncates rather than rounds. Given the value is already quantized to three places, is there any input where that is wrong?

**The conformance fixture (`tests/fixtures/quantity_format_cases.json`)**
- Does it cover the classes that actually matter, or is it just the cases I happened to think of? Anything you would add that would have caught a real bug.
- Both sides read it. Is there a way for one side to silently pass vacuously — e.g. a parse failure, or an empty `cases` array — that my "has cases to run" guard does not close?

**Anything else the two new tests get wrong**, and whether the Python and TypeScript assertions are actually testing the same thing rather than merely both being green.

---

prestomation:
Thanks — your review landed against `ded66c4`, and both things you asked for are already in `209b0ae`, pushed a few minutes later after I audited the fix against the same questions:

- **Negative cases** — added to the fixture (`-2.5`, `-3.0`, `-0.0625`, plus the largest stock a part can hold), each checked against Node first. `-0.0625` → `-0.063` on both sides, so `ROUND_HALF_UP` is symmetric as you expected.
- **Documented precision limits** — the docstring and the fixture comment now state the range the guarantee covers rather than implying it holds everywhere.

**One correction, because a future reader shouldn't rely on it.** You wrote:

> JavaScript's `toFixed()` never switches to exponential—it always returns decimal notation regardless of magnitude. Your implementation matches this.

`toFixed` does switch. ECMA-262 §Number.prototype.toFixed short-circuits to `ToString(x)` when `x ≥ 10²¹`:

```
1e+20  toFixed(3) => "100000000000000000000.000"   formatQuantity => "100000000000000000000"
1e+21  toFixed(3) => "1e+21"                       formatQuantity => "1e+21"
1e+22  toFixed(3) => "1e+22"                       formatQuantity => "1e+22"
```

Python has no such cutover, so the two genuinely part company at 1e21 — which is why `209b0ae` bounds the claim instead of asserting equivalence at all magnitudes. Nothing that large can reach either formatter (`MAX_INTERVAL` caps stock at 10,000 on the way in, and `_positive_quantity` floors the per-use amounts), so it stays a documented edge rather than a bug.

Your audit did surface a real one, though, just not the one you named: quantizing **raised** past about 1e25, because the default decimal context carries 28 digits and a double runs to 309 integer places. `_round_stock` had returned a number there. `format_quantity` is now total the way `_positive_quantity` is — a non-finite or oversized value falls back rather than raising, since a formatter that throws would take down a whole shopping-list sync pass over one malformed part. Pinned in `tests/unit/test_assets.py`, deliberately not in the shared fixture: Python and JavaScript spell infinity differently, so it isn't a conformance case.

On the storage/display rounding split — agreed, and your framing of why is the one I'd use. Nothing further outstanding from my side.

---

prestomation:
/q review This branch is rebased onto `main` at aa08ba0, which added #277. Be critical and sceptical, list the most serious problems first, and do not hold back the small ones.

Look hardest at the rebase itself, because 12 files were touched by both sides:

- `panel-styles.ts`. #277 replaced `.hk-notify-delete` with `.hk-item-actions`. This branch adds `.hk-shopping` chip rules and adds `:not(.hk-shopping)` to the `.hk-status ha-assist-chip` outline rule. Check that neither side lost a rule and that the selector specificity still does what each comment claims.
- `walkthrough.capture.ts` and `screenshots.capture.ts`. Both sides added steps. Check that the shopping steps and the notification steps do not fight over panel state, and that no step reads an element the other side renamed.
- `forms.ts`, `panel-settings.ts`, `types.ts`, the 16 locale files. Both sides added keys to the same objects. Check for a key that is added twice or lost.
- `CHANGELOG.md`, `manifest.json`, `const.py`. The version moved to `0.20.0b4`, because #277 shipped `0.20.0b3` as a tag. Check that the 3 agree and that the b3 section is not changed.

Then review the change on its own terms:

- **Correctness.** `assets.format_quantity` quantizes a binary float half-up to match JavaScript `toFixed`. Look for an input where the 2 still differ: a negative amount, a very large one, a subnormal, or a value that needs more digits than the quantizer keeps. Also check `isBuyTask` against `buy_source` for a source shape that one accepts and the other rejects.
- **The shopping bucket.** A buy task is due immediately and is still counted as overdue by the scope pills and the binary sensors, but is drawn in its own section. Look for a place that reads the bucket where it should read the scope, or the reverse.
- **The mirrored line.** The restock amount goes on the summary that `plan_sync` and `needs_pass` compare. Look for a path where the mirror does not settle and renames the same item on every pass, and for a rename that loses the tracked uid.
- **Profiles.** `exclude_shopping` excludes by kind rather than by id. Check that `normalize_filter` fills it in for a Profile saved before it existed, on every read path.
- **Maintainability, performance, security, and Home Assistant practice.** Name anything that duplicates logic across the 2 languages without a conformance fixture to hold it.

---

prestomation:
/q review Rebased again, onto `main` at 927e0ba, which picked up #288, #290 and #289 since the last review. Be critical and sceptical, most serious first, and do not hold back the small ones.

The rebase is where the risk is. #290 adds `companions` / `exclude_companions` to the same profile filter this branch adds `exclude_shopping` to, so both sides edited the same 6 places and I resolved each by keeping both:

- `profiles.py` — `normalize_filter` and `matches_filter`. Check the exclusion order I chose (labels → areas → devices → shopping → companions) is correct and that no early `return` swallows a later check.
- `card-filter.ts` — the same order in TypeScript. Check the 2 languages still agree case for case, not just in the fixture.
- `forms.ts` / `types.ts` / `panel-settings.ts` — the field, the type, the default for a new profile. Check nothing was added twice or dropped, and that `exclude_shopping` is non-optional in `NotifyFilter` while `profileFormData` still reads it with `?? false`.
- The 16 locale files — I merged the conflicting runs by taking both sides' lines. Check no key is duplicated or lost and no translation is attached to the wrong key.
- `tests/fixtures/profile_filter_cases.json` — rebuilt as #290's 46 cases plus this branch's 6. I also reverted this branch's reflow of the whole file back to the one-line-per-field house style. Check the 6 cases survived intact and that the `_comment` still describes every filter.

Then the change on its own terms, as before:

- **Correctness.** `assets.format_quantity` quantizes a binary float half-up to match JavaScript `toFixed`. Find an input where the 2 still differ — negative, very large, subnormal, or needing more digits than the quantizer keeps. Check `isBuyTask` against `buy_source` for a source shape one accepts and the other rejects.
- **The shopping bucket.** A buy task is due immediately and is still counted as overdue by the scope pills and the binary sensors, but is drawn in its own section. Look for a place reading the bucket where it should read the scope, or the reverse — and whether #290's companion filter interacts with that anywhere.
- **The mirrored line.** The restock amount goes on the summary `plan_sync` and `needs_pass` compare. Look for a path where the mirror does not settle, and for a rename that loses the tracked uid.
- **Profiles.** `exclude_shopping` excludes by kind. Check `normalize_filter` fills it in for a profile saved before it existed, on every read path, now that #290 added its own keys to the same function.
- **Maintainability, performance, security, and Home Assistant practice.** Name anything duplicated across the 2 languages without a conformance fixture holding it.

---

PR #284, opened 2026-09-01
Document a deadline and follow-up automation example, and test it

### LINKED ISSUE #280: Conditonnal task cancelation and creation
Here is my use case and I’m not sure what is best to be done with only Home Keeper, only HA automation or a bit of both: I’d like to get some notification about getting the trash bin out and back in when local collection truck is going to arrive.

Perfect situation from me would be:

Task for getting Trash out every Wednesday at 7pm (this one is easy and I got that created already)
If task was completed before 7am on Thursday then create a new Task to get the trash bin back in at 8pm on Thursday.
If the task was not completed before 9am on Thursday then cancel this occurrence and do not create the task to the trash bin in as it is too late to get trash collected.
So I believe that there are two things in my use case:

Conditionally create a new different task on Home Keeper if a task is completed before a specific time. Not sure if this should be automation or if it is something Home Keeper might be able to do.
Conditionally cancel an occurrence of a task on Home Keeper if it was not completed before a specific time.

prestomation:
Fixes #280

## What changed

#280 asks for 2 things. The first is to skip this week's occurrence when a task misses its deadline. The second is to make a follow-up task due only when the task is complete in time. The Home Keeper services already support both, so this change documents an example instead of adding to the task model.

`README.md` gains a "Deadlines and follow-ups" section under Events and automations. It uses the trash bin case from the issue.

- The first automation reads the task from `home_keeper.list_tasks` with `response_variable`. If the task is still due, the automation calls `home_keeper.skip_task`. A fixed task moves forward exactly 1 occurrence, and Home Keeper records no completion.
- The second automation arms the triggered task "Bring the bin back in" with `home_keeper.trigger_task`. It arms the task only if `last_completed` is inside a grace window. The window starts at `next_due - 7 days`. A completion or a skip already moved `next_due` to the next week, so that value is the occurrence that ended. The automation therefore reads no clock, and a week that you skipped cannot arm the follow-up task.
- A third automation makes a dated one-off task in place of the triggered task.
- The section states what to change. The value `timedelta(days=7)` is correct for a weekly task only. The automation stops with an error if no task has the configured name.

The change also records a result found in testing. Home Keeper creates a triggered task armed, with `next_due` set to the creation time. The task needs 1 completion to become dormant. The comment at `recurrence.py:249` said that `build_task` starts the task dormant. `models.build_task` applies that only to a sensor task. This change corrects the comment.

## The automations are tested

The automations are documentation, so this branch does not keep a copy of them in a test file.

- `tests/integration/ha_config/recipes.yaml` holds the README automations verbatim, and the container loads that file.
- `tests/integration/test_automation_recipes.py` drives them against real tasks. A missed deadline moves the task forward 1 week and records no completion. A completed task stays unchanged. The grace window arms the follow-up task inside the window and refuses a completion on each side. A completion before the occurrence is refused, and a task that nobody completed is refused. A skipped week never arms the follow-up task.
- `recipes_probe.yaml` takes its trigger time from an `input_datetime`. The test sets that value a few seconds in the future. The suite therefore shows that a real `trigger: time` event drives the sequence.
- `test_recipes_match_readme` reads the yaml blocks from the README section. It fails if the fixture and the document become different.

### The boundary tests were checked against a changed fixture

The first tests accepted an incorrect grace window. These 2 tests now fail when the automation is wrong:

| change applied to `recipes.yaml` | test that fails |
| --- | --- |
| `grace_hours: 13` | `test_grace_window_boundary_is_twelve_hours` |
| the `&gt;= as_datetime(occurrence)` condition removed | `test_completion_before_the_occurrence_does_not_arm_the_follow_up` |

A `timedelta(days=8)` change already fails the in-window test.

## Writing style

This branch first wrote its own ASD-STE100 rule into `AGENTS.md` and `CLAUDE.md`. `main` adopted ASD-STE100 in the same period, and its version covers more: code comments, docstrings, PR text and chat replies, plus a glossary and a committed `HomeKeeper` vale style. The merge takes main's rules, and this branch keeps only the note that a Sonnet subagent drafts user-facing text.

The prose is rewritten against main's rules. Numbers are numerals. A condition comes before its instruction. The 2-task setup is a list. The glossary gives "skip" and not "cancel", so the first automation is now "Trash: skip the occurrence if the bin did not go out". The fixture, the entity id and the probe automation carry the same name, and `test_recipes_match_readme` proves that both sides moved together. Code comments and docstrings are in scope, so the test module and both fixture files follow the same rules.

## Test results

```
tests/integration/test_automation_recipes.py ..... 11 passed
tests/integration                            ..... 179 passed
tests/unit                                   ..... 1663 passed, 2 skipped
ruff check / ruff format                     ..... clean
vale README.md                               ..... 0 errors, HomeKeeper style loaded
ci/mutation_scope.py --language python       ..... empty, no mutable code changed
```

CI is green on `b771312` across all 19 checks, including `vale`, `integration` and `e2e`.

## A note on the CHANGELOG

There is no entry and no version bump, because no behaviour of the integration changed. `notify-issues` therefore does not close #280 on a release. To give the reporter a notification, this branch needs a bullet with `(Fixes #280)`. Version 0.20.0 is now released, so per `AGENTS.md` that bullet belongs in a new `0.21.0b1` section. Ask for it and I will add it.

This pull request stays a draft. Mark it ready when you accept the example.

## Checklist

- [x] Tests run locally, `pytest tests/unit` plus the full `tests/integration` suite
- [x] `CHANGELOG.md` updated. Not applicable, because no behaviour changed. See the note above.
- [x] Panel UI changed. Not applicable.
- [x] New user-facing UI surface. Not applicable.
- [x] New user-facing feature. Not applicable. This branch changes documentation and tests.

https://claude.ai/code/session_01UmCodYKUtSFLTyQtYSWHhU


prestomation:
/q review Please review this critically and skeptically, most serious issues first, and don't withhold minor ones.

This PR documents two Home Assistant automation recipes in README.md and backs them with an integration test that drives the real HA container.

Topics I want scrutinized:

- **Correctness of the documented templates.** The follow-up recipe derives the just-passed occurrence as `as_datetime(task.next_due) - timedelta(days=7)`. Where does that break? DST transitions in the America/New_York-style zones, a task whose interval is not 1 week, a user who edits the task's anchor between the two automations running, a completion recorded before the occurrence it is meant to satisfy, or a task that was never completed at all (`last_completed` null). Is the grace-window comparison off by one at either boundary?
- **Failure modes I did not document.** `hk.tasks | selectattr('name','eq','Take out trash') | first` raises if no task matches — is that the right behaviour to ship in a copy-paste recipe, or should the README guard it? Same question for `task.id` when the filter yields nothing.
- **Advice quality.** Is a standing *triggered* task genuinely the right primitive for a weekly follow-up, or does the one-off variant have fewer sharp edges than the README implies? I claim a triggered task is born armed and needs one completion to go dormant (see `models.build_task` vs the comment at `recurrence.py:249`) — please confirm or refute.
- **Test rigour.** `test_recipes_match_readme` parses yaml fences out of README.md — is that gate as tight as it claims, and can it pass while the loaded automations differ? Do the scenario tests actually assert enough to catch the recipe being wrong, or would a mutated template still pass? Are the `time.sleep`/poll waits racy on a loaded CI box?
- **Fixture blast radius.** `recipes.yaml` and `recipes_probe.yaml` load in the shared e2e/screenshot HA config. The one-off variant fires on `home_keeper_task_completed` for any task named "Take out trash" and is only disabled by the test's own fixture. Is that safe for the other suites, and is `automation.turn_off`/`turn_on` state leaking between tests?
- **HA best practices** for the yaml itself (modern `triggers`/`actions` syntax, `mode`, response variables, service-call authorization from an automation context).

---

prestomation:
/q review Second pass, after 071c3bb. Please be critical and skeptical, most serious issues first, and don't withhold minor ones.

The only change since your last review is prose in README.md's "Deadlines and follow-ups" section, to clear a `vale` `ai-tells.VerbTricolon` failure. No yaml and no test code changed.

The earlier review passed without findings, so please push harder on the substance this time rather than the diff:

- **Are the documented templates actually correct?** `as_datetime(task.next_due) - timedelta(days=7)` is claimed to be "the occurrence that just passed". Where does that break — a DST boundary, a task whose interval is not exactly 1 week, an anchor edited between the two automations running, or a `last_completed` that predates the occurrence it should satisfy? Is the grace-window comparison off by one at either end?
- **Unhandled shapes in copy-paste code.** `selectattr('name','eq','Take out trash') | first` on a name that doesn't match, and `task.id` on the result. Should the README guard these, or is failing loudly correct here?
- **Is the advice right?** I recommend a standing *triggered* task over a dated one-off. Argue the other side. I also claim a triggered task is born armed and needs one completion to go dormant — confirm or refute against `models.build_task` and the comment at `recurrence.py:249`.
- **Does the test suite actually constrain the recipes?** Would a mutated template still pass `test_automation_recipes.py`? Is `test_recipes_match_readme`'s yaml-fence parsing as tight as it claims? Are the poll/sleep waits racy on a loaded runner?
- **Fixture blast radius.** `recipes.yaml` and `recipes_probe.yaml` now load in the shared e2e/screenshot HA config, and the one-off variant fires on `home_keeper_task_completed` for any task named "Take out trash", disabled only by the test's own fixture. Safe for the other suites? Does `automation.turn_off`/`turn_on` state leak between tests or across a container restart?

---

prestomation:
Thanks — four of your findings were real and are fixed in 8a4b0d4.

**`mode: single` was missing.** Correct, and it mattered: both time-triggered recipes read a task and then mutate it, so overlapping runs would have the second act on state the first already moved. Added to all three documented automations and to the fixture, so the parity test stays satisfied.

**Weekly-only, and the `| first` footgun.** Both now called out in the README. I kept the fail-loud behaviour — we agree it's right for a named-task automation — but readers now know it's deliberate.

**Boundary coverage.** This was the strongest finding: you were right that a mutated grace window would have survived. Two new tests, both verified by actually mutating the fixture and watching them fail:

| mutant | result |
| --- | --- |
| `grace_hours: 13` | `test_grace_window_boundary_is_twelve_hours` fails on "12h05m should be outside 12h" |
| `>= as_datetime(occurrence)` clause removed | `test_completion_before_the_occurrence_does_not_arm_the_follow_up` fails |

The window's floor genuinely had nothing constraining it before — dropping that clause left the whole suite green. The floor test back-dates a completion with `move_completion` to land before the occurrence. A third test covers a never-completed task (null `last_completed`).

I did not add a `days=8` test: that mutant is already killed, since the existing 3h-after-occurrence case lands 27h after a mutated occurrence and falls outside the 12h window.

**Racy waits.** `_run` no longer sleeps for a second. It waits on the automation entity's own `last_triggered` advancing and `current` returning to zero, with a timeout.

Two I did not act on, with reasons:

- **Unique test task names.** The README recipe uses "Take out trash", and `test_recipes_match_readme` requires the fixture to be that text verbatim, so renaming the tasks would mean the tested automations are no longer the documented ones. That trade isn't worth it. No other suite uses those names, and the autouse fixture disables the one-off variant at *setup* as well as teardown, so a crashed run doesn't leave it armed for the next test.
- **Double-completion in one window.** `last_completed` is the latest completion either way, so the condition's behaviour is unchanged. A test would restate the existing in-window case.

**Your `recurrence.py:249` finding is confirmed and fixed.** Your reading matches mine: `models.build_task` special-cases only `REC_SENSOR` to `next_due = None`; a triggered task falls through to `compute_next_due`, which returns `now`. The comment conflated the two. Corrected in the same commit.

Full `tests/integration` is 178 passed; unit 1592 passed; ruff, mypy scope and vale clean.

---

prestomation:
/q review Third pass, on 8a4b0d4, which acts on your last review. Stay critical and skeptical, most serious first, minor ones included.

Changed since your last look: `mode: single` added to all three documented automations and the fixture; README gained an "adapt this" paragraph (weekly-only, and the deliberate fail-loud task lookup); three new tests (`test_grace_window_boundary_is_twelve_hours`, `test_completion_before_the_occurrence_does_not_arm_the_follow_up`, `test_never_completed_task_does_not_arm_the_follow_up`); `_run` now polls `last_triggered`/`current` instead of sleeping; and the stale comment at `recurrence.py:249` is corrected.

What I want you to attack this time:

- **The new boundary tests.** `test_grace_window_boundary_is_twelve_hours` puts the occurrence 11h55m and 12h05m before a completion recorded at "now". Is that margin large enough to be stable, given the completion timestamp is whatever the container's clock says when `complete_task` lands? Is there a clock-skew or execution-lag path that makes 5 minutes too tight?
- **`move_completion` in the floor test.** I assume re-timestamping a completion leaves a fixed task's `next_due` anchored where it was, so the derived `occurrence` doesn't move under the test. Verify that against `store.move_completion` and the recurrence code, because the test's meaning depends on it.
- **`_run`'s new wait.** Does polling `last_triggered` plus a falsy `current` actually mean the action sequence finished and its service calls were applied to the store? Is there a window where `current` reads zero before `skip_task`/`trigger_task` has persisted?
- **`mode: single` interaction with `automation.trigger`.** The tests call `automation.trigger` back-to-back in one test (`test_cancelled_occurrence_never_arms_the_follow_up` runs CANCEL then ARM). Can `mode: single` now drop a run and silently green a test that should fail?
- **The `recurrence.py` comment edit.** Is what I wrote accurate about `build_task`, and does it stay true for the `merge_update` path that converts an existing task to `triggered`?
- Anything else in the diff, and anything you flagged before that you think I dismissed too easily.

---

PR #286, opened 2026-09-02
chore(deps): aggregate the open Dependabot bumps

prestomation:
## What changed

One branch replaces eight open Dependabot pull requests. Four of them edit the same
root lockfile and two edit the same frontend lockfile, so merged one at a time each
one makes the next conflict. Aggregating resolves the tree once.

| PR | Package | From | To |
| --- | --- | --- | --- |
| #217 | `@stryker-mutator/core` (root) | 9.6.1 | 10.0.0 |
| #218 | `@stryker-mutator/vitest-runner` (root) | 9.6.1 | 10.0.0 |
| #257 | `@vitest/ui` (root) | 4.1.9 | 4.1.11 |
| #258 | `vitest` (root) | 4.1.10 | 4.1.11 |
| #256 | `vitest` (frontend) | 4.1.10 | 4.1.11 |
| #259 | `rollup` (frontend) | 4.62.4 | 4.62.5 |
| #202 | `setuptools` build requirement | >=83.0.0 | >=84.0.0 |
| #201 | `actions/github-script` (ha-beta.yml) | v8 | v9 |

### Notes on the lockfile diff

Both lockfiles are regenerated with `npm install`, so transitive packages move to the
newest version each range allows. Two parts of the diff are larger than the table
suggests:

- **`rollup` lands on 4.63.1, not 4.62.5.** `^4.62.5` allows it and a fresh resolve
  takes the newest match. 4.62.5 has no property 4.63.1 lacks.
- **Stryker 10 pulls Babel 8.** Its instrumenter moved to the Babel 8 line, which
  accounts for most of the root lockfile churn.

Stryker 10 requires Node >=22. Only `mutation.yml` and `test.yml` install the root
lockfile, and both run Node 22.22.2, so Stryker never lands on a Node 20 runner. Every
other workflow installs `custom_components/home_keeper/frontend` or `tests/e2e`, and
neither depends on Stryker.

`actions/github-script@v9` is what `release.yml`, `preview-release.yml` and
`walkthrough-preview.yml` already use. This brings the last v8 call site in line. The
script it runs uses `github.rest.issues.*` and `core.notice`, both unchanged in v9.

### Verification

| Check | Result |
| --- | --- |
| `bash ci/test-frontend.sh` | 847 passed, 27 files |
| `pytest tests/unit` | 1592 passed, 2 skipped |
| `bash ci/build-panel.sh` | both bundles built on rollup 4.63.1 |
| `pip install .` (Python 3.13) | wheel built under `setuptools>=84.0.0` |

The mutation gate is diff-scoped and no mutable source changed here, so it does not
exercise Stryker 10. A full-surface `ci/test-mutation-frontend.sh --all` reached
Stryker 10's dry run — config accepted, 7 files instrumented, 2,719 mutants, 755 tests
green — before it was stopped. The first mutation run on a PR that touches the mutable
surface is what confirms the score, so watch that one.

Closes #201, #202, #217, #218, #256, #257, #258 and #259. Each is closed by hand
after this merges, because the commit here is not the one Dependabot wrote.

## Checklist

- [x] `CHANGELOG.md` updated — not needed, every bump here is a development
      dependency or a CI action
- [x] Panel UI changed → no source change, so no screenshots
- [x] New user-facing feature → none, so no beta bump and no `preview-release` label

https://claude.ai/code/session_01BSXQ5UGKge5mfajdBSjwWB

### DIFF
--- a/.github/workflows/ha-beta.yml
+++ b/.github/workflows/ha-beta.yml
@@ -197,7 +197,7 @@ jobs:
       # One open issue, reused. A breaking beta stays broken for weeks, and a fresh
       # issue every night would bury the original diagnosis.
       - name: Find or create the tracking issue
-        uses: actions/github-script@v8
+        uses: actions/github-script@v9
         with:
           script: |
             const label = 'ha-beta-regression';
--- a/custom_components/home_keeper/frontend/package-lock.json
+++ b/custom_components/home_keeper/frontend/package-lock.json
@@ -11,50 +11,16 @@
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-typescript": "^12.3.0",
         "@rollup/plugin-virtual": "^3.0.2",
-        "rollup": "^4.62.4",
+        "rollup": "^4.62.5",
         "tslib": "^2.8.1",
         "typescript": "^6.0.3",
-        "vitest": "^4.1.10"
-      }
-    },
-    "node_modules/@emnapi/core": {
-      "version": "1.11.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.11.1.tgz",
-      "integrity": "sha512-RSvbQmHzdKzNsLYa/wHrbc3KN4sYLKAdPZxqiM2HATqv/SBk2/ENSHpvXGaLOMcsAyz0poEGqkmmKYG3OWiJEQ==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "dependencies": {
-        "@emnapi/wasi-threads": "1.2.2",
-        "tslib": "^2.4.0"
-      }
-    },
-    "node_modules/@emnapi/runtime": {
-      "version": "1.11.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.11.1.tgz",
-      "integrity": "sha512-vgj7R3y3Wgx24IQaGPA/R6YFXLHVMOZ0uVEyIQPaWs+rd1AzfEMXlAC22FYwO1XkKR6NPsq7mUandH8oIRdZFw==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "dependencies": {
-        "tslib": "^2.4.0"
-      }
-    },
-    "node_modules/@emnapi/wasi-threads": {
-      "version": "1.2.2",
-      "resolved": "https://registry.npmjs.org/@emnapi/wasi-threads/-/wasi-threads-1.2.2.tgz",
-      "integrity": "sha512-c95qOXkHdydNKhscBTebqEC1CVAZpyqOfVfBzQ1qgzyl3gfeldUjIggDbIZgDKsHLgnsM+igH7TJ/eAasaVuMA==",
-      "dev": true,
-      "license": "MIT",
-      "optional": true,
-      "dependencies": {
-        "tslib": "^2.4.0"
+        "vitest": "^4.1.11"
       }
     },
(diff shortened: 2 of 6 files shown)


prestomation:
/q review Be critical and skeptical. This aggregates eight Dependabot bumps into one branch and regenerates both lockfiles with `npm install`, so transitive packages move further than the eight pull requests asked for. Look hardest at:

**Correctness of the aggregation.** Does the regenerated tree hold the versions the eight pull requests asked for, or did a resolve drop one? `rollup` lands on 4.63.1 rather than the 4.62.5 in #259, and Stryker 10 pulls the Babel 8 line — say if either widens the blast radius past what a dependency bump should carry, or if a lockfile entry looks wrong (a missing integrity hash, a dropped optional platform binary, a duplicated package that used to be deduped).

**The Stryker 9 to 10 major bump.** Stryker 10 requires Node >=22 and `mutation.yml` pins 22.22.2, but check every other workflow that could reach `npx stryker` on Node 20. Does `stryker.conf.json` still use only options Stryker 10 supports — `coverageAnalysis: perTest`, `disableTypeChecks`, `ignorePatterns`, `jsonReporter`, `tempDirName`, `vitest.configFile`? Flag any that changed shape or were removed. Also check `ci/mutation_report.py` and `ci/mutation_scope.py` against the Stryker 10 JSON report schema — the report format is where a major bump quietly breaks a parser.

**`actions/github-script` v8 to v9.** v9 changes the runner Node version and the bundled Octokit. Check the script in `ha-beta.yml` for anything that moved: `github.rest.issues.getLabel` / `createLabel` / `listForRepo` / `createComment` / `create`, and `core.notice`.

**`setuptools>=84.0.0`.** Does anything in CI build this package on a Python that setuptools 84 dropped, and is the floor doing real work or is it a bump for its own sake?

**Anything I have missed.** Please raise the most serious findings first, and include the minor ones as well.

---

prestomation:
Thanks — the Stryker config, `github-script` and setuptools checks match what I found. The Node 20 finding does not hold, though, so I am not changing those two workflows.

**Stryker cannot reach a Node 20 runner.** It lives in the root lockfile, and only two workflows install that lockfile:

| Workflow | Install | Node |
| --- | --- | --- |
| `mutation.yml` | `npm ci` (root) | 22.22.2 |
| `test.yml` | `ci/install-deps.sh`, whose first line is `npm ci` (root) | 22.22.2 |

Every Node 20 workflow the review names installs a different lockfile:

- `ha-beta.yml` lines 46, 92, 150 → `npm ci --prefix custom_components/home_keeper/frontend`
- `ha-beta.yml` line 98 → `npm ci --prefix tests/e2e`
- `e2e.yml` lines 21, 27 → the same two prefixes

Neither `custom_components/home_keeper/frontend/package.json` nor `tests/e2e/package.json` depends on Stryker, so `npx stryker` on those runners has nothing to find. `mutation.yml` is also the only workflow that mentions Stryker or mutation testing at all, and it has no `workflow_dispatch` path into `ha-beta.yml` or `e2e.yml` — so "someone manually triggers a mutation test there" is not a state those workflows can reach.

**The remaining question is whether the root lockfile now trips `npm ci` on Node 20** — a lockfile carrying `engines: {"node": ">=22.0.0"}` entries. It does not: `npm` treats `engines` as advisory and prints `EBADENGINE` unless `engine-strict=true`, and this repository has no `.npmrc` at either level. It is moot anyway, because no Node 20 workflow installs that lockfile.

Node 20 in `ha-beta.yml` and `e2e.yml` is a real inconsistency, but it predates this change and is independent of it. Raising the floor there would change what the nightly early-warning job tests, which is more than a dependency bump should carry. Worth its own change.

---

PR #238, opened 2026-08-24
feat: declarative companions + Device Pulse preset

prestomation:
Fixes #231

## What changed

Adds a **declarative companion** system to Home Keeper: instead of hand-coding a glue integration in a separate repo for every ecosystem, a user (or a shipped preset) describes a companion by config — target integration + entity filters + sensor-task trigger + Jinja-templated task fields — and Home Keeper materializes one managed sensor task per matching entity. The recipe reuses the existing sensor-task runtime (`SensorTaskWatcher`) so there is one runtime engine, not two.

### Screenshots

**Settings → Companions with the new Declarative subsection (empty state):**

Settings → Companions with the new Declarative subsection empty state

**The preset picker — Device Pulse is greyed out because the Device Pulse integration isn't installed in the e2e container; Low battery and Firmware update available are enabled:**

Preset picker modal with the three shipped presets, Device Pulse disabled with 'Requires the device_pulse integration' warning

**The Add dialog seeded from the Low Battery preset — every form section (Identity, Which entities?, Trigger, Task template) with the live-preview panel on the right showing "0 of 0 matches" (the e2e container ships no battery binary_sensors):**

Add declarative companion dialog seeded from the Low Battery preset, with the live-preview panel on the right

### What ships

- **Three shipped presets** under Settings → Companions → *Add from preset*:
  - **Device Pulse** — targets studiobts/home-assistant-device-pulse's per-device `*_total_failed_pings` sensors (threshold mode, > 0 for 1h, clear on recover). Requires the Device Pulse integration.
  - **Low battery** — matches every `binary_sensor` with `device_class: battery` reporting `on`. Coexists with the Battery Notes glue.
  - **Firmware update available** — matches every `update.*` entity reporting `on`. Covers UniFi, ESPHome, HACS, Reolink, Bambu Lab in one recipe.
- **New `availability` sensor mode** — arms when an entity is `unavailable`/`unknown` for a hold interval, inverting the "no reading = do nothing" policy of the other three modes. No shipped preset uses it today; it exists for user-authored companions ("task me when my ZHA devices go offline") and covers the *"useful without a specific integration"* case from the issue.
- **Panel UI** — new "Declarative companions" subsection under Settings → Companions, with a preset picker (cards disabled + tooltipped when the required integration isn't installed) and an add/edit dialog. The dialog has a debounced live-preview panel showing how many entities match and what the first ten task names will render as, with a soft warning at >50 matches and a hard 500-match cap so a runaway regex can't fan out silently.
[...]

PR #287, opened 2026-09-02
feat: declarative companions, availability sensor mode (rebase of #238)

### LINKED ISSUE #230: Companion suggestion: Device Pulse
Hi,

I think a good companion integration for Home Keeper would be [Device Pulse](https://github.com/studiobts/home-assistant-device-pulse), which monitors the availability of IP-reachable devices. This way, when a device monitored by Device Pulse is unavailable for a while (a configurable duration maybe), a task could be created to reconnect/troubleshoot it. And when the device comes back online, the task can be completed.
Device Pulse creates an entity to track reachability for each device it monitors, so the integration could be based on that.

### LINKED ISSUE #231: [New feature] Declarative Companions
Instead of having to hand-code a new "Companion" integration for each new integration to integrate with, I believe it should be possible to create declarative companions that would be configured from:
- a target integration
- a regex to identify the entities to monitor (among those owned by the integration)
- the same config options as for sensor-based task creation
- templated task name and notes, where the sensor and its parent device (if any) are injected into the context

prestomation:
Fixes #231
Fixes #230

Supersedes #238, which fell behind `main`. This branch carries the same feature, rebased onto `main`.

## What changed

- **Declarative companions**: a recipe (target integration + entity filters + a sensor-task trigger + Jinja task templates) that Home Keeper expands into one managed sensor task per matching entity. Two bundled presets: **Device Pulse** and **Firmware update available**. New `availability` sensor mode, offered both as a recipe trigger and as a mode in the ordinary task form.
- **Onto main**: the services, websocket commands and events are declared in `api_surface.py`; the store keeps both the declarative-companion map and main's to-do-list-item map; `strings.json` and the 16 translations keep both sets of exception keys.
- **Panel**: the declarative UI is a region module (`panel-declarative.ts`) over `PanelHost`, using `makeDialog` and `ha-form`.
- **No Low battery preset.** The [Battery Notes glue integration](https://github.com/prestomation/ha-home-keeper/blob/7af3c6323292a80c918910189a5f4218d8b93446/docs/GLUE_INTEGRATIONS.md) already opens a task per battery and also supplies the battery type and count, so a bundled recipe for it only duplicated that. Removing it left no shipped preset the committed fixture could match, so the fixture gains one template `update` entity and a new integration test installs `firmware_update_available` from the catalog the way the picker does — the coverage the removed low-battery test used to carry.

## Defects found while verifying, all fixed here

Each has a test that fails without its fix.

| What | Where |
| --- | --- |
| Reported by dfischer0 on #230/#231: switching a threshold-seeded recipe to State kept `comparison` in the trigger and the backend rejected the save. | `triggerForMode`, unit-tested |
| Edit and Delete from the panel never worked — the spec id was sent as `id`, which HA's websocket client overwrites with its own message id. | `companion_id` on both commands |
| `ha-dialog` fires `closed` when it leaves the DOM, so a re-render closed a dialog mid-edit. | `makeDialog` ignores a disconnected dialog |
| The add dialog's body was wider than `ha-dialog` and clipped its switches. | dialog CSS |
| A recipe opened no work for a condition that already stood when the task was made. | `sensor_watcher` skips the baseline for just-made tasks |
| A managed task's notes were frozen at creation, so an armed task showed the reading it was made with. | notes re-rendered on arm |
| A `for_seconds` hold only ended at the next reading, and an entity that goes offline sends none. | `hold_due_at` + a real timer |
| **An availability task rendered with the usage meter's strings** — "Every of use" and "Target 0 (sensor.x)". | `recurrenceText` and `sensorProgress` gain the mode |
| **The task form could not represent an availability task.** Editing one fell through to the *threshold* leg and Save stamped `comparison`/`value` onto the binding, which the backend rejects — the same class of bug as #230. | the form offers the mode, with its own fields |
| **The recipe preview stuck on "Loading preview…" for good after a trigger-mode change.** The rebuilt dialog scheduled its preview, then the dying dialog's handler replaced that timer with one pointing at its own detached node. | `schedulePreview` yields to the live dialog |
| The delete dialog said "can't be undone" twice. | one of them removed |

## Screenshots

The panel surfaces the docs carry, captured against the committed e2e fixture.

The Companions card with the Declarative companions subsection, its two Add buttons and the empty-state line

Choose a preset dialog with two cards: Device Pulse greyed out with a Requires the device_pulse integration line, and Firmware update available

The Add dialog seeded from Firmware update available, previewing Showing 1 of 1 matches

## What it looks like in use

An exploratory walk through the whole feature in a seeded container, with two stub integrations standing in for real upstreams: **Device Pulse** (three monitored devices, each with a `*_total_failed_pings` counter) and a set of appliances with consumables, a run-hours meter and an entity that can be taken away. Every shot below is a real run, not a mock-up. This walk is what found most of the defects in the table above.

### 1. Nothing configured yet

Settings → Companions. The **Declarative companions** subsection sits under the connected companions, with its two entry points and an empty-state line.

The Companions card with an empty Declarative companions subsection

### 2. The preset picker

Two bundled recipes. Device Pulse is pickable here because the demo container has that integration installed — the committed e2e fixture does not, which is where the disabled card is photographed.

Choose a preset dialog with the Device Pulse and Firmware update available cards

### 3. Add, seeded from the preset

Every field the recipe carries, and a live preview naming the three entities it would turn into tasks. The hold is shortened from the preset's one hour to 20 seconds so the walk can watch it expire.

The Add declarative companion dialog seeded from Device Pulse, previewing 3 of 3 matches

### 4. Saved

The recipe takes a row of its own, with the preset badge and how many tasks it currently owns.

The Companions card with a Device Pulse recipe row reading 3 matched tasks

### 5. The tasks it made

One task per matching entity, each attached to its device and marked **Monitored** — nothing is asking for work while the devices answer.

The task list with three Check on tasks under the Monitored group

### 6. A managed task's own page

The task is owned by the recipe: Duplicate is off, Delete redirects to the recipe, and the Sensor row names the binding and its live reading.

The Check on Synology NAS detail page showing its sensor binding and rendered notes

### 7. The condition arrives

`device_pulse.set_failed` drives the NAS to 4 failed pings. After the 20-second hold the task leaves Monitored and falls due.

The task list with Check on Synology NAS now in the Overdue group

### 8. …and the note is re-rendered

Compare the notes line with shot 6: the task was made when the counter read 0, and the note now reads the value that armed it. That is one of the fixes in this branch.

The armed task's detail page, its note reading 4 failed pings

### 9. The device answers again

`clear_on_recover` closes the task by itself and records a completion at the recovered reading.

The task back under Monitored with one completion in its history

### 10. Editing the recipe

Edit opens the stored recipe prefilled, previewing cleanly — a recipe does not overlap itself.

The Edit Device Pulse dialog, prefilled, previewing 3 of 3 matches

### 11. Switching the trigger mode

Threshold → State. Comparison and Value are gone and **State to watch for** has appeared, which is the bug dfischer0 reported. The preview has also re-run: before the fix in this push it stayed on "Loading preview…" until the dialog was closed.

The same dialog in State mode with the threshold fields removed and the preview refreshed

### 12. A second recipe over the same entities

The preview names the recipe that already covers them and how many, before anything is saved.

The Add dialog warning that Device Pulse already covers 3 of these entities

### 13. A usage recipe

A meter-mode recipe on a cumulative run-hours sensor: 20 of 50 hours used, 30 to go, and the list chip counts down in hours rather than days.

A Service Bambu X1C task showing 20 of 50 used with a progress bar

### 14. The meter passes its target

The same task, now due.

The task list with Service Bambu X1C in the Overdue group

### 15. An availability recipe

The entity is taken away and stays away past the hold. The Recurrence and Sensor rows describe the mode in its own words — before this branch they borrowed the meter's and read "Every of use" and "Target 0".

An armed availability task reading When it becomes unavailable

### 16. It comes back

Auto-clear closes it and records the completion.

The same task back under Monitored with one completion

### 17. Three recipes

Threshold, usage and availability side by side, each reporting how many tasks it owns.

The Companions card listing three declarative recipes

### 18. Deleting one

The confirmation says what goes with the recipe. It used to say "can't be undone" twice.

The delete confirmation for the Offline for a while recipe

## One-way doors

Unchanged from #238 — the declarative-companion spec shape (`selection` / `trigger` / `task_template`), the `source.declarative_companion` block on a managed task (`spec_id`, `entity_registry_id`, `entity_id`), the three `home_keeper_declarative_companion_*` events and their payloads, and the four `home_keeper.*_declarative_companion` services. New in this push: **`availability` as a value of `sensor.mode`**, which now reaches the store through the ordinary `home_keeper.add_task` / `update_task` services as well as through a recipe.

**Removed before release:** the `low_battery` preset id and its two `declarative_preset.low_battery.*` strings. It has never shipped in a stable release, so no stored recipe can carry `preset_id: "low_battery"`.

## Local verification

On this head: `pytest tests/unit` 1843 passed, `bash ci/test-frontend.sh` 1060 passed, `ruff check` / `ruff format --check` clean, `bash ci/build-panel.sh` clean, `npx tsc --noEmit` clean, `bash ci/test-mutation-frontend.sh` **100.00%** (55/55, threshold 80%).

Against the container on the committed fixture: `pytest tests/integration/test_declarative_companion_sync.py` 11 passed, the declarative Playwright spec 6 passed, and the walkthrough tour 2 passed.

`mypy` was **not** run locally: Home Assistant is not installed in this environment and its Python is 3.11, below HA's floor, so a local install would silently resolve to a stale HA (the #199 trap). CI runs it.

## Checklist

- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with HTML image tags

https://claude.ai/code/session_01R22nhT8BmBs6XiaUbCRTgq


prestomation:
/q review — please give this a **critical, skeptical** review of the takeover of #238 on top of current `main`; most serious first, and do not withhold minor findings. Areas to scrutinize:

**Correctness**
- The merge itself: `store.py` now carries both the declarative-companion map and main's to-do-list-item map through `load` / `_data_to_save` / `reset`. Is either one dropped on any path, and is a pre-existing store without those keys still loaded with the right defaults?
- `api_surface.py` declares the four services, seven websocket commands, three events and a new `declarative_companion` payload spine. Do the admin flags, `supports_response` kinds and payload key order match the registrations and `events.declarative_companion_event_data`?
- The websocket schema change from `id` to `companion_id` on `update_declarative_companion` / `delete_declarative_companion`: any caller left sending `id`? The service twins in `services.yaml` still take `id` on purpose.
- `panel-dialogs.ts` `makeDialog`: the `closed` handler now ignores a disconnected dialog. Can a real user dismissal ever arrive after disconnection and be lost, leaving `_completion.open` / `_moveCompletion.open` / `_declDialog.open` stuck true?
- `panel-declarative.ts` `triggerForMode` and `TRIGGER_DEFAULTS`: does the allowlist match `models.normalize_sensor` for every mode, including `usage` dropping `for_seconds` / `clear_on_recover`, and can `pickFormData` seeding still leak a foreign key back into the draft?
- The trigger-mode change re-renders the whole panel from inside an `ha-form` `value-changed` handler. Any re-entrancy or lost-focus hazard beyond the intended one?

**Maintainability**
- Region-module boundaries: `panel-declarative.ts` only reaches the `PanelHost` members added to `panel-host.ts`. Anything that should have been passed as an argument instead?
- `panel-settings.ts` counts declarative recipes into the Companions section mark and summary. Is that the right reading of "companions"?

**Tests**
- `tests/integration/test_declarative_companion_sync.py`: are the settle timings (`SETTLE`, the pre-flip pause) robust on a slower CI runner, and does the fixture leave the seeded store exactly as found?
- `tests/e2e/tests/declarative-companions.spec.ts`: the delete test polls the store before the row. Any remaining assertion that can pass transiently on a re-render?

**Security**
- Every mutation stays admin-gated on both the service and websocket paths; Jinja rendering stays `parse_result=False`; the 500-match hard cap still applies to `preview_declarative_companion`.

**HA best practices**
- The seeded `binary_sensor.hk_demo_remote_battery` template in the e2e config, and the `homeassistant.helpers.dispatcher` stub in `tests/unit/ha_stubs.py`.

---

prestomation:
Thanks. Triage of the three suggestions, no follow-up issues needed:

- **Shared `TRIGGER_DEFAULTS`**: the two sides are different languages, so a shared constant is not available. The frontend table exists only to seed what the form displays; the backend stays the validator, and `normalize_sensor` rejects anything the table gets wrong.
- **Mode-switching coverage**: `tests/e2e/tests/declarative-companions.spec.ts` drives a real threshold → state switch in the browser and saves it, and `frontend/test/panel-declarative.test.js` covers `triggerForMode` for each mode transition (12 cases), including the combinations the reporter hit on #230/#231.
- **Unit-testing `TRIGGER_KEYS_BY_MODE` against backend validation**: covered by the same vitest file for the frontend half; the backend half is `tests/unit/test_models.py`'s `normalize_sensor` cases, and the e2e save round-trips the two.

---

prestomation:
Exploratory testing found four defects that the test suite missed. All four are fixed in `e39d410` and `18a8478`.

I stood up a container of realistic devices — a robot vacuum with four consumables, a Brother laser printer, an air purifier, a UniFi router, an ESPHome node, a smart lock and a 3D printer — and pointed recipes at them the way a user would.

**The one that mattered: a recipe reacted only to a change, never to a condition that already stood.**

Two routers had a firmware update waiting. The Firmware update available preset opened three tasks and every one stayed asleep, so the panel showed no work for two updates that were pending right then. A later off→on transition armed correctly, and on→off cleared correctly, so the machinery was sound — the first run was not. The same held for the Low battery preset with a battery that was already low.

Materializing a task that owns entities reloads the config entry, and the reload takes a baseline of every monitored task. The baseline records an already-true condition as met without a crossing, which is right after a restart and wrong for a task made a second ago. It also explains why the integration test passed: its battery sensor has no device, so no reload happens and no baseline eats the edge. Every real integration gives its entities devices, so every real user got the broken path.

The reconciler now names the tasks it just materialized and the baseline leaves their edge unset, so the first reading arms them. A task that existed before setup is untouched, so a snoozed task still survives a restart — `test_a_reload_with_the_sensor_already_on_does_not_rearm` passes unchanged.

**Three smaller ones.**

- A note reading `{{ state }} h left` was rendered once at creation. It still said "145 h left" after the sensor fell to 3 and armed the task. The notes are rendered again on the arm.
- Two recipes could cover the same entities with no sign of it, giving every entity two identical tasks. The preview now says when another recipe already covers the matches, and names it.
- A blank recipe seeded its name from `{{ friendly_name }}`, which already carries the device name, so entity ids came out as `sensor.roborock_s7_replace_roborock_s7_filter_time_left_next_due`. It now uses `{{ device_name or friendly_name }}`, as the bundled presets do.

**Verified as correct while exploring:** the usage meter end to end (baseline at creation, arms at the target, a completion re-anchors it, arms again one interval past the new baseline), the availability mode arming and recovering, auto-clear recording a real completion, one recipe covering four consumables on one device, and the task list placing armed tasks in Overdue and dormant ones under Monitored.

**One thing left alone:** adding a recipe reloads the config entry, so a `home_keeper.*` service call in that window returns 400. An automation that calls a service at exactly that moment fails and would need to retry.

New coverage: three integration tests (the device-backed first-run case, the reload rule that must not break, and the notes re-render), unit tests for `holds_edge_state` and the overlap calculation, and an e2e case for the overlap warning.

Local gates on this head: `pytest tests/unit` 1759 passed, `pytest tests/integration/test_declarative_companion_sync.py tests/integration/test_sensor_watcher.py` 32 passed, `vitest` 991 passed, `ruff` clean, `mypy` clean against real Home Assistant, mutation 85.30%.

---

prestomation:
/q review Be critical and skeptical, and surface the most serious problems first. Do not withhold minor ones. This push removes the `low_battery` preset and finishes the new `availability` sensor mode, so focus on:

**Correctness.** `frontend/src/forms.ts` — the new `availability` leg in `taskSchemaSections`, `taskFormData` and `buildTaskPayload`. `normalize_sensor` defaults `clear_on_recover` to **True** in this mode and only an explicit `False` disables it, so the payload sends a real boolean here while the other modes omit a falsy flag. Is that asymmetry safe on every path — a mode switched mid-edit, a partially seeded draft, an `update_task` that merges rather than replaces? Also check `frontend/src/panel-detail.ts` `sensorProgress`: it classifies availability three ways (available / unavailable / not found) and is meant to mirror `sensor_watcher.read_availability_status`. Does it actually agree with that function, including the attribute-binding case and `unknown`?

**Race conditions.** `frontend/src/panel-declarative.ts` `schedulePreview` now returns early when its `preview` node is detached. A trigger-mode change re-renders the dialog from inside the section's change handler, and both the old and new dialogs share the `decl-preview` debounce key. Is the guard sufficient, or is there an ordering where neither dialog ends up owning the timer and the preview stays on its loading line?

**Maintainability.** Four `if (mode === …)` chains now exist across `forms.ts`, `utils.ts` and `panel-detail.ts`, each of which had to gain the same new case, and each of which silently falls through to the usage meter when it does not. Is there a shape that makes a fifth mode fail loudly instead of rendering "Every of use"?

**Test quality.** `tests/integration/test_declarative_companion_sync.py::test_a_shipped_preset_installs_and_materializes_as_the_picker_installs_it` asserts an exact rendered task name and an exact `sensor` dict. Is either over-specified against Home Assistant behaviour that could reasonably change? And is the new fixture entity (`update.hk_demo_router_firmware`) genuinely the only `update` entity the preset can match, so the "1 of 1 matches" the e2e spec and the screenshot both assert is stable?

**HA best practices and security.** The fixture gained a template `update` entity; anything about that config or about the availability mode's interaction with the entity registry that would break on a future HA release?

---

prestomation:
The `e2e` job went red on `7af3c63`. It was not this branch's change, and `7e7137b` fixes it.

**What failed.** `card-registration.spec.ts` › "renders from an app shell that never imported its bundle", on all three attempts, at its own guard:

```
Error: the strip left a reference behind
expect(received).not.toContain("home-keeper-card.js")
```

**Why.** That spec simulates the #228 case — a service-worker-cached app shell from before the integration was installed — by removing the card's inline `import()` from the shell HTML. Home Assistant **2026.9** changed `index.html.template` to wrap each extra module in a rejection handler:

```jinja
import("{{ extra_module }}").catch(function (err) {
  console.error("Failed to load extra module {{ extra_module }}", err);
});
```

The handler repeats the URL in its message, so stripping the `import(...)` call alone leaves `home-keeper-card.js` sitting in the shell, and the spec's guard against an incomplete strip fires. The `e2e` job pulls the floating `stable` tag and that image was rebuilt today (`2026-09-05T13:00Z`); my local one was from `2026-08-21`, which is why this passed locally and failed on CI. It would fail the same way on `main`.

**The fix** takes the chained `.catch(...)` with the import. The group is optional, so the bare `import("<url>");` older releases render is still stripped.

**Verified** against the new image, with CI's own headless-shell browser rather than the sandbox's full Chromium: the spec fails on all three attempts without the change and passes with it, and the whole suite is 226 passed. No other spec broke on 2026.9.

Everything else on `7af3c63` was already green, including `integration`, `mypy`, `vale`, `walkthrough` and both mutation jobs.

---

prestomation:
/q review Be critical and skeptical, and surface the most serious problems first. Do not withhold minor ones. The previous review returned praise with no findings and did not engage with the questions asked, so please answer these directly rather than summarising the diff.

New in `7e7137b`: `tests/e2e/tests/card-registration.spec.ts` now strips Home Assistant 2026.9's `.catch(...)` rejection handler along with the card's `import()`.

**Correctness of the new regex.** The catch body is matched with `[^{}]*`, which cannot span a nested brace. If Home Assistant later puts a brace in that handler, does this fail loudly (the strip guard fires) or silently pass while leaving a reference behind? Is the optional group genuinely a superset of the old bare-`import("<url>");` shape, or does the `\s*;?` tail now behave differently on it?

**Fragility.** This spec is pinned to an upstream HTML template on a floating `stable` tag, so it breaks whenever Home Assistant restyles that block. Is there a formulation that expresses "the shell does not import this bundle" without matching upstream's exact syntax — stripping the whole `<script>` element that names the bundle, or parsing rather than regexing — and what would that cost in fidelity to what #228 actually simulates?

Also still unanswered from the previous request, and more important than the above:

**The four `if (mode === …)` chains** in `frontend/src/forms.ts`, `frontend/src/utils.ts` and `frontend/src/panel-detail.ts`. Each had to gain the same new `availability` case, and each silently falls through to the usage meter when it does not — which is exactly how this branch shipped "Every of use" for an availability task. Is there a shape that makes a fifth sensor mode fail loudly instead of rendering the wrong mode's words?

**The `clear_on_recover` asymmetry** in `buildTaskPayload`: availability sends a real boolean while the other modes omit a falsy flag, because `normalize_sensor` defaults it to True in that mode only. Is that safe on every path — a mode switched mid-edit, a partially seeded draft, an `update_task` that merges rather than replaces?

---

prestomation:
Brought the branch up to date with `main` (`75d80e5`). The PR was un-mergeable; three files conflicted.

**How each conflict was resolved**

| File | Resolution |
| --- | --- |
| `CHANGELOG.md` | Both sides added to `## [0.21.0b2]`. Kept every bullet: this branch's two **Added** entries plus main's Test-button entry, main's **Changed** entry, and this branch's **Fixed** entry. Nothing dropped. |
| `tests/e2e/tests/card-registration.spec.ts` | Both sides fixed the same thing — HA 2026.9 wraps the extra-module import in a `.catch(...)` handler that the strip has to take with it. Took main's version: its `[\s\S]*?` body matcher spans a multi-line handler, where this branch's `[^{}]*` does not. |
| `docs/images/21-panel-companions.png` | Kept this branch's capture. It is the only one that shows the **Declarative companions** subsection the PR adds. |

**Verified on the merge commit**

- `pytest tests/unit` — 1843 passed
- `bash ci/test-frontend.sh` — 1060 passed
- `bash ci/e2e-up.sh` — **226 passed**, the full Playwright suite on both projects, including `card-registration.spec.ts` (the spec whose conflict I resolved toward main) and `declarative-companions.spec.ts`
- `ruff check` / `ruff format --check` / `npx tsc --noEmit` / `bash ci/build-panel.sh` — clean

`mypy` was not run locally, for the reason the PR body gives: Home Assistant cannot be installed in this environment at a Python above HA's floor, so a local install resolves to a stale HA (#199). CI runs it.

/q review This push is a merge of `main` into the branch — please give it a **critical, skeptical** review, most serious first, and do not withhold minor findings. Focus on the merge itself rather than re-reviewing the feature:

**Correctness of the merge.** `store.py`, `api_surface.py`, `strings.json` and the 16 translations each carry two independent sets of additions now (this branch's declarative-companion surface and main's notification/Test-button and season work). Did any auto-merged hunk drop a key, a service, an event or a translation string that one side added? Check `load` / `_data_to_save` / `reset` in `store.py` in particular, and whether a pre-existing store missing either map still loads with the right defaults.

**The resolved conflicts.** Is taking main's `CARD_IMPORT` regex in `card-registration.spec.ts` right, or does this branch's variant catch something main's misses? And does the merged `## [0.21.0b2]` CHANGELOG section still satisfy the three-sentence budget and the `(Fixes #N)` placement rule for every bullet?

**Interaction between the two feature sets.** `panel-settings.ts` merged main's Profile companion-filter work beside this branch's declarative subsection. Does `companionOptions` / the rail's `companionsOn` count treat a declarative recipe consistently with a glue integration, and can a recipe-owned task be selected by a Profile's `companions` filter the way a user would expect?

**Anything the auto-merge made stale.** A comment, a docstring or a test name that described the pre-merge shape and is now wrong.

---

prestomation:
Fixed the panel-goes-stale bug you reported: after adding a companion from a preset, the Companions list kept its old contents until the page was reloaded by hand.

## What was happening

The panel reads everything it shows in **one** `Promise.all` — tasks, appliances, options, companions, declarative recipes, tags. `_reload` assigns nothing when that batch rejects, so a failed load left the whole panel on pre-change data, with no message on screen and nothing to try again. That is why every list was stale, not only the one you touched.

The batch fails because **Home Keeper reloads its own config entry**. Adding a declarative companion that matches entities materializes tasks, and `_async_reconcile_and_maybe_reload` reloads the entry to baseline the sensor watcher. An options change reloads it too. `find_coordinator` returns `None` while the entry is unloaded, so every Home Keeper command answers `not_loaded`, and the refresh that runs right after such a save is exactly the one that lands in that window.

Confirmed against the e2e container with a websocket probe across a real entry reload:

```
508ms home_keeper/get_tasks FAIL code=not_loaded msg=Home Keeper isn't loaded right now.
```

Whether the refresh loses the race is a timing question, which is why the suite never caught it: in the e2e container the batch finishes in tens of milliseconds and the reload arrives seconds later. On a real install the batch is slower and the window is wider.

## The fix

- `api.isNotLoaded` names the error.
- `_reload` waits it out: five tries a second apart, spent **only** on `not_loaded`, so a genuine failure still gives up on the first try.
- The per-call `.catch(() => fallback)` guards no longer swallow it. Falling back there renders "no companions, no options, no recipes", which is a confident wrong answer; rethrowing puts the batch on the retry path.
- A load that fails after the panel is up now says so. The load-error UI only ever replaced the first-load spinner, so a running panel had no way to report it.
- The recipe dialog's Save shows a busy state, because the dialog now stays up until the panel has re-read the store.

## Coverage

`custom_components/home_keeper/frontend/test/reload-during-entry-reload.test.js` fails on **three of its four cases** without the retry (verified by reverting it): the recipe list, the rendered Companions card, and the task list each stay behind. The fourth asserts a plain failure is not waited out.

No e2e case for the race itself: HA replaces the custom-panel element a few seconds after an entry reload (the capture script already notes this), so a browser-level test of the window is not deterministic here.

## Versioning

`v0.21.0b2` was published from `main` on 2026-09-05, so this branch's bullets could not stay in that section — `changelog-release-gap` fails a PR that edits a released one. They now sit in a new `## [0.21.0b3]` alongside the new fix, with `manifest.json` and `PANEL_VERSION` bumped to match. `## [0.21.0b2]` is left holding exactly what the tag shipped.

## Verification on this head

`pytest tests/unit` 1861 passed, `bash ci/test-frontend.sh` 1072 passed, `ruff check` / `ruff format --check` clean, `npx tsc --noEmit` clean, `bash ci/build-panel.sh` clean.

Screenshot re-captured on this head (the surface is unchanged at rest; the fix adds a transient busy state on the dialog's Save):

Settings with the Companions card, its two connected companions and the Declarative companions subsection

`vale` is not installed in this environment, so the new bullet was checked against the pinned `ai-tells` rule regexes directly, per AGENTS.md. Worth noting: the first draft tripped `VerbTricolon` (two commas after a modal, spanning two sentences of the one bullet) and was rewritten until the rules came back clean.

/q review This push fixes a stale-panel bug. Please be **critical and skeptical**, most serious first, and do not withhold minor findings.

**Correctness of the retry.** `_reload` now recurses on `not_loaded` with a counter. Is the recursion sound against `_refresh`'s `_refreshing` coalescing — can a caller awaiting `_refresh()` observe a half-applied state, or can two retry chains overlap? What happens if `hass` is replaced, or the element is disconnected, part-way through a five-second retry: does the panel write into a detached element, leak a timer, or resume against a stale `this._hass`?

**Is the failure set right?** Only `not_loaded` is waited out. Are there other transient websocket errors around an entry reload that deserve the same treatment, and is there any path where `not_loaded` is *permanent* (the entry removed, or setup failing repeatedly) so that five tries a second apart is just five seconds of nothing before the toast?

**The `_soft` change.** Rethrowing `not_loaded` from the best-effort guards makes the whole batch fail where it used to degrade. Does that trade a stale panel for a *failed* one in any case that used to work — a user without admin rights, an older backend missing a command, or a partially-loaded entry?

**Root cause rather than symptom.** Should `_async_reconcile_and_maybe_reload` avoid the entry reload altogether for the add path, given the reload is what breaks the very refresh the panel does next? Is baselining the sensor watcher achievable without it, and would that be the better fix?

---

prestomation:
Correction to the comment above: the screenshot URL there carries a wrong commit SHA and returns 404. The screenshot, pinned to `9c967ad`:

Settings with the Companions card, its two connected companions and the Declarative companions subsection

---

PR #289, opened 2026-09-03
fix(notifications): Test reports its delivery, and a Settings row keeps its edit

### LINKED ISSUE #255: Feature Request: Notification Channels

prestomation:
Fixes #255

## What changed

Started from the follow-up reports on #255. Both reported bugs are ours; chasing the second one turned up a third, and reviewing that turned up two more in my own fix. Five commits.

### 1. "No task is due. Home Keeper sent nothing." on a Test that did send

`home_keeper.notify` answers `sent` with the **task id** the walk surfaced (`null` for a digest, and for an empty queue); `matched` carries the count. The panel read `sent` as a count:

```js
const { sent } = await api.runNotification(hass, current().id);
toast(p, sent > 0 ? t('notify.test_sent') : t('notify.test_none'));
```

`"t1" > 0` coerces to `NaN > 0` → `false`, so **every** walk that delivered reported as empty, and a digest (`sent: null`) did too. Verified against the real service before touching anything: `{"matched": 1, "sent": "t1"}`. It now reads `matched`, and `NotifyRun.sent` is retyped `string | null` so the misread cannot come back silently. The service contract is unchanged — it is published in `services.yaml` and automations depend on it.

**Why CI was green.** The vitest fixture invented `sent: 1`, a value the service cannot produce. And `waitFor` returns `null` on a timeout instead of throwing, so `await waitFor(() => toasts.includes('Notification sent.'))` was a wait and never an assertion. Both fixed; every toast case now has a real `expect`.

### 2. "It says Saved, but the field is blank after a refresh"

The reporter [narrowed this](https://github.com/prestomation/ha-home-keeper/issues/255#issuecomment-5533513099) from "the channel picker isn't working" to a save that does not stick. Reproduced in a real browser on **both** HA 2026.8.3 and 2026.9.0 — type a channel into one notification, start on the next, and the first is discarded:

```
row 1  channel ""       <- typed "Trash"
row 2  channel "Meds"
```

`persistDebounced` had **one timer per option list**, so touching a second row inside the 600 ms window cleared the first row's pending save. It also built the payload when the timer was **armed**, so even with a per-row timer the second row would write a list still holding the first row's pre-edit value. Both halves are load-bearing: reverting either turns the tests red. The timer is now keyed `<list>:<row id>` and the list is built when the timer **fires**. Profiles autosave through the same helper and had the same defect.

### 3. Out-of-order save answers (found by review, verified before acting)

Two rows can have writes in flight at once, and a websocket does not promise ordered answers. An early answer landing last put the panel's cached options back to before the later row saved, and the next row to build a list from that cache wrote the staleness to disk:

```
stored after two saves : ["Alpha", "Beta",  ""]
panel cache (stale)    : ["Alpha", "",      ""]      <- Beta lost
stored after a third   : ["Alpha", "",      "Gamma"] <- Beta gone for good
```

Same silent loss as #2, reachable the same way, so it is fixed rather than documented. `persistOptionList` takes a per-list sequence number and ignores its answer once a newer save has started. Worth recording: **serialising the sends would not have fixed this** — the first answer still lands before the second is sent, so the cache still drops to the stale state. The fix has to be on the answer.

### 4 and 5. Two defects in my own fix for #3

* The sequence check returned early, so a save another row overtook skipped `expandLast`, `_render()` and the toast as well as the stale cache. Pressing *Add notification* while another row's autosave was in flight would have looked like it did nothing. Narrowed to the cache assignment alone.
* `expandLast` then read the row out of the shared cache, which a superseded save never updates — so an add could park a blank id in `_itemExpanded`, and two rapid adds would open the wrong row. It now reads its own answer, and skips a row with no id.

### On the Android channel itself

The reporter's original framing was that the channel arrived on Android's "General". Bug 2 accounts for it: the channel was never stored, so nothing ever sent it. Home Keeper's send path is verified correct on both ends and unchanged here — the stored channel reaches the wire as the companion app documents:

```json
"data": {
  "tag": "home_keeper_n1", "group": "home_keeper", "actions": [...],
  "channel": "Trash", "importance": "high",
  "push": { "thread-id": "Trash", "interruption-level": "time-sensitive" }
}
```

Every delivery path (`build_notification`, `build_digest`, `build_all_clear`) goes through the one `payload_data` chokepoint. `_send` now also debug-logs the payload it sent, so a future "the channel did nothing on my phone" can be settled from `home-assistant.log` against the companion app's own Notification History rather than guessed at.

### Tests

* `frontend/test/settings-autosave.test.js` (new, 6 cases) — two rows back to back, out-of-order answers, the toast on a superseded save, add-expansion, no write storm, and the Profiles variant. Every case was run against each broken variant it is meant to catch (shared debounce key; per-row key with an eager list; the early-return guard; `expandLast` off the shared cache). A case that could not fail either way was dropped rather than shipped.
* `tests/unit/test_notifier_service.py` (new) — pins the `home_keeper.notify` response contract (`sent` is a **string** task id; `null` for a digest that delivered; `matched: 0` with no send for an empty queue) and asserts a saved channel and urgency reach `data.channel` / `data.importance` / `data.push` on the **service** path, not just in `payload_data` in isolation.
* `tests/unit/notifier_harness.py` (new) — the loader, fakes and pinned clock moved out of `test_notifier_blocking.py` so both suites share one notifier-under-test.
* `tests/e2e/tests/settings.spec.ts` — Test against a due task and a real target reads the right toast; two notifications configured back to back both keep their channel.

Run locally: `pytest tests/unit` (1612 passed, 2 skipped), `npx vitest run` (867 passed), the full `settings.spec.ts` green against **both** HA 2026.8.3 and 2026.9.0, ruff lint + format clean, `tsc --noEmit` clean, `vale` clean on the added CHANGELOG lines. mypy is not runnable here (HA will not install on this runner's Python 3.11 against its 3.14 floor); CI's mypy job has passed on every push.

### Screenshots

The panel surface does not change — the fixes are in which toast the Test button raises and in when a row's edit is written. A fresh capture of the Notifications card came back **byte-identical** to the committed image, so nothing is re-committed:

Settings → Notifications, showing a notification's channel, urgency and the Test button

## Checklist

- [x] `CHANGELOG.md` updated — two `### Fixed` bullets under a new `## [0.20.0b4]`, each with `(Fixes #255)`. Commits 4 and 5 fix regressions from earlier in this PR, so they get no bullet of their own.
- [x] Panel UI changed → capture re-run; identical to the committed `docs/images/22-panel-notifications.png`, embedded above
- [x] New user-facing UI surface → none; the tour needs no edit
- [x] New user-facing feature → not a feature, but `manifest.json` + `const.py` bump to `0.20.0b4` anyway: `0.20.0b3` is already released, and this repairs a feature its beta testers are using. No `preview-release` label (bug fix only).

https://claude.ai/code/session_01XBQXnMwPnd4Xkv9PhHkz3u


prestomation:
/q review Be critical and skeptical, and put the most serious findings first. Do not withhold minor ones.

Context: this fixes the panel's Test button reporting "no task is due" on a run that delivered. `home_keeper.notify` answers `sent` with the walked **task id** (`null` for a digest, and for an empty queue) and `matched` with the count; the panel read `sent` as a count, so `"t1" > 0` was false. It now reads `matched`.

1. **Correctness of the new predicate.** Is `matched > 0` really equivalent to "a notification went out" on every path through `notifier.async_run_notify` / `_send`? Trace the cases I may have missed: a notification whose `profile_id` points at a deleted profile, a `target:` override, `persistent_notification` as the only target, and the `_send` branch that returns `len(queue)` after warning about no targets — can any of them report `matched > 0` while nothing reaches a phone, or the reverse?

2. **The service contract.** `services.yaml` documents `sent` as "which was sent". Retyping it `string | null` in `types.ts` is a doc change on our side only, but is leaving the backend response shape alone the right call versus renaming the field? Consider automations already reading it.

3. **Test quality.** `waitFor` in `frontend/test/panel-harness.js` returns `null` on timeout instead of throwing, so a bare `await waitFor(...)` asserts nothing — that is why this bug shipped green. I backed each toast case in `notify-test-button.test.js` with an explicit `expect`, but I did **not** change `waitFor` itself. Is the file-local fix the right scope, or is every other `await waitFor(...)` in the suite a latent false-green worth fixing here? Are there specific ones you can see that assert nothing?

4. **The new suites.** `tests/unit/test_notifier_service.py` and `tests/unit/notifier_harness.py`: is the harness extraction from `test_notifier_blocking.py` faithful (the pinned clock, the `dr`/`ar` stubs, the module-cache guard in `load_notifier`)? Could the shared `hk.notifier` cache leak state between the two suites depending on collection order?

5. **The e2e case.** It locates the channel field as `form.locator('ha-selector-text').nth(1)`. Is the label assertion enough to keep that honest, or is it brittle? It also polls `home_keeper/get_options` through `page.evaluate` — any better in-repo idiom for that?

6. **The debug log in `_send`.** It now logs the whole payload. Any concern about volume, or about task names and notes reaching the log at debug level?

7. **HA best practices and performance** anywhere in the diff.

---

prestomation:
/q review Second push. Be critical and skeptical, most serious findings first, and do not withhold minor ones.

New commit `cd9d3c5` fixes the second half of #255, which the reporter narrowed to "it says Saved but the field is blank after a refresh". Reproduced in a real browser on HA 2026.8.3 and 2026.9.0: `persistDebounced` kept **one timer per option list**, so touching a second Settings row inside the 600ms window cancelled the first row's pending save; and it built the payload when the timer was **armed**, so even with a per-row timer the later save would write a list holding the earlier row's pre-edit value. The timer is now keyed `<list>:<row id>` and the list is built when the timer fires.

Please scrutinise:

1. **Correctness of the late build.** The claim it rests on is that `persistOptionList` sets `p._options` *synchronously* before its first `await`, so a row saving second already sees the row that saved first. Is that actually true on every path through `persistOptionList`, including the `catch`? What happens when the first save **fails** — the optimistic `p._options` is left holding the failed value, and the second row's rebuild will then re-send it. Is that the behaviour we want, or should a failed save roll back?

2. **Concurrency.** Two rows can now have timers in flight at once, so two `set_options` calls can overlap. Each sends the whole list. Walk the interleavings — can a slow first response clobber a faster second one, given the backend merges per key rather than per row? Would per-row saves be better serialised through a queue?

3. **Deletes and adds.** `deleteNotification` / `addNotification` / `deleteProfile` call `persistOptionList` directly and do not clear pending per-row timers. I believe the late build makes a pending edit *stop* resurrecting a deleted row, because the rebuilt list maps over the current options and the row is gone. Verify that, and check the add path (a row added while another row's timer is pending).

4. **The debounce key.** `${key}:${itemId}` shares one `_persistTimers` map with every other `_debounce` caller in the panel. Any collision risk with an existing key? Are ids guaranteed non-empty here?

5. **Test quality.** `frontend/test/settings-autosave.test.js` — I checked each case fails against both broken variants (shared key; per-row key with an eager list) and dropped a fourth case that could not fail either way. Are the three that remain actually discriminating, or is one of them passing for an incidental reason? Is the "no write storm" assertion (`saves` length 2) over-specified?

6. **Anything I have missed in the same helper**, and HA best practices / performance across both commits.

---

prestomation:
Thanks — findings 1, 3, 4 and 5 confirmed as you described. Acted on 1 and 2; here is what changed in `93d9f6a`.

**#2 (concurrent saves) — this one is real, and it is a data-loss bug, not just a race.** I built a test double that applies writes in arrival order but answers them out of order, and it reproduces:

```
stored after two saves : ["Alpha", "Beta",  ""]
panel copy (stale)     : ["Alpha", "",      ""]      <- Beta lost from p._options
stored after a third   : ["Alpha", "",      "Gamma"] <- Beta gone from disk
```

The third edit is what makes it permanent: it rebuilds from the corrupted copy and writes it back. So it is the same silent loss as the bug the previous commit fixed, reachable by the same behaviour (editing several rows in a row), which is why I fixed it rather than documenting it.

Your proposed serialiser would **not** have been enough on its own, which is worth recording. With the sends queued, the first answer still lands before the second is sent, so `p._options` still drops to the stale state — and a third row firing in that window still builds from it. The fix has to be on the answer, not the send: `persistOptionList` now takes a sequence number per list and ignores its answer when a newer save has started. Nothing is lost by ignoring it, because the newer write already carries every earlier row's value.

**#1 (catch path)** — agreed it is acceptable, and it is not a regression: the eager build had the same behaviour. Documented in the catch block as you suggested, with the reasoning (a rollback would pull the text out of the field under the user on a transient failure, which is worse than re-sending it).

**#6.1 (typing `buildList`)** — leaving as is. The conditional-return type you sketched does not work here: `OptionListKey` is the union, not the call's key, so it would need `persistDebounced` to be generic over the key. That is real complexity for a helper with two call sites that are correct by construction.

**#6.3 (no timer cancellation on delete)** — intentional, and confirmed rather than assumed. The late build is what makes it safe: the timer fires, maps over the current options, and the deleted row simply is not there to be resurrected.

Local after the change: 865 vitest, 1612 pytest, `settings.spec.ts` green on HA 2026.8.3 and 2026.9.0, ruff/`tsc` clean.

---

prestomation:
/q review Third push (`93d9f6a`), acting on your last review. Be critical and skeptical, most serious first, and do not withhold minor ones.

`persistOptionList` now claims a per-list sequence number and **discards its answer when a newer save has started**, so an out-of-order answer cannot put `p._options` back to a stale state. I verified your concern #2 was a real data-loss path first, and also that a send-side serialiser alone would not have fixed it.

Please scrutinise specifically:

1. **Is the sequence check sufficient, and is discarding the answer safe?** The claim is that the newest write already carries every earlier row's value, because each list is built from `p._options` after the earlier save's optimistic update. Find a case where that does not hold — in particular the `expandLast` path (`addProfile` / `addNotification`), which needs the backend-assigned id out of its own answer. If an add is superseded by a debounced row save landing after it, does the new row still get expanded, and does its id still reach `p._itemExpanded`? Is dropping `render` and the `settings.saved` toast on a superseded save the right call, or can that swallow the only feedback for a save the user made?

2. **The WeakMap.** `saveSeq` is keyed by the panel host. Any way a host outlives its counter or gets a fresh one mid-flight (the panel element being replaced by Home Assistant, which this codebase already knows happens), and if so what breaks — a stale answer being *accepted* because the counter reset to 1?

3. **Counter overflow / reuse** — is `(byKey.get(key) ?? 0) + 1` sound for the lifetime of a session, and is the closure over `seq` correct if `persistOptionList` is called re-entrantly?

4. **Test quality.** The new `drops an out-of-date answer instead of writing it back over a newer row` case uses a `hass` double that applies on arrival and holds the first answer until the second is answered. Is that a faithful model of a Home Assistant websocket, or does it prove something narrower than it claims? It is red without the sequence check — but is it red for the right reason?

5. **Anything the three commits together get wrong**, plus HA best practices and performance.

---

prestomation:
/q review Fourth push (`9d3e294`) — please review this head, not `93d9f6a`. Critical and skeptical, most serious first.

**Question 1 of my previous request answered itself, and it was a real defect I introduced.** The sequence check in `93d9f6a` returned early, so a save another row overtook skipped `expandLast`, `_render()` and the "Saved" toast along with the stale options copy. Pressing *Add notification* while another row's autosave was in flight would then look like it did nothing and the new row would not open. `9d3e294` narrows the guard to the `p._options` assignment alone, with a test that is red against the early return. My earlier questions about `expandLast` and the swallowed toast are therefore settled — please check the narrowed version instead.

Still open from that request, on this head:

1. **The WeakMap.** `saveSeq` is keyed by the panel host. Can a host get a *fresh* counter mid-flight — Home Assistant replaces the panel element, which this codebase already knows happens — and if so, does a stale answer get wrongly **accepted** because the counter reset to 1?

2. **Is skipping only the assignment now too narrow?** A superseded save still runs `expandLast` against whatever `p._options` holds at that moment. Trace the add path: can it expand the wrong row, or add an empty id to `p._itemExpanded`?

3. **Counter arithmetic and re-entrancy** — `(byKey.get(key) ?? 0) + 1`, and the closure over `seq`.

4. **Test faithfulness.** The `outOfOrder` double applies writes on arrival and holds the first answer until the second is answered. Is that a fair model of an HA websocket, or does it prove something narrower than claimed? Are all five cases in `settings-autosave.test.js` discriminating, and is any of them green for an incidental reason?

5. **The four commits together** — anything they get wrong as a set, plus HA best practices and performance.

---

prestomation:
Your #2 is right and is fixed in `e6f684e` — we arrived at the same patch independently, reading the row out of `merged` rather than `p._options`. I added the `id` guard too, since the blank an add sends before the backend names it would otherwise sit in `_itemExpanded` for the life of the panel. There is a test for it, red against the old version.

**Your #1 is not a defect, and the reasoning in it has a factual error worth correcting.** It says "after replacement `p` references the new host". It does not. `p` is a parameter captured by the closure in `claimSave`; a replacement panel is a *different* object, and nothing reassigns `p`. So an in-flight save from the replaced panel:

* checks `saveSeq.get(p)` against the **old** host's Map, which still has its own count, and
* on success assigns `p._options`, which is the **old**, detached element's field.

Neither reaches the new panel. A replacement starting its own count is correct rather than a gap, and keying by a stable id would be strictly worse: it would let a discarded panel's answer be compared against the live panel's counter, which is the corruption the current shape rules out by construction. I have written that into the comment on `saveSeq` rather than changing the code.

**Your #4 is a fair limitation, and I am not closing it in code.** You are right that the double applies writes serially on arrival and only reorders the answers. Interleaved backend application is a different scenario. It does not change the outcome here: every write is a full-list replacement built from `p._options` after the previous save updated it, so whichever applies last carries every earlier row. Modelling HA's internal interleaving faithfully in a unit double would be inventing a contract the websocket does not publish.

On "case 3 is only green if both the sequence check and the late list-building work" — that is by design, and I checked it rather than assuming. Each case was run against both broken variants (shared debounce key; per-row key with an eager list) and against the early-return version of the sequence check. The matrix is in the commit messages; no case is green for an incidental reason.

`#3` and `#5` — agreed, nothing to do.

Local on `e6f684e`: 867 vitest, 1612 pytest, `settings.spec.ts` green on HA 2026.8.3 and 2026.9.0, ruff/`tsc` clean.

---

prestomation:
**CI status on `e6f684e`: 15 of 17 checks green. `integration` and `e2e` are `cancelled`, and neither is this PR's.** Both were killed by their own `timeout-minutes` cap, not by a failing assertion.

| check | cap | ran for | where it died |
|---|---|---|---|
| `integration` | 10 min | 10m21s | 79% through the suite — **every test that ran passed**, including all of `test_notifications.py` and `test_panel_stability.py` |
| `e2e` | 15 min | 15m19s | still pulling the Home Assistant image: `e3b9155c10f4 Extracting 286.9MB/392.2MB`. HA never started and **no test body ran** |

The e2e one is the clearest: the entire 15 minutes went on environment setup. Earlier in this PR the same job spent **7 minutes on `npm ci` for 3 packages**. That is runner throughput, and it is the same pattern on the first push of this branch, so it is not a one-off either.

To be explicit about the one thing that could have made it mine: this PR adds two e2e cases to `settings.spec.ts` (~30s locally). They cannot be the cause, because the job was cancelled before Playwright started. Locally the full `settings.spec.ts` runs green in ~27s against HA 2026.8.3 and again against 2026.9.0.

I have already spent my one re-run on this branch (on the first push, same two jobs, same cause), so I am not re-running again — a third attempt would just be spinning.

**Not fixing this here.** The remedies are repo-wide CI changes — raising the two caps, or caching the HA image and the npm trees between runs — and neither belongs in a bug-fix PR about notification saves. Flagging it for @prestomation to decide as its own change.

---

prestomation:
/q review Fifth push (`a2e62d8`), a UI change on top of the four bug-fix commits. Critical and skeptical, most serious first, and do not withhold minor ones.

Each Settings card now reports its own autosave beside its name (*Saving…* / *Saved* / *Not saved*) and the success toast is gone. Errors still toast, because only the toast carries the backend's message. Both autosave paths report the same way: the row lists through `persistOptionList` and the option cards through `saveOptions`.

Please scrutinise:

1. **The pending counter.** `markSaving` / `markSettled` in `panel-settings.ts` count in-flight writes per card so the card leaves *Saving…* only when the last one answers. Can the counter get stuck above zero — an exception path, a save started against a card that then stops rendering, `_render()` mid-flight? A card stuck on *Saving…* forever is worse than the toast it replaced. Also check the `failed` stickiness: it is meant to persist until the next save starts, and I want to know if it can be cleared early or never.

2. **`aria-live` correctness.** The element is created once per render and mutated in place, specifically so the live region is in the document before its text changes. Is that actually sufficient here, given `_render()` replaces the whole subtree — does a fresh region built during a re-render announce its initial content spuriously? And is `role="status"` plus `aria-live="polite"` redundant or contradictory?

3. **Losing the toast.** Success feedback is now visual-only in the card header. On a phone the user may be scrolled well below the header while editing a row, so the status could be off-screen entirely. Is that a real regression versus the toast, and if so what is the least-bad fix — a per-row status instead of per-card, or something else? I deliberately did not add one, but I want it challenged.

4. **The `display: none` guard.** `.hk-save-status` is in the DOM from first render but hidden until the card has saved, so headers do not shift. Does hiding an `aria-live` region with `display: none` stop it announcing when it later becomes visible and populated in the same frame?

5. **Test quality.** Six new cases in `settings-autosave.test.js`. Are they discriminating, and is the `hold`/`fail` double faithful? In particular the "stays on Saving until the last write answers" case depends on write ordering in the double — is it green for the right reason?

6. **Anything the five commits together get wrong**, plus HA best practices and performance.

---

prestomation:
Checked all six against the code. #4 and #5 confirm what I believed, #1 and #2 do not survive tracing, and #3 stands as an open design question.

**#2 — the concern is inverted, and the code it describes does not exist.** `settingsCard` creates a **fresh** `inner` div per render, sets `innerHTML` on it, and only *then* appends the status (`panel-settings.ts` — `inner.querySelector('.hk-form-title')?.appendChild(...)` runs after the assignment). Nothing wipes the title, because the title is created by that same assignment a line earlier.

The real question underneath it is the opposite of the one asked. A live region inserted into the document *already holding* text does not announce that text — assistive tech announces mutations to a region that was already there. So a re-render landing mid-save risks **swallowing** an announcement, not emitting a spurious one. It does not, and here is the path: the only `render: true` caller is `persistOptionList`, which calls `_render()` **inside** the try and `markSettled` after it, so the new region is inserted showing *Saving…* and then mutated to *Saved* — the transition that matters is a mutation on an element already in the document. The debounced row saves use `render: false`, so nothing is replaced between the two states at all.

**#1 — no stuck counter, and nothing orphaned.** Both callers wrap the write so every path settles: `markSaving` at the top, `markSettled(…, true)` at the end of the `try`, `markSettled(…, false)` in the `catch`. The `if (!p._hass) return` guard sits *before* `markSaving` in both, so there is no early return that increments without decrementing.

On orphaned state: cards are destroyed and re-created with the same id on **every** `_render()` — that is the design, and it is exactly why the state lives on the host rather than in the DOM. The map is keyed by host, holds at most one entry per card id (five in total), and is weakly held. There is nothing to leak and nothing to clear.

**#3 — agreed, and left open deliberately.** You are right that the status can be off-screen on a phone once the user has scrolled past the card header. I am not taking the floating-status suggestion, because a single element pinned to the viewport is a toast again, and covering the form being edited is the problem this change exists to solve. The two real candidates are per-row status or nothing, and that is a product call rather than a review finding — it is with the maintainer.

Also on this push: `main` moved under the branch (#290 merged) and the PR went un-mergeable. Merged in and resolved — 16 locale conflicts, caused by #290 re-sorting those files, resolved by taking main's ordering and re-applying my one-key delta programmatically rather than by hand; and the `0.20.0b4` CHANGELOG section, where both sides had added a heading. 906 vitest and 1621 pytest green on the merge.

---

PR #290, opened 2026-09-03
feat(profiles): filter tasks by the companion that created them

prestomation:
A profile gains `companions` and `exclude_companions`, keyed on the
`managed_by.integration` that an owning integration stamps on every task it creates. A
dashboard card or a notification can then cover one source: "just the battery tasks", or
"everything except what Battery Notes raises".

A Profile filtered to the Battery Notes companion in Settings then Profiles

## Why this lives here and not in each glue

This started as a review of
[ha-home-keeper-battery-notes#21](https://github.com/prestomation/ha-home-keeper-battery-notes/pull/21),
where a contributor added a labels option to the Battery Notes glue so its tasks could be
scoped by a card. That works, but it does not scale: every glue reimplements the option,
and it applies to nothing that already exists.

Home Keeper already knows which integration owns a task. One axis here covers every
companion, applies to tasks already in the store, and asks the user to configure nothing.

## Design

`managed_by` is the right key by contract, not convenience. `docs/INTEGRATING.md` states
that Home Keeper **never inspects `source`**, while `managed_by` is the well-known block
it acts on, and both `integration` and `display_name` are required whenever it is present.

Semantics follow the existing axes: include lists OR within and AND across, exclusions
applied last and winning. No enrichment is needed, unlike labels, which inherit from
device and area.

**A task no integration owns has no companion.** A `companions` list never selects it and
an `exclude_companions` list never drops it. So `companions: [battery_notes]` gives a pure
battery card, and excluding that companion still shows the tasks you made yourself. That
holds for an absent `managed_by`, an explicitly null one, and a block missing its
`integration` key.

The picker offers connected companions **plus** any integration already owning a task, so
an unregistered owner stays pickable and an uninstalled one does not vanish from the list
while a saved profile still names it. Suggested-but-not-installed companions are left out:
they own no tasks, so filtering by one selects nothing.

## The combination rule is now written down

Adding a fourth filter made an undocumented rule more consequential, so this PR states it
in both places a user looks.

The panel gains a second Profiles intro line, `notify.filters_help`: **"A task must match
at least one value in each filter you set."** One sentence carries both halves. "At least
one value" is the OR inside a filter, "each filter" is the AND across them, and "you set"
keeps the blank filters out of it.

It is a new key rather than an extension of `notify.profiles_help`, because `i18n-parity`
checks that a key exists and not that its value is current. Extending the old key would
have left 15 languages silently on the shorter text. A new key fails that test until every
locale has it.

`README.md` gains a matching **How filters combine** subsection with the worked trap: a
profile with both Labels and Companions set shows the overlap, not the two groups
together. Issue #291 tracks making the combination configurable.

## Scope

Profiles only. A card scopes to an integration by selecting a profile, which is how the
card already consumes a profile's filter, so the card gains no new config keys. The card's
existing `hide_managed` boolean is untouched — "hide anything owned" is a different
question from "show me these sources".

## Verification

- `pytest tests/unit` — 1623 passed.
- `ci/test-frontend.sh` — 893 passed. `tsc --noEmit` clean.
- `ruff check` + format clean. `mypy custom_components/home_keeper` — no issues in 51 files.
- **Mutation: 100% both languages.** Python 177/177 killed; TypeScript 49/49. Two mutants
  survived the first Stryker run and both were real — a redundant optional chain, and a
  missing assertion that the exclude field is multi-select. Fixed rather than annotated.
- 14 cases added to `tests/fixtures/profile_filter_cases.json`, the cross-language
  conformance fixture, so the Python and TypeScript matchers are pinned to agree.

Translations for the four new strings are included for all 15 non-English locales, each
reusing that language's established word for "companion" from the existing
`companions.heading` key.

The tour is untouched: this adds fields to an existing profile form rather than a new
surface, so it stays on the screenshots gate.

## Review history

Cue found one real gap, an untested `managed_by: null`, now pinned on both sides. Its
other suggestion (guard before the Map lookup in `companionOptions`) was already
implemented. CI's vale caught three findings a local vale run missed, which is the pinned
-binary gap `AGENTS.md` warns about; all three are fixed. Both the `integration` and `e2e`
tiers were cancelled once at their timeout caps with every test passing, which is a
pre-existing flake rather than this branch. Each passed on a re-run, and the whole suite
then went green in one pass.

https://claude.ai/code/session_01BgrUw9GJch8LgWpsT94xKV


prestomation:
/q review Be critical and skeptical, most serious first, and do not withhold minor points. This adds a `companions` / `exclude_companions` axis to the Profile filter, keyed on `managed_by.integration`. Please scrutinize:

1. **Correctness of the matcher, both sides.** `profiles.matches_filter` (Python) and `card-filter.profileMatches` (TypeScript) must select identically for every input, since a notification is decided server-side and the card client-side. Look for divergence in the edge cases: `managed_by` absent, `managed_by` present but without an `integration` key, `managed_by: null`, an empty include list versus an absent one, and an exclude list that names a domain no task carries. Check that the 12 new cases in `tests/fixtures/profile_filter_cases.json` actually pin the agreement rather than just passing.

2. **The unowned-task rule.** A task with no `managed_by` is deliberately never selected by `companions` and never dropped by `exclude_companions`. Is that self-consistent, and does it match how `exclude_areas` treats an area-less task? Is there an input where the two halves disagree?

3. **The provenance key.** I used `managed_by.integration` rather than the `source` namespace, because `docs/INTEGRATING.md` says Home Keeper never inspects `source`. Is `managed_by.integration` reliably present in practice, and is anything in the codebase already treating `source` as provenance in a way this now contradicts?

4. **`companionOptions` in `forms.ts`.** It unions connected companions with integrations seen on tasks. Check the precedence rule (registered name wins over a task's `display_name`), the domain fallback, and whether a large task list makes this expensive given it runs on every profile-form build.

5. **Migration and persistence.** `normalize_filter` rebuilds from a fixed key set, so I claim a profile saved before this change reads back with empty lists and needs no migration. Verify that, and verify `options.current_options` re-normalizes on read as the docstring claims.

6. **HA best practices and typing** for the selector construction and the `NotifyFilter` type change.

---

prestomation:
Thanks — triaged. Pushed 9141821.

**Point 1, the missing `managed_by: null` case: valid, fixed.** You are right that it is a distinct shape from an absent key, and that both reach the matcher, since `add_task` stores the block it is handed. Both sides already resolved it to "unowned" (`(task.get("managed_by") or {})` and `task.managed_by?.integration`), but nothing held them to it. Added the include case you proposed plus its exclude twin, so the "never selected, never swept up" pair is pinned for null the same way it is for an absent key, along with a unit test either side.

Fixture is at 46 cases. 1623 Python / 893 TypeScript pass, Python mutation still 177/177.

**Point 4, the `if (!domain) continue;` optimization: already there.** `companionOptions` reads:

```ts
for (const task of tasks) {
  const owner = task.managed_by;
  const domain = owner?.integration;
  if (!domain) continue;          // <- before the Map lookup
  if (!names.has(domain)) names.set(domain, owner.display_name || domain);
}
```

The guard precedes the `has` check, so an unowned task never reaches it. No change needed. Worth flagging so the suggestion is not carried into a later review as outstanding.

One correction on the writeup: the `NotifyFilter` change *is* in the diff, in `frontend/src/types.ts` — both keys are declared there as required `string[]`, which is what keeps `addProfile`'s blank profile honest. `ProfileFilter` in `card-filter.ts` declares them optional on purpose, since that type also has to describe a profile saved before this change.

Agreed on the rest of the analysis, particularly the read of `normalize_filter` as a no-op migration and `current_options` re-normalizing on read.

---

prestomation:
**`integration` was cancelled on 9141821, and it is not this PR's failure.** Re-running it once; details so the next person does not re-derive them.

It is a `timeout-minutes: 10` kill, not a test failure. The job reached 86% of 167 tests with **every one passing**, then:

```
2026-09-03T23:16:13Z test_sensor_watcher.py::test_deleting_an_older_completion_leaves_the_meter_alone PASSED [ 86%]
2026-09-03T23:16:20Z ##[error]The operation was canceled.
```

The tier is flapping against its own cap, repo-wide:

| Run | Head | Duration | Result |
|---|---|---|---|
| 942 | `9141821` (this PR) | 10m26s | cancelled |
| 941 | `32db82e` (this PR) | 7m34s | success |
| 940 | `05022d0` (this PR) | 8m33s | success |
| 939 | `62fc9df` (this PR) | 10m24s | cancelled |
| 938 | **#289**, unrelated branch | 10m26s | cancelled |
| 937 | `main` | 6m28s | success |

Two things rule this PR out. #289 hit the same signature at the same duration tonight on a different branch, and two heads of this PR that differ from the current one only by two fixture cases and two unit tests — none of which run in this tier — passed in 7 to 8 minutes. This PR adds no integration test, and its runtime change is two list-membership checks in `matches_filter`.

The real cost sits in `test_sensor_watcher.py`, which spends about 2.5 minutes of the run on its own because those cases wait on real time backstops. A normal run finishes in 6 to 8 minutes against a 10-minute cap, so a slow runner tips it over.

**No fix exists to port.** Raising the cap or trimming the sensor-watcher waits would fix it, but that is a CI change unrelated to companion filtering and belongs in its own PR rather than widening this one. Flagging it because it will keep costing cycles on every PR until someone takes it.

---

prestomation:
**`e2e` was cancelled on 487ed1e, and it is not this PR's failure either.** Re-running it once. This is the same cap problem I flagged for `integration`, in a second tier, so recording it here rather than leaving it to be re-derived.

`integration` passed this time, in 6m24s. `e2e` is the one that died, and the way it died is worth seeing:

```
03:44:49Z   211 passed (3.8m)
03:44:49Z   Run docker compose down -v
03:44:54Z   Container integration-homeassistant-1  Removed
03:44:54Z   Post job cleanup.
03:44:54Z   Cleaning up orphan processes
```

**All 211 tests passed and the teardown finished.** The job still ended `cancelled`, at 15m06s against its `timeout-minutes: 15`. The suite itself is 3.8 minutes of that. The rest goes on setup: the panel build, the container pull and start, and `npx playwright install`.

So the tier is not slow because of its tests. It is slow to get to them, and the cap leaves so little headroom that a run can pass every assertion and still be killed while cleaning up.

The commit that triggered this changes `README.md` and nothing else.

Both tiers sit near their caps:

| Job | Cap | Typical | Seen |
| --- | --- | --- | --- |
| `integration` | 10 min | 6 to 8 min | cancelled twice tonight, also on #289 |
| `e2e` | 15 min | about 15 min, 3.8 of it tests | cancelled after passing |

Raising the caps, or caching the browser and image so setup stops eating the budget, would fix both. That is a CI change unrelated to companion filtering, so it stays out of this PR.

---

PR #292, opened 2026-09-04
release: 0.20.0, with the fixes its exploratory pass found

### LINKED ISSUE #220: A few feature requests and questions

prestomation:
Fixes #220

## What changed

An exploratory pass over everything since 0.19.0 — the shopping amounts and mirror, the buy-reminder sections, Duplicate, the notification channel/urgency and Test button, the Settings autosave, and the Profile companion/shopping filters — driven against the live e2e container. Four objective bugs, all in the shopping/Duplicate work, two of them raised on review. Everything else behaved: the shopping mirror renames a line in place rather than duplicating it, a partial restock keeps the reminder, and the Profile matcher agrees exactly between Python and the browser (12/15/2/13 selected on the same live task set).

### 1. A buy reminder read "Overdue" away from the task list

b6 taught the panel's list row and the dashboard card to say **Low stock**. Two other renderers draw the same status and were missed: a reminder's own detail page, and an appliance's Tasks tab. One task showed two different statuses depending on where you looked at it — and on the appliance page that red pill sat directly under the parts list already saying the part was low.

Four copies of one conditional is what let two of them drift, so the fix is a single `statusChipHtml` in `utils.ts` that all four surfaces call, beside the `isBuyTask` moved there from `card-filter.ts` (re-exported, so the pure list-shaping code still imports it from one place). It also settles a case the copies disagreed on: a reminder bought while its part stayed under its reorder point is filed under **Completed** by `statusBucket`, so the pill now runs the same completed-before-buy order and reads "Completed" rather than arguing with the section heading above it.

An appliance's Tasks tab: three maintenance tasks with red Overdue pills, and two buy reminders with amber Low stock pills

A buy reminder's detail page showing an amber Low stock chip and a greyed Duplicate button

### 2. Duplicate was offered on an auto-created buy reminder

Pressing it created a plain one-off carrying no `source.buy` — verified end to end, it produced `Buy Anode rod (copy)` with `source: null`. Nothing retired it on restock, the shopping-list mirror never saw it, and an `exclude_shopping` Profile could not drop it (confirmed: the orphan survived that filter in both matchers). It would sit in Overdue forever. `_canDuplicate` now blocks it, so it gets the greyed button and the same explanation the other reconciler-owned tasks give.

### 3. The Overdue filter counted the shopping tasks

Raised on review. A buy reminder is overdue by the clock — minted dateless, and a dateless one-off is due now — but it is not late work, and that is what the pill means. It read "Overdue 13" above a list whose own headings said Overdue 10 and Shopping 3, and clicking it drew a Shopping section under a heading that says Overdue.

The b6 note defended the count as keeping the panel in step with the sensors, but there is no aggregate overdue sensor to be in step with: the entities are a per-task `next_due` timestamp, a per-asset date, and a per-task `_overdue` boolean. Nothing publishes a total.

Fixed on the panel pill and the card's `filter: overdue` **together** — fixing one and not the other would recreate exactly the divergence bug 1 is about. This is also the reporter's own first choice in #220: *"Either pulling Buy/reorder tasks out of Overdue entirely, or just styling them distinctly"* — b6 did the styling, this does the pulling-out.

Deliberately **not** changed: a Profile's `status: overdue` still selects buy reminders, because `exclude_shopping` is the opt-out b6 shipped for that and a household may be relying on those notifications; the per-task `_overdue` binary sensor likewise stays on, since an automation may key off it. The split is safe because the pills are only rendered under `onTasks && !profile` — a Profile hides them, so no view asks both questions.

⚠️ **This changes what an existing dashboard card configured `filter: overdue` shows** — buy reminders move to `filter: shopping`. Shipped as a plain fix on the grounds that the card and the pill must not disagree about one word, and the changelog says so. Reverting is one line in `matchesFilter` if you would rather gate it. **Worth exercising specifically in the preview build below.**

### 4. A duplicate asked for less than its original

Capture mode is two fields: `completion_detail` says whether anything is mandatory on completion, `completion_required_fields` says which. `duplicateTaskSeed` carried the first and dropped the second, so a task requiring a note and a photo copied into one requiring the default note alone.

Half the fix is in `buildTaskPayload`, which never sent the list at all. The edit path never showed the bug because `merge_update` keeps the stored list when an update omits it — a duplicate goes through `add_task`, where there is nothing to keep. Sent only when non-empty, which loses nothing: `normalize_completion_required_fields` ends `return result or ["note"]`, so an empty list is not a state a `required` task can hold.

### Also: the last #220 answer, now pinned by a test

The reporter asked whether a `clear_on_recover` task linked to a part draws its stock down when the sensor recovers, or only when somebody presses Done. It was answered "yes" on the issue and nothing asserted it. It is true by construction — `sensor_watcher._apply_edge` clears the task through the same `store.complete_task` every surface uses, and the stock side-effect branches on the task's *source*, not on who completed it — but the failure mode is silent, so `test_clear_on_recover_draws_down_a_linked_consumable` now pins it. Verified by gating the draw-down on `origin != ORIGIN_SENSOR_RECOVER` and watching it fail with the stock unmoved. The part measures in millilitres and spends 250 per completion, so the amount is pinned too, not just the fact.

## Checklist

- [x] Tests run locally — `pytest tests/unit` (1663 passed), `npm test` (968 passed), full e2e (217 passed), full Docker integration (168 passed), and `ci/test-mutation-frontend.sh` at **100%** on the changed surface. Every fix and the new test was verified by reverting the code it covers and confirming failure
- [x] `CHANGELOG.md` updated — stable `## [0.20.0]` written for someone upgrading from 0.19.0, rolling b1–b6 into Added/Changed/Fixed. Beta-to-beta framing removed: the Test button and Duplicate are new in 0.20.0, so fixes *to* them are not separate bullets. Names #220, #255 and #279, so `notify-issues` closes all three; `release-issues.py --missing` reports nothing omitted
- [x] Panel UI changed → `45b-panel-shopping-detail.png` and `45c-panel-appliance-tasks.png` committed under `docs/images/` and embedded above with HTML `` tags. Captured by a new focused harness, `screenshots-shopping-status.capture.ts`, rather than two more steps in the main tour — matching `screenshots-companion-filter` and the other focused captures, since the tour is one 8-minute test that has to walk the whole panel to reach this state
- [ ] New user-facing UI surface → n/a, no new surface; the fixes correct how existing ones render
- [x] Version bumped — `manifest.json` + `const.py` to `0.20.0`
- [x] `preview-release` label applied, so `0.20.0.dev292` is installable from this PR head via HACS before merge. Worth a look at the two behaviour changes a test cannot judge for you: an existing card configured `filter: overdue` no longer lists buy reminders, and the Overdue pill count drops by however many you have

https://claude.ai/code/session_01JEcVryF8asZbg133wAv8W2


prestomation:
/q review Be critical and skeptical, most serious issues first, and don't withhold minor ones.

This PR consolidates four copies of the task status-pill conditional into one `statusChipHtml` in `frontend/src/utils.ts`, moves `isBuyTask` there from `card-filter.ts` (re-exported), and blocks Duplicate on an auto-created buy reminder in `panel.ts`.

Please scrutinise:

**Correctness.** Is the new ordering in `statusChipHtml` right — buy-before-overdue, but completed-before-buy? It is meant to mirror `statusBucket` in `card-filter.ts`; check the two cannot disagree for any task shape, in particular a buy reminder that is completed while its part is still under its reorder point, one with a null/malformed `next_due`, and a repeating task carrying `source.buy`. Also check the `elapsed` day count: floor division, `>= 1` boundary, and the timezone behaviour of `next_due` parsing.

**Parity.** `isBuyTask` requires both ids truthy; Python's `reconcile.buy_source` requires both keys *present*. Is `{asset_id: "", part_id: ""}` reachable through `add_task` from a companion, and does the divergence matter for `exclude_shopping` (matched in JS for the panel and card, in Python for notifications)?

**Regression risk.** Four call sites changed. Did moving `isBuyTask` to `utils.ts` create an import cycle (`card-filter.ts` imports `utils.ts`)? Are the removed `isOverdue`/`dueLabel` imports genuinely unused now? Does the panel list row still get its danger rail on exactly the tasks it did before?

**Maintainability.** Is `statusChipHtml` the right home and the right seam, or does the `opts.elapsed` flag leak the list row's concern into a shared helper?

**HA best practices** and anything about the greyed-Duplicate accessibility path (`role=button`, keyboard) that the buy-reminder case now newly exercises.

---

prestomation:
Thanks — that matches what I found. Taking the points in turn:

**Parity.** Agreed it is unreachable, and I checked the two ways in before deciding: `reconcile_buy_tasks` mints uuids, and `models.build_task` is the only other path. So no guard, but I have recorded the divergence on `isBuyTask` in `bb9ece6` — including that the agreement the fixture actually pins is the *missing*-id case, not the empty-string one. If that ever stops being unreachable the symptom is a task one surface excludes and another does not, which is worth being able to recognise.

**`opts.elapsed`.** Landing on the same trade-off. Worth adding that the flag is not only the list row's preference: on a buy reminder the elapsed count is actively wrong, because "3 days overdue" there means "low for 3 days". That is handled before `elapsed` is consulted, and `utils.test.js` pins it ("says Low stock even where the row asks for elapsed days") so a future caller cannot reintroduce it by passing the flag.

**Malformed `next_due`.** One correction to the reasoning, not the conclusion: a malformed value does not reach the day count. `isOverdue` parses it, `NaN <= now` is false, and the chip falls through to `dueLabel`. The `: 0` branch you cite only covers a *null* `next_due` on a task that is somehow already overdue, which is unreachable for the same reason. Behaviour is right either way.

One thing the review did not cover, so to be explicit about it: the **Overdue filter pill still counts buy reminders**, which is why the pill reads 13 where the Overdue section holds 10. That is deliberate and unchanged from b6 — only the sections and the wording move, so no count disagrees with the binary sensors or a Profile. `shopping-rows.test.js` asserts the pill count directly.

---

prestomation:
/q review Second pass on `bb9ece6`, which is a comment-only change. Please be critical and skeptical, most serious first, minor ones included.

Two things I would like scrutinised that the first pass did not reach:

1. **The release cut itself.** This PR cuts stable `0.20.0`, not another beta. Check `CHANGELOG.md`'s `## [0.20.0]` section against the rules in `AGENTS.md`: is every bullet inside the three-sentence budget counting the bold lead, is each `(Fixes #N)` in its bullet's first paragraph, and is the section written for someone upgrading from 0.19.0 rather than carrying beta-to-beta framing? `manifest.json` and `const.py` both read `0.20.0` — is anything else version-pinned that I have missed?

2. **Test quality, not just presence.** The mutation gate scores 100% on the changed surface, which is exactly the number that should invite suspicion. Are the new assertions in `utils.test.js`, `shopping-rows.test.js` and `task-duplicate.test.js` actually pinning behaviour, or are any of them shaped to kill a mutant without testing anything a user would notice? In particular the `statusChipHtml` cases: would they still fail if the buy-before-overdue ordering were reversed, or if `boughtAlready` dropped one of its three conjuncts?

Also worth a look: `screenshots-shopping-status.capture.ts` mutates the committed seed fixture in `beforeAll` and restores it in `afterAll`. Is the restore complete, and is there a failure mode that leaves the fixture dirty?

---

prestomation:
Checked both findings by experiment rather than by reading. One does not hold; the other is right on the mechanism but wrong on the consequence.

**Ordering is pinned — the finding does not hold.** I reversed it in the working tree so the overdue chip wins over a buy reminder, and ran the suites:

```
× says Low stock on a buy reminder, never Overdue
× says Low stock even where the row asks for elapsed days
× needs all three marks of a bought reminder before it stops saying Low stock
× reads as low stock, not as overdue work
× reads as low stock on its own detail page
× reads as low stock in an appliance's related tasks
Tests  6 failed | 166 passed
```

Six failures across `utils.test.js` and `shopping-rows.test.js`, including the panel-level ones. The reason the conjunct test is not merely mutant-shaped: each variant drops one conjunct while leaving a task that is still an *open* buy reminder, so each case only passes if the buy branch is reached before the overdue and due-label branches. I also tried the weaker reversal (hoisting the `!isOverdue` early return above the buy check) — that one is a genuine no-op, since it only short-circuits tasks that are not overdue, and the suite correctly stays green on it. Worth being precise about which of the two you meant, but neither leaves the ordering unpinned.

**Fixture restore: right that a hard kill leaves it dirty, wrong that it is harmless.** `tests/integration/ha_config/.storage/home_keeper` *is* committed — `tests/integration/ha_config/.gitignore` ignores `.storage/*` and then re-adds it with `!.storage/home_keeper`. So a dirty fixture is a tracked-file diff, not scratch state, and "a full rebuild cleans it" does not apply.

The window is narrower than described, though: Playwright runs `afterAll` when the test body fails, and the restore is guarded on `assetId`, so an assertion failure or a timeout still restores. Only a killed process leaves it dirty. That is identical to `screenshots.capture.ts`, which mutates the same fixture the same way — I hit it myself during this work and recovered with `git checkout`. Pre-existing and shared, so not something to fix inside this PR; noting it here rather than silently.

No code change from either finding. Everything green locally: 963 frontend tests, 217 e2e, mutation 100% on the changed surface.

---

prestomation:
/q review Two more fixes in `c69a3dd`. Critical and skeptical please, most serious first, minor ones included.

**1. The Overdue filter no longer counts buy reminders** — changed in two places together, `scopeMatches` (panel pill) and `matchesFilter` (card `filter: overdue`). Deliberately *not* changed: a Profile's `status: overdue`, and the per-task `_overdue` binary sensor.

Please challenge that boundary. Is leaving Profiles and the binary sensor as they are defensible, or does it just move the inconsistency somewhere less visible? Specifically: a Profile with `status: overdue` now selects tasks the panel's Overdue pill does not — is there a surface where those two are shown side by side and would visibly disagree? And is there anywhere else in the codebase that means "overdue" and now diverges from the pill?

Note this changes what an existing dashboard card configured `filter: overdue` displays. Is that acceptable as a bug fix, or should it be gated?

**2. Duplicate now carries `completion_required_fields`** — seeded in `duplicateTaskSeed` and sent from `buildTaskPayload`, the latter only when the array is non-empty.

Check the second half hardest, since it touches the **edit** path too, not just duplication. Is "send only when non-empty" right, or does it make a list impossible to *clear* through the panel once set by a service call? `merge_update` keeps the stored value when the key is absent, so I believe clearing was already impossible from the panel and this does not regress it — but please verify rather than take my word. Also: does sending the list on a task whose `completion_detail` is `none` or `optional` cause any harm, given `normalize_completion_required_fields` forces it empty for those modes?

Finally, is `completion_required_fields` the only field still split like this — a setting whose halves live in two keys where the copy takes one and not the other?

---

prestomation:
Both verified against the source, and one of them lands harder than the review found.

**Clearing the field.** The asymmetry does not exist, for a stronger reason than the panel lacking a control. `normalize_completion_required_fields` ends:

```python
return result or ["note"]
```

So an empty list is **not a state a `required` task can hold** — the backend refuses it and substitutes `["note"]`. And a `none`/`optional` task is forced empty there whatever the payload says. Skipping an empty array cannot suppress a meaningful value, because no meaningful empty value exists. That makes the guard purely about keeping an ordinary edit's payload byte-identical, which is what I have now written into the comment (`bd2ca73`) — it previously read as a convenience, which is exactly why it invited the question.

**The pill/Profile boundary.** Confirmed from source rather than by inspection of the render path: the pills are built under `onTasks && !profile`, with the comment beside them saying so — *"Only meaningful without a Profile, which is also the only time these pills are shown."* So a Profile does not merely win over the inline filter; the pills are not rendered at all while one is active. There is no view that asks both questions, which is the condition the split depends on.

One correction to the reasoning, not the verdict: the line cited for the profile-replaces-filter behaviour is in `panel-controls.ts`, not `panel-lists.ts`. Same conclusion.

**On the card config change.** The question I actually wanted pressed went unanswered, so recording it for the maintainer rather than treating it as settled: this changes what an existing dashboard card configured `filter: overdue` displays — buy reminders move to `filter: shopping`. I have shipped it as a plain fix on the grounds that the card and the panel pill must not disagree about one word, and the changelog says so in the bullet. It is a behaviour change to saved YAML, so it is a reasonable thing to want gated instead; that call is @prestomation's, and reverting it is one line in `matchesFilter`.

---

PR #294, opened 2026-09-05
Make the Test notification button deliver in every profile state

prestomation:
Fixes #

## What changed

The **Test** button on a saved notification called `home_keeper.notify`, and `notifier._send` returned early on an empty `due_queue`. So a notification *being configured* — before anything is overdue, which is exactly when a user sets one up — delivered nothing, and the target, channel, urgency and buttons could not be checked until real work came due.

Settings → Notifications with two rows. Walk my chores, whose profile holds tasks, shows Test beside an enabled Test all clear. Bedroom jobs, whose profile matches nothing, shows Test beside a greyed Test a task.

### The service

`home_keeper.notify` gains two optional fields that **compose** rather than override each other:

| field | values | meaning |
| --- | --- | --- |
| `status` | `overdue` \| `due_soon` \| `all` \| `none` | replaces the resolved profile's due-state filter for this call only |
| `when_empty` | `skip` (default) \| `all_clear` | what a queue that matched nothing does |

`status` is useful on its own — a weekly "everything on the list" digest needed a duplicate profile before. `status: none` matches nothing by design and is **service-only**: `profiles.normalize_filter` still clamps a stored `none` back to `overdue`, so no saved profile can be set to select nothing. `when_empty` defaults to `skip`, so the automatic triggers and every existing automation behave exactly as before.

### The panel

Test now sends `status: all, when_empty: all_clear`. Because `due_queue` already sorts ascending by `next_due`, the walk head is the most overdue task or — when nothing is late — the next one coming up. So Test delivers a real notification for a task whenever the profile holds one, and the "All caught up" notification when it holds none.

A second button beside it offers whichever notification Test does not send:

| profile state | Test delivers | second button |
| --- | --- | --- |
| holds a task | a task's notification | **Test all clear** → `status: none` |
| holds none | the all-clear | **Test a task**, greyed with a reason |

Detection is client-side via a new `profileHasAnyTask` in `card-filter.ts` (evaluated with the status forced to `all` — what the profile *saves* decides what it delivers on a schedule, not what a test can reach), so there is no extra round trip. The footer repaints from the form's own change handler, because a form edit saves with `render: false` and a later `set hass` only pushes into `_liveHassEls` — nothing else repaints the row.

### A wart the widening exposed

`notifications._overdue_phrase` answered `"Due soon."` for **anything** not yet due, including a task months out. That was reachable before (a saved `status: all` profile hits it) but rare; `status: all` on every Test press makes it the common case. It now says how far off the task is past the 3-day window, floored the same way the overdue side is.

### Where the logic lives

New branches went into `profiles.py`, `notifications.py` and `card-filter.ts` — all on the mutation allowlist — rather than `notifier.py` and `panel-settings.ts`, which are not scored. `notifier.py` gains no new branch of its own: it delegates to `profiles.with_status` and `notifications.sends_when_empty`.

## Testing

- `pytest tests/unit` — full suite green. New cases pin the `due_soon` → `due_in` boundary at exactly `DUE_SOON_WINDOW` and one second past it, the `status: none` rejection, and that a stored `none` still reads back as `overdue`.
- `npx vitest run` — green, including a case that drives a `profile_id` change through `value-changed` and asserts the second button flips **without** a re-render. That one protects the repaint design.
- `bash ci/e2e-up.sh` — all 79 e2e tests pass against a real HA container, including a new click of **Test all clear** asserting the all-clear toast.
- Screenshots recaptured on a clean container and inspected before committing.

## One-way doors

1. **`home_keeper.notify` field `status`**, vocabulary `overdue | due_soon | all | none` — a service input automations will read. Renaming it or narrowing the vocabulary later breaks them.
2. **`home_keeper.notify` field `when_empty`**, vocabulary `skip | all_clear`. **The default is the door**: `skip` preserves the promise that a send costs nothing on a quiet day. Flipping it later would make every silent schedule chatty.
3. **`none` is service-only and unstorable.** `normalize_filter` keeps clamping it to `overdue`. Allowing a stored `none` later would need a migration.
4. **The response contract stays `{matched, sent}`** — deliberately no `kind` key. Under `when_empty: all_clear`, callers read `matched == 0` as "the all-clear went out". Adding `kind` later is additive and safe; changing what `matched` means is not.
5. **`notification_strings` keys `due_in.{one,few,many,other}` and their `{days}` token** — a translation contract across 16 locale files.
6. **Delivered message text changes** for a task past the due-soon window: "Due soon." → "Due in {days} days." This is text on a user's phone. Nothing changes inside the 3-day window.

Not doors, but pinned test contracts worth naming: the `.hk-notify-test-alt` class, and the removal of the frontend locale key `notify.test_none`.

## Security

`notify` is not admin-gated, and `status: all` does not widen what a non-admin can reach. `docs/SECURITY.md` already grants non-admins task and profile reads, `targets` stay clamped to `mobile_app_*` / `persistent_notification` by `notifications.split_targets`, and a caller could already get `status: all` behaviour by naming a saved all-status profile.

## Noted, not fixed here

`profiles.py` and `card-filter.ts` both reference `notifications.is_walkable` in a docstring. No such function exists — the real one is `actions_for`. Comment-only, so left alone rather than widening this PR.

## Checklist

- [x] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` already frames
      `.hk-item-actions` for `BEAT * 3`, so the new button appears in the tour without
      an edit — confirming on the regenerated preview comment
      `const.py` → `0.21.0b2`), `preview-release` label applied

---

https://claude.ai/code/session_01N6fznLVS2n45SHvnnnZm1p


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first — but don't withhold the minor ones. This adds two composing per-call overrides to `home_keeper.notify` (`status`, including a service-only `none`, and `when_empty`) and a second contextual button in the Settings → Notifications footer.

Things I most want scrutinised:

**Correctness**
- `notifications._overdue_phrase`: the boundary between "Due soon." and the new "Due in {days} days." is `next_due - now <= DUE_SOON_WINDOW`. Does that agree exactly with `recurrence.is_due_soon` / `profiles.matches_filter`, so a task a `due_soon` profile queues always reads "Due soon."? Both operands are timezone-aware, so `.days` is elapsed time, not calendar days — is the DST behaviour acceptable, and is it consistent with the existing overdue branch?
- The `status: none` boundary. `profiles.STATUSES` stays three-valued so `normalize_filter` clamps a stored `none` back to `overdue`, and the override is applied *after* `normalize_profile` in `async_run_notify`. Is there any path where `none` could reach storage, or where the override could leak into the live options dict that `resolve_profile` returns?
- `notifier._send`: I moved the empty-queue early return and made the no-target warning fire only for a non-empty queue. Did I change behaviour for the automatic (`async_send_auto`) or walk-advance paths, which still pass the `skip` default?
- `_build_payload` now builds the all-clear for an empty queue. Is the executor hand-off still the single site `tests/unit/test_notifier_blocking.py` relies on?

**Maintainability**
- `itemCard` in `panel-settings.ts` grew a `secondary` option and a third `repaint` argument to `fill`, and its footer is now created before `fill` but appended after. Is that a reasonable shape, or is the row abstraction being overloaded?
- The `secondary.host` field exists only so an unavailable press can toast. Better alternatives?

**Performance**
- `profileHasAnyTask` runs over `p._tasks` on every repaint (each form change). Is that a concern at a realistic task count?

**HA best practices**
- `vol.In` on the two new service fields rather than `cv.string` plus a clamp in the pure module. The rest of this integration uses the latter — is the stricter validation right here, or is the inconsistency worse?
- The response shape stays `{matched, sent}` with no `kind` key: callers read `matched == 0` under `when_empty: all_clear` as "the all-clear went out". Is that too subtle a contract?

**Security**
- `notify` is not admin-gated. Does `status: all` widen what an unprivileged caller can reach in any way I've missed?

---

prestomation:
/q review Re-requesting after a push that only removed an AI-tell transition from the README. The substantive review request above still stands — please work from that one; this is just so the review runs against the current head.

---

prestomation:
### e2e is red, and it is not this PR's

**Failing check:** `e2e` — one spec, `tests/card-registration.spec.ts:54` *"renders from an app shell that never imported its bundle"*. Everything else passes (225 passed), including this PR's own new `settings.spec.ts` assertions for the second Test button and the all-clear delivery.

**Why it is not this PR's.** The same spec fails identically on **#287** (`claude/feature-ideas-r55rrb`), which touches declarative companions and nothing about card delivery — same assertion, same received string, same three Playwright retries. #287's e2e was green at 07:26 today and red at 19:27; this PR has been red on the same spec since its first commit. `main`'s last e2e run was green at 04:56, before the window. Nothing in this diff touches `card.ts`, `card-index.ts`, `card_resource.py`, or card delivery.

**Root cause.** Home Assistant's `index.html.template` now wraps each `extra_modules` import in an error handler:

```js
import("/home_keeper_panel/home-keeper-card.js?v=ed5bb928e240").catch(function (err) {
  console.error("Failed to load extra module /home_keeper_panel/home-keeper-card.js?v=ed5bb928e240", err);
});
```

`CARD_IMPORT` (line 51) matches only the bare `import(…)` call, so the strip removes the import and leaves the `.catch(…)` behind. That trips the self-check at line 114 because the filename survives inside the `console.error` message — and it also leaves the shell holding a dangling `.catch(…)` with no receiver, which is a syntax error. The container tracks `HA_TAG=stable`, so this arrived with an upstream image, not with a commit here.

**No fix exists to port** — #287 is red on it too — so per AGENTS.md I am not widening this PR with a shared test-infra change. Proposed patch, verified against the exact shell text CI reported (strips the whole statement, leaves other panels' imports untouched, leaves no reference behind):

```diff
-const CARD_IMPORT = /import\(\s*(["'])[^"']*home-keeper-card\.js[^"']*\1\s*\)\s*;?/g;
+// HA wraps each extra-module import in a `.catch` whose message repeats the filename,
+// so stripping the bare `import(…)` leaves both a reference and a dangling handler.
+const CARD_IMPORT =
+  /import\(\s*(["'])[^"']*home-keeper-card\.js[^"']*\1\s*\)(?:\s*\.catch\(\s*function\s*\([^)]*\)\s*\{[\s\S]*?\}\s*\))?\s*;?/g;
```

Happy to send that as its own PR so #287 and `main` get it too — say the word and I will.

**Re-run:** not spent. Playwright already retried the spec twice per run, and an unrelated PR reproducing it identically is stronger evidence than a third attempt on the same commit would be. I will keep this PR watched until it is green.

---

prestomation:
/q review Two things changed since your last pass, so please re-review the current head rather than the diff you saw.

**Taken from your review:** you asked for the response contract to be written down rather than left in the PR body. `async_run_notify`'s docstring now says it — under `when_empty: all_clear` something is always delivered, so `matched` says *which* card went out rather than whether one did, and under the default it keeps its plain meaning.

On your other suggestion — cross-referencing the two isolation points — I believe both already do: `normalize_filter`'s docstring names `home_keeper.notify` as the only place `none` is legal, and `async_run_notify` explains why the override must land *after* `normalize_profile`. Tell me if you meant something more explicit.

**New since then:** 16 locale files for `notification_strings` (`due_in.*`) and the panel, plus `services.yaml` / `strings.json` / `translations/*.json` for the two new fields. Worth a skeptical look at:

- Whether the `services.yaml` `select` options and the `vol.In` vocabularies actually agree, in both directions, for `status` and `when_empty`.
- The `_COGNATE_IDENTICAL` addition in `tests/unit/test_translations_parity.py` — is allowlisting "Status" in seven locales right, or is it papering over a translation that should exist?
- `tests/unit/test_notifications.py` gained three assertions to kill surviving mutants, one of which pins that `build_notification` threads `window` through to `_overdue_phrase`. Is that test proving the thing it claims, or just exercising a parameter nobody passes?

Please keep the earlier correctness questions in scope too — particularly the `_overdue_phrase` boundary and DST, and whether `status: none` can reach storage by any path.

---

prestomation:
### Folded the strip fix in — `f01ad08`

Following up on my stand-down comment above: on the maintainer's call this landed here rather than as its own PR, since it blocks this branch and there was no fix to port.

The patch is the one proposed above, and I was able to verify it properly rather than by inspection. Two corrections to what I said earlier are worth recording:

**The CI browser *is* available in this environment.** `ci/setup-browser-env.sh` runs `npx playwright install --with-deps chromium`, which seeds both `chromium` and `chromium_headless_shell` into `/opt/pw-browsers`. With `CHROMIUM_EXEC` unset, Playwright resolves the headless shell — the same browser `e2e.yml` drives. AGENTS.md currently says that install fails here and steers to `CHROMIUM_EXEC`, which for this spec is the *wrong* browser; that guidance looks stale.

**So this is a reproduction, not a guess.** Same headless shell, same `HA_TAG=stable` container:

| `CARD_IMPORT` | result |
| --- | --- |
| old pattern | ✘ failed ×3 — *"the strip left a reference behind"*, identical to CI |
| new pattern | ✓ passed |

Also checked, against the exact shell text: other panels' imports survive the strip, no dangling `.catch` is left, and the older single-statement template shape still strips — so this works whichever template the image ships.

It is the strip that was incomplete, not the assertion that was too strict, so nothing is skipped or relaxed and the guard at line 114 still does its job.

**#287 is still red on this spec** and does not get the fix from here until this merges — that is the accepted cost of folding it in rather than opening a separate PR.

---

PR #298, opened 2026-09-06
fix(panel): hide Done on a monitored task, and open the recipe that built it

### LINKED ISSUE #231: [New feature] Declarative Companions

prestomation:
Fixes #231

## What changed

Two defects reported on #231 by a beta tester running a Device Pulse recipe, both on the page of a task a declarative companion materialized.

**A monitored task offered Done.** `statusBucket` puts every dormant `triggered` *and* `sensor` task in the Monitored section, but the three surfaces that draw a Done button each re-spelled that rule for themselves, and all three tested only `triggered`. So every dormant edge-mode sensor task (state / threshold / availability, which is every declarative-companion task until its condition fires) carried a Done that recorded a completion and moved nothing, because `next_due_after_completion` leaves a sensor task dormant. The rule is now one predicate, `utils.isMonitoredDormant`, read by the task page, the task list and the dashboard card. A dormant **usage** meter deliberately keeps its Done: it is counting to a target, and an early completion re-anchors the baseline through `store._reset_usage_baseline`. The task page's Next due row now reads "Monitored" there too, rather than contradicting its own status chip with "No due date".

**"Edit in Device Pulse" opened the Home Keeper integration page.** `declarative_companions.build_managed_by` stamps Home Keeper's own `config_entry_id` on a materialized task and puts the *recipe's* name in `display_name`. The deep link therefore resolved to the `home_keeper` domain while its label named an integration that is not installed, and the sibling caption read "Delete from Device Pulse instead", naming a place that does not exist. Such a task now offers **Edit recipe**, which opens the recipe in the same dialog Settings uses, and the caption names the recipe to change. Home Keeper's own domain is no longer a valid target for the generic deep link at all.

### Screenshots

Recaptured after rebasing onto #299, so they show the task page with its new Schedule / Notes / History sub-tabs: Monitored, no Done, and Edit recipe beside Edit and Duplicate.

Task page showing a Monitored chip, no Done button, and an Edit recipe button beside Edit and Duplicate

Edit recipe opens the recipe over the task page.

The recipe editor dialog open over the task page with the live preview showing one match

Both shots come from a new step in `tests/e2e/screenshots-declarative.capture.ts`, which adds the recipe over the service, asserts the three things the fix changed before it photographs anything, then deletes the recipe so the container is left as it was found.

### Tests

- `isMonitoredDormant` pinned directly in `utils.test.js` (every mode, armed and dormant, and a stale `sensor` block on a non-sensor task).
- New `declarative-task-detail.test.js` covers the task page and the task list row. Five of its seven assertions fail against the previous code.
- `card.test.js` covers the dashboard card row.
- After the rebase: 1198 frontend tests, 1922 pytest unit, and the full e2e suite (273) all green on the CI browser with `CHROMIUM_EXEC` unset.
- Stryker on the changed surface: 100% (21/21), threshold 80%.

### One-way doors

None. Nothing here changes a service input, an event payload, storage, or an entity attribute. `managed_by` keeps the exact shape `build_managed_by` already writes, so a task materialized by an older beta needs no migration. The change is confined to what the panel and the card draw, plus two new frontend locale keys (`btn.editRecipe`, `managed.deleteFromRecipe`) in all 16 languages.

### Release

Bumped to `0.21.0b6` (`manifest.json` + `const.py`) with a matching CHANGELOG section. #299 took `0.21.0b5` and it is already a published tag, so this branch opens the next beta above it rather than adding to a section that has shipped.

### Also here, and separable

One line of test config, unrelated to #231: the walkthrough capture's per-test timeout. It is in this PR because this PR's walkthrough comment was carrying the failure. The desktop tour measures 168s on CI and 168-174s locally, and the old 180s cap left under ten seconds of margin, so it timed out on all three attempts. Now 240s, fitted to that measurement. Happy to split it out if you would rather.

## Checklist

      step through it — not applicable, this is a bug fix on an existing surface
- [ ] New user-facing feature → version bumped to the next beta (`manifest.json` +
      `const.py`), `preview-release` label applied — version bumped (b5 is taken and
      shipped), no `preview-release` label since this adds no feature

---

https://claude.ai/code/session_013ZSoB1oGgmH5R6qDj6Va4f


prestomation:
/q review Please be critical and skeptical, and rank the most serious findings first without withholding the minor ones.

This is a panel-only bug fix for #231. Two changes worth attacking:

**1. `utils.isMonitoredDormant` (custom_components/home_keeper/frontend/src/utils.ts)** is now the single rule three surfaces read to decide whether a task gets a Done button: `panel-detail.ts`, `panel-lists.ts` and `card.ts`. It answers true for a dormant `triggered` task and for a dormant `sensor` task in an edge mode (`state` / `threshold` / `availability`), and false for a dormant `usage` meter.

- **Correctness:** is the edge-mode list the right boundary? It is meant to mirror `sensor_tasks.holds_edge_state` on the backend — check that it actually does, and that a mode added to one side cannot silently diverge from the other. Is hiding Done on a dormant `threshold` task right, given a user may genuinely service the thing before the reading crosses, and completion there records history without moving `next_due`? Does an absent `sensor.mode` behave the way `models.normalize_sensor` says it should?
- **Consistency:** `card-filter.statusBucket` buckets a dormant *usage* task as `monitored` while this predicate says it is not monitored. Is that divergence defensible, or does one surface now contradict another (a Monitored heading over a row that still has Done)?
- **Reachability:** are there other surfaces that draw a completion affordance and still carry the old `recurrence_type === 'triggered'` test? The `button.py` mark-done entity and the to-do list were left alone on purpose — argue whether that is right.

**2. The "Edit recipe" path (`panel-detail.ts` + `panel-declarative.ts`)** replaces a deep link that pointed at Home Keeper itself.

- **Correctness:** `declarativeRecipeFor` looks the spec up in `p._declarativeCompanions`. What happens on the window before that list loads, or when the websocket call soft-fails to `[]`? The page then falls back to the generic managed captions — is that the right failure mode, or does it flicker?
- **Maintainability:** `panel-detail.ts` now imports from `panel-declarative.ts`. Check the region-module boundaries in `.amazonq/rules/architecture-and-code.md` → "The panel is one element plus flat region modules" and say whether this import runs the wrong way or risks a cycle.
- **HA / panel conventions:** the recipe dialog is opened over a task page rather than navigating to Settings. Does that break the panel's deep-linking rule (every destination maps to a URL under `/home-keeper`)? Back/Forward behaviour with the dialog open is worth a hard look.
- **Security:** `data-spec-id` is written into markup via `escapeHTML` and read back through `dataset`. Confirm nothing user-controlled reaches the DOM unescaped here, and that the recipe name in `managed.deleteFromRecipe` is escaped on the way in.

**3. Everything else:** the two new locale keys across 16 files, the `Stryker disable next-line StringLiteral` annotation on the mode fallback (is that mutant genuinely equivalent?), and the new e2e capture step, which creates a recipe over the service and deletes it in a `finally`. Does that leave the container clean if the assertions throw partway?

---

prestomation:
Standing down on the one red check, because it is not this PR's and the fix is not in this repository.

**Failing check:** Discord Notify / notify, [run 34042352228](https://github.com/prestomation/ha-home-keeper/actions/runs/34042352228), on the superseded head 918dff7. It fired on the pull_request_review event when Amazon Q posted its review, not on any push.

**Root cause — a shell injection in the shared reusable workflow.** notify-discord.yml here is four lines that call prestomation/.github/.github/workflows/notify-discord.yml@main. That reusable workflow builds its Discord payload with an *unquoted* heredoc:

    PAYLOAD=$(cat <<ENDJSON
    ...
        "description": "${BODY_SNIPPET}",
    ...
    ENDJSON
    )

An unquoted heredoc delimiter makes bash expand the body, so backticks inside the interpolated comment text are run as a command. Amazon Q's review quoted a symbol name in backticks, and the job died with:

    line 128: isMonitoredDormant: command not found
    ##[error]Process completed with exit code 127

**Why it is not this PR's.** The diff does not touch notify-discord.yml, and the trigger is comment prose rather than code. The same job is already red on main for the same reason, twice, both from ordinary comments: [run 34039794767](https://github.com/prestomation/ha-home-keeper/actions/runs/34039794767) (a comment on #231) and [run 34014448814](https://github.com/prestomation/ha-home-keeper/actions/runs/34014448814) (an Amazon Q comment). No fix for it exists anywhere I can port from, and the file that needs changing lives in prestomation/.github, outside this repository — so there is nothing to push here. I have not re-run the job: the failure is deterministic in the review body, so a re-run would only fail again.

**Worth treating as a security bug, not a flake.** Anyone who can comment on an issue or PR here controls BODY_SNIPPET, so anyone who can comment can run arbitrary commands on the runner, in a step whose environment holds DISCORD_WEBHOOK_URL. Two changes in prestomation/.github would close it:

1. Quote the heredoc delimiter, which stops every expansion in the body:

       PAYLOAD=$(cat <<'ENDJSON'

   The values then have to reach the payload some other way, which is what 2 does.

2. Build the JSON with jq rather than string interpolation, passing each untrusted value as an argument so it is escaped as data:

       PAYLOAD=$(jq -n \
         --arg title "$EMBED_TITLE" \
         --arg url "$OBJ_URL" \
         --arg description "$BODY_SNIPPET" \
         --arg author "$AUTHOR in $READABLE_REPO" \
         ...
         '{username: "GitHub Notify", suppress_embeds: true, embeds: [{title: $title, url: $url, description: $description, ...}]}')

   This also fixes a second latent bug: a comment containing a double quote or a newline currently produces invalid JSON and Discord rejects the payload.

Until that lands, expect this check to go red whenever a comment on this PR quotes anything in backticks — including this one, which had to describe the failure without using any.

Everything else on the current head (d409384) is green or still running: ruff, mypy, vale, hacs and the frontend checks have passed, and the rest are in progress. I will keep watching and act on anything that turns red.

---

prestomation:
/q review Please be critical and skeptical about the latest push (e5347cd), which is one line of test config: the walkthrough capture's per-test timeout goes from 180s to 300s.

Attack the reasoning, not just the diff:

- **Is raising the budget the right fix, or does it hide a regression?** The desktop tour timed out on all three attempts at 180s, while measuring 168-174s locally on the same commit and passing on both the full Chromium and the headless shell CI drives. My read is that the tour has always been within ten seconds of its budget and a slower runner spends that margin. Argue the alternative: that something in this branch made the tour slower. Nothing here touches walkthrough.capture.ts, and the tour's only Done interaction is on the scan-locked furnace filter, a floating task that isMonitoredDormant does not affect — check whether that reasoning holds, and whether any panel change on this branch could slow a step the tour walks through.
- **Is 300s the right number,** or does a budget that loose stop the timeout from catching a genuinely hung tour? Is a per-step assertion timeout the better guard than the whole-test one, given actionTimeout is already capped at 20s?
- **Does this belong in this PR at all,** or should it be split out as its own change? It is unrelated to the #231 fix and I pushed it because this PR's own walkthrough comment was carrying the failure.

Rank the most serious concerns first and do not withhold minor ones.

---

prestomation:
Taking point 5, pushing back on points 1 and 3, because both rest on a measurement that does not exist.

**The CI figure in the review is not real.** The recommendation to use 240s is justified by "CI measured 175-182s across three attempts before this change." Nothing measured that. All three attempts were killed at the cap and reported `Test timeout of 180000ms exceeded` — a timed-out test reports its budget, not its own length. Those runs establish exactly one thing: the tour needs more than 180s on CI. They cannot distinguish 181s from 400s, so "175-182s" is below the one bound the data gives, and cannot be written into the comment as a measurement.

That is also the answer to point 1, "have you measured the tour's actual CI runtime?" — no, and it is not measurable from a timed-out run. It becomes measurable on the first green one, which is what 300s is for.

**So the number is deliberately loose, once.** 240s is a bet that CI is at most ~38% slower than this container. I have no evidence for that bound. If the bet is wrong the tour times out again, the capture comment stays broken, and it costs another push and another review round to land the number we could have had by reading it off a green run. 300s makes the next run print the real duration, and the comment now says the budget is meant to be tightened against that figure rather than fitted to a guess. Happy to do that tightening as a follow-up once there is a number.

**On the timeout losing its purpose:** agreed in general, but this cap was never the guard against a hung tour. `actionTimeout: 20_000` is, and it fails a bad selector in twenty seconds — the review makes that point itself under 3. What the per-test cap has to be is clear of the tour's own length. The comment now says both things.

**Point 5 taken**, with the honest version of the measurement rather than the invented one: 63059df records what the three attempts do and do not establish, and why the number is what it is.

**Point 2 confirmed independently** — thank you for checking it rather than taking my word. That matches what I found: `isMonitoredDormant` only removes elements, and the tour's one Done interaction is the scan-locked furnace filter, a floating task it does not touch.

**Point 4 is the maintainer's call,** not mine to settle. The timeout change is unrelated to #231 and is here only because this PR's walkthrough comment was carrying the failure. Happy to split it if @prestomation would rather.

---

prestomation:
There is now a real number, so I have taken the 240s recommendation. Pushed as f07daa0.

**Measured on the first green run:** the desktop tour takes **168s on CI**, against 168-174s in the dev container. The walkthrough job is green and the sticky comment above is embedding the gif again.

That number corrects my own reasoning as much as it settles the timeout. I said a slower CI runner was spending the margin. It is not — CI runs the tour at the same speed as here. The tour simply sits on the line at a 180s cap, and ordinary run-to-run variance is what carries it over. The comment in the config now says so, because the wrong version of that story would send the next person hunting for CI-specific slowness that does not exist.

240s is 168s plus ~40%, which answers the concern that 300s was loose enough to stop catching cumulative drift. The earlier 300s was never meant to stand: it was the value that let a run report its own length instead of dying at the cap, which is the only way to get the measurement that this commit is fitted to.

Still open for @prestomation rather than me: whether the timeout change rides along here or gets split out (your point 4). It is unrelated to #231 and is in this PR only because this PR's walkthrough comment was carrying the failure.

Everything else is green on the current head: ruff, mypy, vale, hacs, changelog-release-gap, pytest and coverage, test, e2e, integration, release, preview, walkthrough, and Mutation via its Python and TypeScript jobs.

---

prestomation:
/q review Please be critical and skeptical about f07daa0, and rank the most serious findings first without withholding minor ones.

It is one line of test config plus its comment: the walkthrough per-test timeout goes 300s to 240s, fitted to a measurement that did not exist when you last reviewed it. The desktop tour reports 168s on CI and 168-174s in the dev container.

What to attack:

- **Is 240s actually enough?** It is 168s plus ~40%. The same tour exceeded 180s at least once, so the run-to-run variance is real but its size is unmeasured — one green run and three timeouts is a thin sample. Argue whether a single measurement justifies this bound, and whether the failure mode of getting it wrong (another broken capture comment, another push) is worse than leaving it loose.
- **Is the causal story right?** I now claim CI is not slower than the dev container and that variance alone carries the tour over a 180s cap. The evidence is one CI figure matching the local range. Attack that inference — what else explains three consecutive timeouts followed by a comfortable pass, and does anything in the walkthrough job itself (container warm-up, the ffmpeg apt install, the browser download) contend for wall clock with the Playwright step in a way I have not accounted for?
- **Does the comment mislead?** It tells a future reader not to hunt for CI-specific slowness. If that conclusion is wrong, the comment is worse than no comment. Check it against the diff and say so.
- **Correctness of the surrounding config** while you are here: `actionTimeout: 20_000` is claimed as the real guard against a hung tour. Is that true given the tour's use of `waitForTimeout`, which no action timeout bounds?

The #231 fix itself is unchanged since your earlier pass; no need to re-review it unless something in this diff touches it.

---

prestomation:
Finding 1 is right and it was my error. Fixed in 32fb176, along with finding 4. Two of the others do not hold.

**1 — `actionTimeout` is not the hang guard. Correct, and it was the load-bearing claim.** `actionTimeout` bounds Playwright *actions*; the tour is mostly `page.waitForTimeout(BEAT)`, which no action timeout touches. So the per-test budget is the only thing that stops a hung tour. That is worse than a stray wrong sentence, because it was the reason I gave for why a loose cap was harmless — and it argues *for* the tighter number, not against it. The comment now says the cap is the only bound, and that what `actionTimeout` actually buys is a bad selector failing in 20s at the step that broke, instead of surfacing as a timeout on the whole tour.

**4 — the comment discouraged investigation. Taken.** It now says one CI figure inside the local range is one sample against three timeouts: enough to say the tour sits on the line at 180s, not enough to prove CI never runs slower. It tells the next reader to suspect the margin first and go looking for CI slowness only if a green run ever reports a length the container cannot reproduce.

**2 — "potentially blocking legitimate PRs in CI" is not the failure mode.** The walkthrough is a soft gate by construction: the capture step is `continue-on-error`, the job posts a "capture failed" note and exits 0, and its check run reports **success**. That is exactly what happened on the run that started this — three timeouts, sticky comment showing the warning, check run green. The cost of getting the number wrong is a broken comment and a push, not a blocked PR.

**Also 2 — `retries` is already there, at 2, not the 1 you suggest.** `playwright.config.ts:81` sets `retries: process.env.CI ? 2 : 0`, and `captureConfig` spreads the base config, so the walkthrough inherits it. That is why the failing run shows "retry #1" and "retry #2". The tour therefore gets **three independent attempts at 240s**, which is the part of the risk calculation the recommendation was missing: all three would have to exceed 240s against a measured 168s. The comment now records that.

**3 — fair on sample size, and I have left it as a known limit rather than papering over it.** Instrumenting the job's setup costs to separate them from test wall clock is a reasonable follow-up, but it is not this PR's to carry: the Playwright reporter already times the test independently of the `apt-get ffmpeg` and browser-download steps that precede it, which is where the 168s comes from.

---

prestomation:
/q review Rebased onto main (now `f8269f1`, which carries #299). Please be critical and skeptical, and rank the most serious findings first.

The reason to look hard: **#299 restructured the same surfaces this PR touches, and git merged them without a single conflict in code.** Only `CHANGELOG.md` conflicted. A clean textual merge is exactly the case where a semantic break hides, so treat "it merged cleanly" as no evidence at all.

Specifically:

- **The task page now has Schedule / Notes / History sub-tabs with their own URLs (#299), and this PR changes what the task page's action row renders.** Check that `isMonitoredDormant` still governs the right button on the right surface, that Done and Edit recipe sit in the action row rather than inside a tab panel, and that a deep link to a sub-tab does not bypass or duplicate the logic. `panel-detail.ts` is where the two changes overlap most.
- **#299 also compacted the task rows and made the phone layout a two-column grid.** `panel-lists.ts` is the other overlap: confirm the Done suppression survived that restructuring and that the compact row does not render a Done the full row would have hidden, or vice versa.
- **`utils.ts` took additions from both sides** (`partStockStep`, `snapStock`, `TASK_TABS` from #299; `isMonitoredDormant`, `HK_DOMAIN` from here). Check nothing was silently dropped or duplicated by the merge.
- **The CHANGELOG conflict resolution.** #299 shipped as `0.21.0b5` and that tag is published, so I opened `0.21.0b6` above it rather than adding to a shipped section. Confirm that is right, that #299's b5 content is intact and unedited, and that `manifest.json` and `const.py` agree with the heading.
- **The screenshots were recaptured** against the rebased layout, so they now show the sub-tab bar. Check they depict the state the diff actually produces.

Everything is green locally after the rebase: 1198 frontend, 1922 pytest unit, 273 e2e on the CI browser, Stryker 100% (21/21). That is what I would expect to see whether or not the merge is semantically sound, so please reason from the code rather than from those numbers.

---

prestomation:
/q review Two new commits since your last pass, f080058 and 18e2ea6, fixing two follow-ups the reporter found while running the Firmware update available preset. Please be critical and skeptical, rank the most serious findings first, and do not withhold minor ones.

**1. `completion_blocked` is now derived, not hardcoded.** `declarative_companions.build_managed_by` reads the recipe trigger's `clear_on_recover` straight into `completion_blocked`, and adds a localized `completion_prompt` when it is set.

- **Correctness of the premise.** I claim that when `clear_on_recover` is set, a hand-pressed Done actively hides live work, because `sensor_tasks._evaluate_edge` will not re-arm while the condition merely stays true — so completing "Update firmware" dismisses an update that is still pending until that update is installed and a different one appears. Verify that against `_evaluate_edge` and `recurrence.next_due_after_completion` rather than taking my word. If the task *would* re-arm, the whole change is wrong.
- **Is `clear_on_recover` the right discriminator,** or does it miss a case? My reasoning: without it, Done is the only thing that ever clears the task, so blocking it would strand the task forever; a `usage` meter drops the key entirely and completing early re-anchors the baseline. Attack both halves.
- **Blast radius.** `completion_blocked` is read by the panel, the card, `todo_list.py` and `notifications.is_completion_blocked`. Check that flipping it to True for these tasks does not break the to-do sync or the notification builder in a way I have not tested, and that `defer.ts` withdrawing skip alongside it is right.
- **`lang` threading.** `build_managed_by` gained a `lang` kwarg, threaded through `_build_task` and `reconcile_declarative_tasks` from `store.reconcile_declarative_companion_tasks` (`self._hass.config.language`). Check the default does not silently English-ify a non-English install through some path I missed, and that a language change restamps.
- **Migration.** I claim existing materialized tasks pick the new flag up because the reconciler recomputes `managed_by` each pass and updates on difference. Confirm that from `reconcile_declarative_tasks`.

**2. A declarative task is now source-owned.** `sourceOwnedTask` gained `source.declarative_companion`, so Edit, Delete and Duplicate go away and **Edit recipe** moved out of the `!sourceOwned` branch to survive.

- Check nothing else keys off `sourceOwnedTask` in a way this breaks — `_canDuplicate` and the drawer opener are the two I looked at.
- The inline notes editor deliberately stays live (notes are not in `_LOCKED_FIELDS`). Is that consistent with how a problem-sensor task behaves?
- The source-owned caption now names the recipe rather than using `managed.sourceOwned`. Confirm the fallback still applies to wear parts and problem sensors.

**3. Anything I got wrong in the docs.** The README now claims a recipe's task has no due date until its condition is true, that arming stamps `next_due` to that moment, and that all bundled presets auto-clear. Check each against the code.

`complete_task` stays permissive on purpose — problem sensors reject completion in the store as well, but a recipe's task is real work someone may want to log, so hard-blocking the service changes a documented contract. Argue if you think that is wrong.

Local: 1926 pytest unit, 1201 frontend, Stryker 100% (21/21), ruff and vale clean. mypy could not be installed in my container (the Home Assistant dependency fails to build PyRIC), so CI is its first run — please look at the type changes yourself rather than assuming they were checked.

---

prestomation:
/q review Please be critical and skeptical. The last commit fixes the `integration` job, which was red on three heads: `test_a_state_recipe_materializes_one_managed_task_per_matching_entity` still asserted `managed_by["completion_blocked"] is False`, but this PR makes the declarative reconciler derive that flag from the recipe trigger `clear_on_recover`, and the water-tank recipe sets it. The assertion now reads True and checks the prompt names the recipe; the device-backed threshold recipe, which does not clear on recover, gets the opposite assertion in the test that already builds it.

Look hardest at **correctness**: is deriving `completion_blocked` from `clear_on_recover` the right rule, or does it block Done on a task that nothing else will ever close? Trace every reader of the flag (`todo_list.py`, `notifications.is_completion_blocked`, the panel and the card) and say whether any of them now behave wrongly for a declarative task. Check that `complete_task` and the websocket twin agree with what the UI shows, since the flag is advisory and the service is not gated.

Also review **maintainability** (is `build_managed_by` the right home for the `lang` plumbing, and is `resolve_string` called on a path where the language is always known), **HA best practices** for the reconcile-and-reload path, and **test quality** — whether the new integration assertions actually cover the branch or just restate the implementation. Most serious issues first, and do not withhold the minor ones.

---

prestomation:
Triage of the review above. CI is green on `ae9a7a2`: integration, mypy, e2e, walkthrough, both mutation jobs, vale, ruff and hacs all pass. The integration job was red on three heads because `test_a_state_recipe_materializes_one_managed_task_per_matching_entity` still asserted the pre-fix `managed_by` shape; `ae9a7a2` corrects it and adds the opposite assertion on the device-backed threshold recipe.

**On the locale suggestion — already enforced, and more strictly than described.** The review points at the runtime check in `store.py`, which only warns. The real gate is `tests/unit/test_backend_strings_parity.py`, which fails the build on a missing key, a dropped `{name}` token, an unbalanced brace, an invalid placeholder identifier, or a value left identical to English. A malformed `declarative_task.completion_prompt` cannot reach a release. (The bundle is 16 locales, not 20.)

**On the completion gate — this is incorrect, and the gap is real.** The review states that `_reject_synced_problem` prevents manual completion of declarative tasks "since they're checked via `task.managed_by?.completion_blocked`". It does not. That gate reads `_problem_source(task)`, which is set only by the problem-sensor sync. `complete_task` calls exactly two gates, `_reject_synced_problem` and `_reject_scan_required`, and the string `completion_blocked` does not occur anywhere in `store.py`.

So `completion_blocked` is advisory. The panel, the card, `todo_list.py` and `notifications.is_completion_blocked` all honour it, but a direct `home_keeper.complete_task` call — from an automation, a script or voice — still completes the task. For an auto-clearing recipe that matters: the watcher will not re-arm while the condition still holds, so the completion writes a history row and drops a pending firmware update off the list.

This predates the PR (before it, the panel offered the same Done and did the same damage), and closing it changes a documented service contract, so it is not folded in here. The fix would be a third origin-gated rejection beside the two above, following the same pattern. Flagging it for the maintainer rather than deciding it.

---

PR #301, opened 2026-09-06
feat(panel): point a task's device chip at its appliance, and lay both lists on fixed rails

prestomation:
Fixes #

## What changed

Two changes to the panel's list surfaces.

### 1. Where a device chip goes

From a **task**, the chip means "the appliance this work is about", so it now opens that appliance's Home Keeper page. A Home Assistant device page knows nothing about Home Keeper and was never the useful answer there. From an **appliance**, the chip still means "the Home Assistant device behind this" and keeps the device page — so the device is always exactly one hop from a task, and the two directions never disagree.

| Surface | Destination |
| --- | --- |
| Task list row | **Home Keeper appliance page** (changed) |
| Task detail page | **Home Keeper appliance page** (changed) |
| Appliance list row | HA device page (unchanged) |
| Appliance detail page | HA device page (unchanged) |

When no appliance claims a task's device, the chip falls back to the device page and carries a small leaves-the-panel mark, so the destination reads before the click rather than after it.

`assetForTask` in `utils.ts` is the single ranking that answers "which appliance does this task belong to" — its part link first, then the appliance that owns the device, then one that merely lists it as related, with a live appliance always ahead of an archived one at the same strength. `taskRelatesToAsset` is now that predicate's boolean face, and the task form's consumable and document pickers read the same ranking, so the chip and the form cannot drift apart.

`wireDeviceChips` now takes the panel, only to reach `_openDetail` — that is what keeps an in-panel hop on the panel's own history, so Back returns to the list rather than ejecting from the panel.

### 2. Both lists lay their parts on fixed rails

A wide row is a grid of fixed tracks instead of a shrink-to-fit flex line. The chip cluster used to begin wherever the task name happened to end (**121px of drift** measured over the seeded list), and the status pill was right-anchored but variable width, so its left edge wandered too.

The tracks are anchored from the **left**, and that is load-bearing. A row's action column is sized by what is in it — Done and a caret on most rows, a caret alone where there is nothing to mark done, a caption where the task clears itself — a 178px spread. Packing the tracks against the right moves every rail with that spread, which is the ragged list again by another route.

Other pieces of the same change:

- The appliance row grew a status column of its own and takes the same rails, so the two lists stop reading as two different products. It buys no Done, so it spends that width on a wider chip track and wraps inside it rather than truncating.
- The whole appliance row still opens the appliance. Splitting it into tracks left the opener holding only the first one, so a stretched-link overlay puts it back; the chip and count tracks let their empty space fall through and lift only what is really a target.
- The overdue rail is now **reserved on every card** and only coloured on an overdue one. A border is layout: 2px of it put an overdue row's content on a different x from every other row's, which split the rails in two.
- The status pill is left-aligned in its track and set in `tabular-nums`. A pill is read by where it starts, and "8 days overdue" beside "128 days overdue" only ever agreed on where it ended.
- A row's meta line takes one line and ends in an ellipsis. Wrapping made row height depend on how long the recurrence sentence happened to be, which is the ragged list in the other axis.
- A chip too wide for its track fades out instead of being cut through.
- Below 1150px, and inside the 268px appliance picker pane, rows keep the stack they had. The rails need width.

Ideas artifact behind the design choice (five row grammars with live alignment guides): https://claude.ai/code/artifact/b1d3b546-8f29-4edf-86ce-48583aa57a7e

### Screenshots

Task list — chips, status pills and Done each on one x:

Home Keeper task list with every row's chip, status pill and Done button starting at the same horizontal position

The same list with a device chip on most rows:

Task list where five rows carry a Garage water heater device chip, all aligned on one column

Appliance list, now on the same grammar:

Home Keeper appliance list with name and details on the left, device chips in a fixed column, and sub-device counts in a status column

A single row, close up:

A single task row showing its name, meta line, device chip, overdue pill and Done button

Phone layout, unchanged:

Home Keeper task list on a phone, each row stacking its title, status and chips beside a Done button

86 screenshots were regenerated across every panel capture harness.

## Rebased onto the 0.21.0 release

`main` shipped **0.21.0** while this branch was open, so the merge resolves three version conflicts:

- The bump moves to **`0.22.0b1`**, not `0.21.0b7`. PEP 440 sorts `0.21.0bN` *below* the shipped `0.21.0`, so HACS would have offered the stable to a beta tester as an upgrade.
- The CHANGELOG section is renamed to match and sits above the shipped `## [0.21.0]`.
- `main` also tightened how a bullet opens — the bold lead is a heading, not a sentence. Both bullets now take a noun-phrase lead.

## One-way doors

**None.** This change commits no new external surface.

- No service input, event payload, storage shape or entity attribute changes.
- The URL the chip navigates to, `/home-keeper/appliances/<asset_id>`, already exists and is already minted by the backend as a device page's "Visit" link (`devices.py`). Nothing new is written into a device's `configuration_url`.
- `deviceChip`'s third argument and `wireDeviceChips`'s new parameter are internal panel signatures with no integrator-facing contract.
- The two new locale keys (`chip.device.appliance.tip`, `chip.device.tip`) are additive.
- The CSS custom properties (`--hk-name-col`, `--hk-chip-col`, `--hk-chip-col-wide`, `--hk-status-col`, `--hk-ext-mark`) live inside the panel's shadow root and are not a themable contract.

## Verification

Run locally, all green:

- `pytest tests/unit` — 1936 passed, 2 skipped
- `npx vitest run` — 1215 passed across 38 files, including the i18n key-and-value parity gate over all 16 locales
- `bash ci/e2e-up.sh` — **274 passed** across the desktop, tablet and phone projects
- `bash ci/test-mutation-frontend.sh` — **100.00%** (37/37) on the changed surface in `utils.ts`, threshold 80%
- `vale` — the hit set for `CHANGELOG.md` and `README.md` is identical to `origin/main`'s, compared in a matching worktree rather than by line number

New coverage:

- `frontend/test/device-chip.test.js` — six cases over the chip's destination on all four surfaces, including that a click opens the appliance without also opening the task beside it
- `frontend/test/utils.test.js` — the `assetsForTask` ranking, one case per branch, plus stable order across three same-claim appliances and the no-device guard
- `tests/e2e/tests/list-honesty.spec.ts` — asserts every task row's chip cluster and status pill share one x, that an appliance row opens from either end, and that it still says what it holds
- `tests/e2e/screenshots-device-link.capture.ts` — its task-row leg now asserts the appliance page and that the chip carries no leaves-the-panel mark

Four defects local verification caught and this fixes: the status pill wrapped and clipped inside its fixed container height; the action column's content-driven width was moving the rails; the overdue border was moving them again by 2px; and splitting the appliance row into tracks left most of it unclickable.

## Checklist

      step through it — n/a: no new surface. The tour already steps through the task
      list and the appliance detail, and both show the change as they are.
      `const.py`), `preview-release` label applied — bumped to `0.22.0b1`; please add
      the label.

https://claude.ai/code/session_01WcQS9U1ZjHJdb4zsfqgQpq


prestomation:
/q review Be critical and skeptical. Surface the most serious problems first, and do not withhold the minor ones.

Two things changed: a device chip's destination now depends on the surface it sits on, and both list rows became CSS grids of fixed tracks.

**Correctness.** `assetForTask` / `assetsForTask` / `assetRank` in `frontend/src/utils.ts` decide which appliance a task belongs to, and one chip can only go to one place. Is the ranking right when several appliances claim the same device? Does the archived-behind-live rule hold at every rank? `taskRelatesToAsset` is now `assetRank(...) > 0` — check that this is behaviour-preserving against the backend twin `assets.task_relates_to_asset`. I dropped an explicit index tiebreak from the sort on the grounds that `Array.prototype.sort` is spec-stable; challenge that if you disagree. `panel-task-form.ts` now builds its consumable and document pickers on the same helper, so check the option sets did not change.

**Navigation.** `wireDeviceChips` gained the `PanelHost` and dispatches on `data-asset-id` before `data-device-id`. Does `stopPropagation` still stop a chip click reaching the row opener it sits beside, on the appliance list where the chip is nested inside the row's `.detail-open` block? Is there any path that pushes two history entries for one click, given `wireDetail` returns early for a task page and falls through for an appliance page?

**CSS cascade and layout.** `panel-styles.ts` is one long stylesheet, so look for rules that cancel each other. The rails are gated in a `@media (min-width: 1151px)` block while `.hk-row-asset` also has flex rules outside it; the `@media (max-width: 700px)` phone grid and the `.hk-master` picker pane both override again. Does every band resolve to one intended layout? Is the reserved transparent `border-left` on every `ha-card.hk-card` safe against HA's own card border and border-radius? The chip fade uses a `mask-image` on `.hk-chips-inline` — check it cannot hide a chip that fits, or break the `+n` unfold.

**Maintainability.** `deviceChip` returns a bare chip or a wrapped one depending on an optional argument. Is that a good seam, or should the mark be a separate concern? Note that `wireLists` recounts the "+n" label from the child count of `.hk-chips-inline`, so the wrapper is deliberate.

**HA best practices.** The chip is an `ha-assist-chip` with `role="link"`, `tabindex="0"` and a `title`, and it now sometimes navigates in-panel and sometimes out. Is that accessible and honest? Should the in-panel case be a real anchor?

**Security.** Every interpolation into the chip's HTML should be escaped; `assetId` reaches `data-asset-id` and then `_openDetail`.

---

prestomation:
/q review Be critical and skeptical about this second commit specifically, and rank the most serious problems first. Do not withhold minor ones.

The full e2e suite caught that splitting the appliance row into grid tracks left `.grow.detail-open` holding only the first track, so a click on most of a row no longer opened the appliance. The fix is a stretched-link overlay in `panel-styles.ts`:

```css
.hk-row-asset > .grow.detail-open::after { content: ''; position: absolute; inset: 0; }
.hk-row-asset .hk-chips, .hk-row-asset .hk-status { pointer-events: none; }
.hk-row-asset .hk-chips > *, .hk-row-asset .hk-status > *, .hk-chevron {
  pointer-events: auto; position: relative; z-index: 1;
}
```

**Correctness and cascade.** The overlay resolves `inset: 0` against `ha-card.hk-card`, which carries `position: relative`. Is that the containing block in every state — tree view with `--hk-tree-depth`, the `.hk-master` picker pane, an archived row, an overdue row with its reserved rail? Does `pointer-events: none` on the chip track break anything else that lives there, and is `.hk-chevron` still hittable above the overlay in tree view?

**Accessibility.** The overlay is a pseudo-element, so the keyboard path is unchanged and `.grow` keeps its own `role="button"` and `tabindex`. Is a stretched link the right pattern here, or does it break text selection or a screen reader's reading of the row in a way that matters?

**The seam.** Is a `::after` overlay the right way to make a grid row wholly clickable, or should the opener be the `.hk-card-row` itself with the chips relying on `stopPropagation` as they already do? Argue for whichever you think is more maintainable.

**Test change.** `a11y.spec.ts` now clicks `.hk-card[data-id] .detail-open` rather than the card's centre, because the centre now lands on the device chip, which has a destination of its own. That row-opens-from-either-end behaviour moved to an explicit assertion in `list-honesty.spec.ts`. Tell me if you read that as weakening coverage rather than aiming it.

---

prestomation:
/q review Be critical about the merge commit specifically. `main` shipped 0.21.0 mid-branch and this resolves the three conflicts it caused.

**Version.** I resolved `manifest.json` and `const.py` to `0.22.0b1` rather than `0.21.0b7`, on the grounds that PEP 440 sorts `0.21.0bN` below the shipped `0.21.0` and HACS would then offer the stable to a beta tester as an upgrade. Check that against AGENTS.md's beta-versioning rule and against what `release.yml` and `ci/check-changelog-release-gap.py` expect. Is `0.22.0b1` the right number, and are the two files the only places the version lives?

**CHANGELOG.** My section is renamed `## [0.22.0b1]` and sits above the shipped `## [0.21.0]`, with the older `0.21.0bN` beta sections left below it as `main` has them. Is the ordering and the retained-beta convention right? `notify-issues` reads the shipped version's section, so confirm nothing in my section would be misread.

**Bullet leads.** `main` added the rule that a bold lead is a heading, not a sentence: 2-5 words, no articles or prepositions. I rewrote both leads to `**Appliance-first device chips.**` and `**Aligned list rows.**`. Do those satisfy the new rule and read as standalone headlines, given `summarize()` in `ci/release-issues.py` quotes only the lead? Also check the sentences after each lead against the three-sentence budget and the "what a user gets, not the mechanism" rule.

**Merge correctness.** Nothing outside those three files conflicted, and `main`'s other changes were to `AGENTS.md` and two `.amazonq/rules/` files. Confirm the merge did not silently drop anything from either side, and that no conflict markers survive.

---

PR #302, opened 2026-09-06
feat(notifications): give a notification its own icon and accent color

### LINKED ISSUE #293: Feature Request: Customizable Notification Icons
Now that notification channel and urgency are configurable, I'd like to request adding notification icon customization. This helps different notifications stand out on your notification bar. I believe the companion app supports any MDI icon.

prestomation:
Fixes #293

## What changed

Each saved notification gains a **Notification icon** and an **Accent color (iPhone)**,
beside the channel and the urgency it already has.

The Notifications page with an icon and a color on each collapsed row

Each collapsed row carries its icon as a chip filled with its color, so the list is a
legend of the whole house. The 2 fields sit under Urgency in the editor, each with
helper text naming what the phone does with it:

The notification editor with the Notification icon and Accent color fields

### Backend

`notifications.py` holds the whole backend change.

- `normalize_icon` and `normalize_color` clamp a value to `mdi:<name>` or to
  `#rrggbb`, and to `""` when it is unusable. They repair rather than raise, because
  `normalize_notification` repairs a stored document.
- `payload_data` emits `notification_icon` and `color`. An empty field emits no key, so
  a notification nobody has configured sends the payload it always sent
  (`test_payload_data_unconfigured_is_what_it_always_was` passes untouched).

### Frontend

- `notificationSchema` gains `icon` through the existing `selIcon()`, and `color`
  through a new `selColorRgb()`. `hexToRgb` and `rgbToHex` convert at the form
  boundary. The picker uses RGB and the companion app uses hex.
- `utils.notifyRowChip` builds the row chip. `itemCard` gains an optional `badge`
  builder, because Profiles and Companions share it.
- 4 keys in all 16 locale files: the 2 labels and their 2 helpers.

## What the phones actually do, and how that was established

The maintainer tested the first version on Android and reported that neither field did
anything. Reading `NotificationFunctions.kt` in `home-assistant/android` and Android's
own notification design guide turned up 3 claims in the Home Assistant companion docs
that are wrong. Each one had already shipped into this branch as a bug or a
misleading doc line:

| Doc claim | What the source says |
| --- | --- |
| `notification_icon_color` is iOS-only | `handleColor` reads it **first** and falls back to `color` only when absent. Sending the iOS default of white replaced the Android accent with white, so the color field looked inert. **The key is now never sent.** |
| An unresolvable icon shows no icon | `handleSmallIcon` falls back to `R.drawable.ic_stat_ic_notification`, the Home Assistant icon, which is why a bad name reads as the feature doing nothing. |
| Android applies `color` | [Android's guide](https://developer.android.com/design/ui/mobile/guides/home-screen/notifications): *"For most styles, the system applies this color only if the notification is for a foreground service notification."* Ours is neither that nor MediaStyle, and the app never calls `setColorized`. |

So on Android the icon lands in the **status bar only** (the notification in the shade
keeps the app's icon, which no payload key changes), and the color does nothing from
Android 12 on. Android also resolves the name against the Iconics font the app bundles
(`community-material-typeface`, MDI 7.0.96), which trails the set the panel's picker
offers, so a recent icon silently falls back to the Home Assistant icon.

The color field stays, because it works on iPhone and it fills the row chip. It says so
rather than pretending otherwise: the label reads **Accent color (iPhone)** and the
helper reads *"iPhone shows this color behind the icon. Android 12 and later ignore
it."*

This reverses a convention added earlier in this same PR, which said a platform
difference producing a native result on both sides needed no documentation. That rule
is what produced the confusion, and `.amazonq/rules/architecture-and-code.md` now says
the opposite, plus: read the companion app's source before trusting its docs about a
payload key.

## One-way doors

| Surface | Shape | Where it is relied on |
| --- | --- | --- |
| `icon` on a notification | `"mdi:<name>"`, or `""` | `.storage/home_keeper`, and the `home_keeper.set_options` service |
| `color` on a notification | `"#rrggbb"` in lower case, or `""` | The same 2 places. A named color such as `red` is never stored. |
| `notification_icon` reaches iOS | iPhone draws the notification in its message style | To stop that later is a second visible change |
| An empty field sends no key | An unset field adds nothing to the payload | `test_payload_data_unconfigured_is_what_it_always_was` |
| A bad value clamps to `""` | It does not raise, and it does not reach the phone | A mistake falls back to the Home Assistant icon |

The 2 fields are separate keys and not an object, so a third field costs nothing later.

## Versioning: this cuts 0.22.0b3

`main` moved 3 times under this branch: 0.21.0 stable, then 0.22.0b1, then 0.22.0b2
with #303's search box. Each time the version this branch carried became one that
`release.yml` would refuse to publish, and `changelog-release-gap` caught the last 2.
The bullet now sits in its own `## [0.22.0b3]` section; `## [0.22.0b2]` is byte-identical
to main again.

## Checklist

      step through it

## Testing

- `pytest tests/unit` — 1949 passed.
- `npx vitest run` — 1262 passed in 38 files, with the locale parity gate.
- `python3 ci/check-changelog-release-gap.py` — reproduced the failure, then passes.
- `bash ci/test-mutation-python.sh` — **96.94%**, threshold 80%. The survivors are
  pre-existing lines in `normalize_notification` that the diff pulled into scope, and
  each is equivalent (`bool(None)` matches `bool(False)`).
- `bash ci/test-mutation-frontend.sh` — **85.90%**, threshold 80%.
- `ruff check`, `ruff format --check` and `tsc --noEmit` are clean.
- `tests/e2e/tests/settings.spec.ts` — 17 passed, asserting both selectors render, both
  helpers say what the phone does, and the row chip carries the saved icon and color.
  Also passed with `CHROMIUM_EXEC` unset, which is the browser CI drives.
- Screenshots recaptured and read after each copy change.

**mypy did not run here.** This container has Python 3.11, and Home Assistant needs
3.14, so pip resolves Home Assistant 2024.3.3 instead. `ci/check-ha-version.py` fails on
that, which is what it is for. The `lint.yml` job is the check, and it has passed.

## Notes for review

- `color_rgb` renders as `ha-selector-color_rgb`, with an underscore. Home Assistant
  names the element after the raw selector key, and `icon` renders as
  `ha-selector-icon`.
- `normalize_icon` and `normalize_color` reject a non-string rather than coerce one.
  `str(value or "")` left an equivalent mutant behind, because no fallback string can
  pass either check.
- `writing-style.md` gains a rule that a product or platform name takes no article, as
  the one exception to "do not omit articles". The article still belongs to a common
  noun the name modifies, so "An Android channel" stays.
- `tests/integration/ha_config` is a bind mount, so `docker compose down -v` does not
  reset it. A partial capture run left `home_keeper_intro_dismissed` in the gitignored
  `.storage`, and every later capture then failed at step 0.
  `git clean -fdX tests/integration/ha_config/` recovers it.

https://claude.ai/code/session_01SczDhNzCqkYXp32RUyV412


prestomation:
/q review Be critical and skeptical. Surface the most serious issues first, and do not withhold minor ones. This adds 2 stored fields on each saved notification (`icon`, `color`) that reach the companion app as `notification_icon`, `color` and `notification_icon_color`.

Please scrutinize:

**Correctness and edge cases**
- `normalize_icon` / `normalize_color` in `notifications.py`. They clamp instead of raising. Is `""` right for every unusable input, and does any real value get rejected that should not be? The character set is `^mdi:[a-z0-9-]+$`, and it lower-cases the whole string.
- `payload_data` emits `notification_icon_color` only when both an icon and a color are set. Is that the right condition, or does iOS need it in some other case?
- The round trip `hexToRgb` / `rgbToHex` in `forms.ts`. `color_rgb` returns `list[int]` of 3 bytes. Check the padding, the empty case, and anything the picker can hand back that these do not cover.
- The stored shape survives `set_options` from a script, not only from the panel. `payload_data` re-normalizes for that reason. Is that enough?

**Security**
- `utils.notifyRowChip` builds markup that reaches `innerHTML`. The icon goes through `escapeHTML` after a character-set check, and the color is only accepted as `#rrggbb` before it lands in a `style` attribute. Is either gate escapable?

**HA best practices**
- `selColorRgb()` uses the `color_rgb` selector, which renders as `ha-selector-color_rgb`. Is that the right selector for a value the phone reads as hex?
- `itemCard` gained an optional `badge` builder that Profiles and Companions also pass through. Is that the right seam, or does it belong somewhere else?

**Maintainability and performance**
- The badge repaints on every form change, alongside the existing footer repaint. Is that wasteful in a card with many notifications?
- The frontend duplicates the backend's icon validation. The backend is the authority. Is the duplicate worth its drift risk?

**A design decision worth challenging.** Both platforms read these keys and render them differently. Android draws the icon in the status bar and tints the glyph with `color`. An iPhone draws it as a sender icon on a circle that `color` fills. Home Keeper sends 1 value and documents nothing about the difference, on the grounds that each result is native. Argue the other side if you think a user is going to be surprised.

---

prestomation:
/q review This push is release plumbing only, so keep it short and skeptical about that specific change rather than re-reviewing the feature.

`changelog-release-gap` failed on the previous head. v0.22.0b1 is already a published tag, so folding the Added bullet into its CHANGELOG section would have documented a change that never ships: `release.yml` reads an unchanged `manifest.json` version, sees the tag exists, and skips. That is the #236 gap the check exists to catch.

The fix moves the bullet into a new `## [0.22.0b2]` section and bumps `manifest.json` and `const.py` (`PANEL_VERSION`) to match.

Please check:
- Is `0.22.0b2` the right next version, and does anything else in the repo carry a version that has now drifted?
- The `## [0.22.0b1]` section should be byte-identical to main again. Confirm the bullet moved rather than being duplicated.
- `(Fixes #293)` sits in the first paragraph of the bullet, which is what `ci/release-issues.py` reads. Confirm the move did not break that, and that `notify-issues` will still find #293 in the shipped section.
- Anything about this that would make the release job behave unexpectedly.

---

prestomation:
/q review Be critical and skeptical. This push is a one-line e2e fix, but review the whole change with fresh eyes and put the most serious issues first. Do not withhold minor ones.

Focus on:

- **Correctness of the payload.** `payload_data` in `notifications.py` now adds `notification_icon` and `color`. It deliberately never sends `notification_icon_color`, because the Android app reads that key first and falls back to `color` only when it is absent. Check that the omission holds on every path, and that an unset field still adds no key at all.
- **The normalizers.** `normalize_icon` and `normalize_color` clamp to `""` rather than raise. Look for an input that reaches storage or the phone unvalidated, and for a value that a user could set in the panel but the backend then throws away without saying so.
- **Escaping.** `notifyRowChip` in `utils.ts` puts a stored color into a `style` attribute and a stored icon name into an `ha-icon` attribute. Try to find a value that escapes either one.
- **Locale parity.** 16 locale files gained 4 keys. Check for a translation that changed meaning, and for a string that reads as a machine translation.
- **Maintainability.** The optional `badge` builder on `itemCard` is shared with profiles and companions. Is that the right seam, or does it leak notification concepts into a generic card?
- **Home Assistant practice.** The `color_rgb` selector round-trips through `hexToRgb`/`rgbToHex` at the form boundary. Is that conversion lossy or surprising in any case?

---

PR #303, opened 2026-09-06
feat(panel): filter the task and appliance lists by typed text

### LINKED ISSUE #297: Add a Search box to the main screen
On the main Tasks screen, it'd be nice to add a search/filter input box that filters the tasks to only show those containing the text typed (case-insensitive). When the box is cleared, all tasks are displayed. 
When you have a lot of "automatic" (companion) tasks, it can be really hard to find the one you're looking for without this.

prestomation:
Fixes #297

## What changed

One search box in the control row above both lists. Typing narrows the list as the words arrive; an empty box shows everything. It is shared by the Tasks and Appliances tabs, and it is deliberately not persisted.

**Matching** is pure and lives in `card-filter.ts`, on the Stryker mutate surface. `normalizeSearch` folds case, diacritics and whitespace; `matchesQuery` requires every typed word somewhere in the row's own text, in any order, so `water filter` finds the task "Replace filter" on the appliance "Water heater". A task matches on its name, notes, device, effective area, the integration that supplied it, and that integration's chips. An appliance matches on its name, notes, manufacturer, model, serial number, device and area. Matching the owning integration is the part that answers the issue: a household running several companions cannot tell one generated task from another by name.

**Typing does not re-render.** `_render` replaces the whole shadow tree, which would replace the box being typed in and take the caret with it. `_setQuery` patches `#hk-list` and the scope pills' counts in place instead, the way `_patchSettingsSection` already patches the Settings page. The counts are part of that patch on purpose: `filterCounts` now applies the query too, so a pill cannot promise twelve above a list of three.

Two consequences worth a reviewer's attention:

- The `details.hk-group` collapse handlers move from `wireControls` to `wireLists`. `renderGroups` emits those inside `#hk-list`, so the partial rebuild replaces them. Left where they were, the collapse memory died silently after the first keystroke.
- `wireLists`, `wireDetailOpeners` and `wireDeviceChips` now take a `ParentNode` rather than a `ShadowRoot`, so the re-wire can be scoped to the list. On an appliance's page `.detail-open` also matches the detail pane beside it, which this pass does not rebuild and must not bind twice.

A search opens every section it draws, and reads `_collapsed` without writing it. Monitored and Completed start collapsed, and a monitored companion task is what the reader is most often hunting for, so a match behind a shut heading would fail at the one job this has.

Also in this PR, since it is a convention rather than code: `AGENTS.md`, `CLAUDE.md` and `.amazonq/rules/testing-and-workflow.md` now require **both** a desktop and a phone screenshot for every changed surface. The phone layout below 700px is a different layout, not a narrower one, so a desktop-only shot leaves half the change unreviewed.

### Merged with main

`main` moved twice while this was in flight, and both moves matter:

- **0.21.0 shipped, then 0.22.0b1.** This branch had bumped to `0.21.0b7`, which PEP 440 sorts *below* the released `0.21.0` — HACS would have offered the stable as an "upgrade" to anyone on the beta. Rebased onto `0.22.0b2`, with the CHANGELOG section renamed to match.
- **#301 landed on the same rows**, giving both lists fixed rails and pointing a task's device chip at its appliance. `wireDeviceChips` gained a `PanelHost` parameter there and a `ParentNode` root here, so it now takes both. Every panel screenshot has been recaptured against the merged UI, because the shots on this branch were drawn before half of what they document.

## Screenshots

Desktop, Tasks. The counts follow the box: All reads 4 above a list of 4, and the two empty scopes recede.

The Tasks tab with the word filter typed in the Search box and the list narrowed to four matching tasks

Desktop, Appliances.

The Appliances tab with the word water typed in the Search box and the list narrowed to the water heater

Phone, Tasks. The search chip takes a row of its own under the wrapped scope pills, and the field grows to the width instead of holding the 16ch it has beside them on a desktop.

The Tasks tab on a phone with the word filter typed in the Search box, the scope pills wrapped above it and the list narrowed

Phone, Appliances.

The Appliances tab on a phone with the word water typed in the Search box and the list narrowed to the water heater

## One-way doors

**None.** No service input, event payload, storage key, entity attribute or URL segment is added or changed.

Three decisions that are cheap to revisit, recorded so the review can disagree with them now rather than later:

- The query is **session-only**. It is not written to `localStorage` and not put in the URL, so every panel load starts with the whole list. Every other control in that row does persist, so this is the odd one out on purpose: a scope pill states its own name and count, while a remembered substring is a short list with no visible reason for being short. Adding persistence later is three lines; removing it later is a behaviour regression.
- **One query for both tabs** rather than a task query and an appliance query. A text query has the same vocabulary on both, unlike `_filter` and `_assetFilter`, whose vocabularies differ.
- **Every word must match, in any order**, rather than one literal substring. This differs from the issue's literal wording only for multi-word queries, where it is strictly more forgiving.

## Testing

Run locally against the merged head:

| Gate | Result |
| --- | --- |
| `npx vitest run` | 1249 passed. 25 new matcher tests, 13 new panel tests. |
| Playwright e2e, desktop + phone + tablet | 278 passed pre-merge; the merged head is re-running in CI. |
| `pytest tests/unit` | 1937 passed. `297` added to `_KNOWN_ISSUES` in `test_release_issues.py`. |
| `tsc --noEmit` | Clean. |
| `bash ci/test-mutation-frontend.sh` | **100%**, 23 killed, 3 annotated as equivalent. |
| Vale | 0 new hits, comparing hit *sets* against `origin/main` rather than filtering by added line. |

The three annotated mutants were each checked over a spread of field sets and queries before being called equivalent, not assumed: the two `.filter(Boolean)` removals cannot change the answer because `normalizeSearch` has already trimmed and collapsed, and the `?? []` replacement cannot, because the fallback array's elements carry no `label`.

New coverage worth naming:

- `panel.test.js` asserts the input element is the *same node* after typing, that quick Done and row-opening still work on the patched rows and bind exactly once, and that the group-collapse memory survives.
- `list-honesty.spec.ts` asserts the pill count equals the visible card count, which is the contract that file exists for.
- `a11y.spec.ts` brings the box into the "no two controls share a name" check, and asserts a character typed into the middle of the box lands in the middle.

## Checklist

- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML `` tag (desktop and phone)
- [x] New user-facing feature → version bumped to `0.22.0b2` (`manifest.json` + `const.py`), `preview-release` label applied

https://claude.ai/code/session_013jNUNQkX6kFgv13dVw83LJ


prestomation:
/q review Please be critical and skeptical, and rank the most serious problems first without withholding the minor ones. This PR adds a text filter to the panel's two lists and, unusually, makes typing patch the DOM in place instead of re-rendering, so the riskiest parts are the ones a passing test suite would not notice.

**Correctness of the partial re-render (`panel.ts` `_applyQuery`).** `_render()` rebuilds the whole shadow tree; `_setQuery` instead replaces `#hk-list`'s `innerHTML` and re-runs `wireLists`, `wireDetailOpeners` and `wireDeviceChips` scoped to that element, then patches the scope-pill counts via `patchFilterCounts`. Please hunt for anything a full render does that this pass skips and should not: listeners left on detached nodes, a handler bound twice, state read at render time that is now stale, the open deferral menu, `_liveHassEls` (which it deliberately does not reset), `_chipsExpanded`, `_treeCollapsed`, focus restoration. Is the fallback to `this._render()` when `#hk-list` is absent correct on a task detail page and on Settings? Is there a path where `_query` and the rendered list can disagree?

**The wiring move.** The `details.hk-group` collapse handlers moved from `wireControls` to `wireLists` because `renderGroups` emits them inside `#hk-list`. Check I have not created the mirror bug: is anything still bound in `wireControls` that lives inside the list, or anything now bound in `wireLists` that lives outside it? `wireDetailOpeners` matches `.detail-open`, which also exists in the appliance detail pane — is scoping to the list element sufficient on every route, including the appliance master/detail split?

**Correctness of the matcher (`card-filter.ts`).** `normalizeSearch` does NFD + `\p{Diacritic}` strip + `toLowerCase` + whitespace collapse; `matchesQuery` requires every word. Edge cases welcome: Turkish dotless i, Han and Cyrillic, combining marks that are not diacritics, a query of only punctuation, very long input, and whether the space-joined haystack can ever let a match span two fields. Three mutants are annotated as equivalent — please check those claims rather than trusting them.

**List honesty.** `filterCounts` now applies the query so a pill cannot promise more than the list shows. Is there any surface left where a count, a section heading count, or an empty-state message can disagree with the rows actually rendered? The appliance tree view filters before `buildAssetTree`, which promotes a matching child whose parent does not match — is that sound, including under the area-grouped branch?

**Performance.** The matcher runs per row per keystroke, and `filterCounts` runs it again for the counts. There is no debounce, on purpose. At a few hundred tasks, is the double pass worth avoiding, and is anything else in the patch accidentally O(n²)?

**Security.** The query is free user text interpolated into an `innerHTML` assignment as `value="${escapeHTML(p._query)}"`. Please look for any path where it reaches markup unescaped.

**Home Assistant practice and maintainability.** The control is a hand-rolled `<input type="search">` plus an `ha-icon-button`, not an `ha-textfield`. Is that the right call for a custom panel, and does the accessibility story hold up (name, the composition guard for IME input, the 16px font that stops WKWebView zooming)? Also: is `_query` on the `PanelHost` seam justified, or should it have been threaded as an argument?

**One-way doors.** I claim there are none, and that the query being session-only rather than persisted in `localStorage` is the right default. Push back if you disagree.

---

prestomation:
Thanks — I checked both ⚠️ items rather than taking them on trust, and I think both are false positives. Reasoning, so it can be argued with:

**1. `_chipsExpanded` applying to the wrong task.** The scenario needs a task to be deleted and a new one to be minted with the same id. Ids come from `str(uuid.uuid4())` in `models.build_task`, so an id is never reused. Independently of that, `_applyQuery` cannot reach the scenario at all: it re-renders from `p._tasks` without touching it, and the only thing that replaces that array is `_refresh()`, which calls the full `_render()`, not this patch. And `_chipsExpanded` has always survived a full render, so this PR does not change the exposure either way. Not fixing.

**2. Cross-field matching through the space-join.** This one cannot happen, rather than being unlikely. A term can never contain a space: `normalizeSearch` collapses runs of whitespace and then `split(' ')` separates on it, so every term is a single word. The join inserts a space exactly where two fields meet, so no single-word term can straddle the boundary. `matchesQuery(['Replace filter', 'Water heater'], 'filterwater')` returns `false`, and `card-filter.test.js` pins it as *"never lets a word match across two fields"*. Not fixing.

Everything else in the review matches my own reading, including that the three annotated mutants are genuinely equivalent — thanks for checking those specifically rather than taking the annotations at face value.

One correction to the review's line references for anyone reading later: the `?? []` fallback is at `card-filter.ts:373`, not 382, and the two `.filter(Boolean)` calls are at 344 and 346. The conclusions are unaffected.

CI on the current head (`131c2bd`) is green so far — mypy, ruff, vale, TypeScript, Python, test, hacs and pytest all pass; e2e, integration and the walkthrough capture are still running.

---

PR #307, opened 2026-09-07
feat(sensor): show the usage between completions on a metered task

### LINKED ISSUE #305: Feature request: Show actual usage between maintenance completions
First of all: thank you for Home Keeper. I originally wrote down a fairly long [maintenance-tracking wishlist for Homebox](https://github.com/sysadminsmedia/homebox/discussions/10), covering things like time- and usage-based service intervals, history, costs, parts and "whichever comes first" scheduling.

Most of that was unfortunately never implemented in Homebox, while Home Keeper already checks off almost everything I was looking for.

One small thing that would still make the history much more useful for me is showing the actual usage between two completions for Usage/Meter tasks.

For example, if an oil-change task has completion readings at:

```
120,000 km
134,800 km
150,200 km
163,900 km
```

it would be helpful if the history could also show:

```
+14,800 km
+15,400 km
+13,700 km
```

The same idea would apply to operating hours, cycles, litres, wears, etc.

Optionally, a few simple derived values could be useful too:

```
Last interval:     13,700 km
Average interval:  14,633 km
Shortest:          13,700 km
Longest:           15,400 km
```

The nice part is that Home Keeper already stores the completion readings, so this would mostly be about turning existing history data into something easier to evaluate.

My main use cases are cars, bikes, household equipment and outdoor gear, where it is useful to know not only when something is due, but also how long the previous part or service interval actually lasted.

Thanks again for the work on Home Keeper — it is currently much closer to the maintenance tracker I had been looking for than anything else I tried.

prestomation:
Fixes #305

## What changed

A usage task has logged its meter reading at each completion since #236, and the history has shown it. What it never showed is the subtraction. Four oil changes at 120,000 / 134,800 / 150,200 / 163,900 km leave the reader to work out that the last one ran 13,700 km, and no reading on the page says whether that is short.

**Per row.** Each completion now carries the usage since the completion before it, in accent ink beside the reading it comes from, so the two reconcile as you read down the list.

**Above the list.** A strip of four figures: last interval, average, shortest, longest. With only one interval it collapses to the last one, because the average, the shortest and the longest of a single number are all that number.

**Digits.** Readings and intervals go through a new `formatReading`, so an odometer reads `163,900 km` in the viewer's language rather than `163900`. This changes the reading that was already there, not just the new figure.

**On the entity.** The same four figures ride the next-due sensor beside the `usage_consumed` family, so a template can ask "is this one running short?" without fetching the log and subtracting.

Only a usage meter gets any of this. A threshold task records a reading too, but that reading is a measurement (airflow at 58%) rather than a meter that only climbs, so the difference between two of them is not usage.

### The interval rule

Written once per side and tested to the same edge cases on both, so a change to one that the other does not follow is a red test rather than two screens disagreeing:

- **Completions only.** A skip resets the meter too, but it is the record of work that was *not* done, so an interval spanning one is still a single service interval. Same rule that keeps a skip out of the completion tally and the cadence.
- **Ordered by parsed timestamp.** `completions` keeps insertion order, so a back-dated completion sits at the end of the list, and a store that has seen a timezone change holds mixed offsets. Sorting the strings would put those in the wrong order.
- **A negative difference is dropped.** The meter was reset or replaced between the two, so there is no usage figure to report. A zero difference is kept, because two completions at the same reading really are 0 units apart.

`meter_start` is deliberately not used, even though it would give the oldest completion an interval too: it measures usage since the last re-anchor, so with a skip in between it prints a number that does not reconcile with the two readings on screen.

## One-way doors

Four new attributes on `sensor.<task>_next_due`, absent until two completions carry a reading. Automations and templates will read these names.

| Attribute | Shape |
| --- | --- |
| `usage_last_interval` | float, meter units, rounded to 3 |
| `usage_avg_interval` | float, mean of the kept intervals |
| `usage_min_interval` | float, shortest kept interval |
| `usage_max_interval` | float, longest kept interval |

Nothing else is committed. The strip, the delta chip, the locale keys and `formatReading` are internal panel surfaces.

## Screenshots

The seeded nozzle task grows two earlier services at 120 and 375 h, so the four figures are four different numbers. Nothing else in the fixture moves — the meter still reads "120 of 300 h used" and the card still counts down "in 180 h".

Desktop:

A metered task history with a Last, Average, Shortest and Longest interval strip above rows reading at 660 h and plus 285 h

Phone, where the strip becomes a two by two block and the delta rides the chips line rather than the date row:

The same history at 390px, with the four interval figures in two rows of two

## Design

The summary treatment was chosen from mocked options before implementation, per the new AGENTS.md rule below. The shipped labels read "Last interval" and not "Last": each cell is what a screen reader gets on its own, and "Shortest" alone says nothing.

## Also in this change

Two new AGENTS.md workflow rules, both asked for while planning this one:

- A plan starts with the CHANGELOG entry it will ship.
- A UI change with a subjective visual choice starts with an artifact of mocked options, drawn in the panel's own tokens at both widths, agreed before the plan.

## Checklist

- [x] `CHANGELOG.md` updated — `0.22.0b4`, with `(Fixes #305)`
- [x] Panel UI changed → screenshots committed under `docs/images/` and embedded above with an HTML `` tag
- [ ] New user-facing UI surface → not applicable: the history tab is already a walkthrough beat and this adds no new surface
- [x] New user-facing feature → version bumped to `0.22.0b4` (`manifest.json` + `const.py`), `preview-release` label applied

## Verified locally

- `pytest tests/unit` — 1972 passed
- `npx vitest run` — 1286 passed
- `pytest tests/integration/test_sensor_watcher.py` — 25 passed, including a new test that drives three completions through the real service and reads the four attributes off the real entity
- `npx playwright test tests/usage-intervals.spec.ts` — 6 passed, desktop, tablet and phone
- `ci/test-mutation-python.sh` — 100%, `ci/test-mutation-frontend.sh` — 100%
- `ruff check` / `ruff format --check` clean; `vale` adds no hits over `origin/main` on any edited file

`mypy` could not run here: Home Assistant does not install on this container's Python. CI covers it.

https://claude.ai/code/session_01RhiU285PmCWSGjgz4avryd


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first without withholding the minor ones.

This derives "usage between completions" from the meter readings a usage task already logs, in two places that must agree: `sensor_tasks.usage_intervals` / `usage_interval_stats` (Python) and `utils.usageIntervalStats` (TypeScript). Focus on:

**Correctness of the shared rule.** The two implementations are meant to be identical. Look for ways they can disagree: Python sorts parsed `datetime`s while TypeScript sorts `Date.getTime()`; Python guards with `isinstance` plus `math.isfinite` while TypeScript relies on `Number.isFinite` alone; Python catches `ValueError` from `recurrence._parse` while TypeScript checks for `NaN`. Are there timestamp shapes one accepts and the other rejects? Naive (offset-free) stamps? DST boundaries? Sub-second precision? Both sorts are stable, and `sensor_tasks` claims that entries at the same instant keep their stored order — does that actually hold on both sides?

**The dropped-negative rule.** A negative difference is discarded because the meter was reset between the two completions. Consider whether that silently hides a real problem: a mistyped reading corrected later, `set_task_meter` re-anchoring, a rolled-over counter, or the automatic re-baseline in `evaluate_usage`. Is dropping the pair the right call, and does dropping it distort `average` (which divides by the surviving count) in a way a user would misread?

**Skips.** Skips are excluded, so an interval spanning a skip is reported whole even though the skip re-anchored the meter — the row's figure then differs from what the task itself measured. I chose this over `meter_start` so the delta reconciles with the two readings shown. Argue the other side if you think it is wrong.

**The new entity attributes.** `usage_last_interval`, `usage_avg_interval`, `usage_min_interval` and `usage_max_interval` on the next-due sensor. Are those good HA names next to the existing `usage_consumed` / `usage_remaining` / `usage_percent`? Is `round(x, 3)` right, and is it a problem that they vanish entirely when there is only one completion rather than reporting `None`? Any concern about attribute-count or state-size limits on a task with 500 completions?

**Performance.** `extra_state_attributes` now walks the whole completion history (capped at 500) on every state write. Is that a hot path worth caring about, and would the panel re-deriving `usageIntervalStats` on every history render matter for a long log?

**Maintainability.** Is `formatReading` in the right module, and should the existing `formatQuantity` and `round1` call sites converge on it? Is gating the panel strip on `task.sensor?.mode === 'usage'` inside `panel-history.ts` the right place, or should that predicate live in `utils.ts` beside `taskRecordsReading` and `meterRemaining`?

**Security.** The delta chip is the one place `completionMeta` emits raw HTML around an escaped value. Please check that closely.

---

prestomation:
Thanks — the timestamp finding was a real bug, and it was worse than described. Pushed a50ecdc.

**1. Timestamp parsing divergence — confirmed, and it crashed.** Not just a divergence: a naive stamp beside an aware one raises `TypeError: can't compare offset-naive and offset-aware datetimes` out of the sort, and `usage_intervals` runs inside `extra_state_attributes`, so one such entry would take the next-due sensor's state write with it. Reproduced before fixing:

```
CRASH: TypeError can't compare offset-naive and offset-aware datetimes
```

The panel had the quieter half of the same bug: `new Date('2024-02-04T09:00:00')` parses as the *viewer's* zone, so one history sorted differently in Berlin and in Seattle. Both sides now drop an offset-free stamp, and both suites pin the shapes Home Assistant actually writes — `Z`, `+00:00`, a real offset, and the colon-free `+0000`. Nothing Home Keeper writes is offset-free; hand-edited storage can be.

`sensor_tasks.latest_decision_ts` has the same latent exposure on the same input. Out of scope here rather than unnoticed.

**2. Delta chip HTML — checked, and it is safe.** `escapeHTML` wraps the whole `t()` result, not the interpolated value:

```ts
`<span class="hk-hist-delta">${escapeHTML(
  t('completion.usageInterval', { value: formatReading(interval, unit, lang) }),
)}</span>`
```

`unit` is user input, `formatReading` concatenates it, and the escape happens after interpolation, so nothing user-supplied reaches the markup unescaped. The template is `"+{value}"` in every one of the 16 locales, and `i18n.interpolate` only substitutes `{word}` — it cannot introduce markup. No change.

**3. Negative interval — not fixed here, and I want the maintainer's call.** You are right that the drop is silent. Two things reduce the harm: both readings stay on screen, so the fall is visible in the rows themselves, and the average divides by the surviving count, so it stays the mean of real intervals rather than being pulled down. Adding a "meter reset" marker means a new chip, a new locale key in 16 files and fresh screenshots at both widths, which widens the PR past what #305 asked for. Better as a follow-up.

**4. Performance — measured the shape, not worth acting on.** Capped at 500 completions of one float subtraction each; the surrounding `extra_state_attributes` already walks the same list for `completions_count` and `max()`. Memoizing would need a cache key over a mutable list, which costs more than it saves.

**5–7. Sort stability, sub-second precision, attribute naming.** Agreed on all three, no change.

**8. `formatReading` placement.** Agreed it is in the right module. Converging the `formatQuantity` and `round1` call sites is a separate change: `formatQuantity` is spare-part stock at 3 decimals with no grouping, which is deliberately not a meter reading.

**9. Extract the strip predicate — done.** Now `showsUsageIntervals` in `utils.ts`, beside `taskRecordsReading` and `meterRemaining`, so the mode check is unit-tested and mutation-scored rather than inline in the renderer.

**Your recommendation 1, followed further than asked.** Adding the timestamp-shape tests surfaced a second gap the mutation run caught: text that *ends* in a valid offset but is not a date (`not a date+00:00`) gets past the tail check and has to be caught by the parse. Both sides now have that test, and both mutation gates are back to 100%.

---

PR #309, opened 2026-09-08
feat: import and export, a portable document for migrations

### LINKED ISSUE #308: Feature request: Create a completed task
Installed Home Keeper recently and you've done a nice job!  Thank you.  Have maintained a OneNote collection for documenting all of my house maintenance information and history for years but started thinking about how Home Assistant would be a natural repository for that information and found Home Keeper.  The Appliances collection supports my needs very well but would be very helpful if there was a UI for logging a one-off completed task/history event.  A lot of appliances have issues that I have to fix that are not scheduled and I want to capture that information.  Also want to backfill that history into Home Keeper for reference.  Not easy with current UI.  Thanks

prestomation:
Fixes #308

## What changed

Home Keeper reads and writes its data as one YAML document. Export writes it, import
reads it, and they are the same shape — so an export is also a worked example of the
format, which is what makes "show an assistant your export and ask for twelve more
like it" a complete instruction.

Every export names a **published JSON Schema** on its first line, so opening one in an
editor gets validation and field completion with no setup.

The reporter asked to backfill years of one-off repair history that lives in OneNote.
This is the general answer: a migration from a spreadsheet, a dead app, another Home
Assistant, or a generated file is a document rather than a hundred panel visits.

## Screenshots

Settings → Import and export, showing a preview of what an import would change:

The Import and export card with a link to the format documentation, a pasted YAML document, and a preview reporting one new appliance, one new task and one history entry

On a phone the section opens on its own and the two actions take a row each, so both
keep a full tap target:

The same card at phone width, scrolled so the Preview and Import buttons sit on their own row above the bottom tab bar

## The document

```yaml
home_keeper:
  format: 1
appliances:
  - external_id: furnace
    name: Furnace
    area: Basement
    parts: [{ name: Filter, part_number: FILXXFCC0021, stock: 2 }]
tasks:
  - external_id: furnace-filter
    name: Replace furnace filter
    appliance: furnace
    interval: 3
    unit: months
    history:
      - { completed_at: "2026-03-04", note: Used a MERV 13, cost: 24.50 }
```

**A record takes the same fields as the action that creates it.** A `tasks` record is
an `add_task` payload plus `external_id` / `appliance` / `area` / `enabled` /
`history` / `skips`; an `appliances` record is an `add_asset` payload plus
`external_id` / `area` / `archived`. So `services.yaml` and the generated API
reference document the format for free, and a unit test fails if they ever drift.

## The published schema

`ci/generate_schema.py` writes
`https://prestomation.github.io/ha-home-keeper/schema/home-keeper-1.schema.json` at
docs build time. Nothing is committed and nothing is hand-written: `voluptuous_openapi`
— the library Home Assistant core itself uses to expose service schemas — converts
`TRANSFER_DOCUMENT_SCHEMA`, which is `ADD_TASK_SCHEMA` and `ADD_ASSET_SCHEMA` plus the
eight keys a document carries that no action takes. **Add a field to an action tomorrow
and it appears in the schema, typed, with no edit to the generator.** Because the
converter reads the real validators, `parts` and `card_links` come out as typed arrays
of typed objects — the `object:` selector in `services.yaml` would have flattened both
to `{"type": "object"}`.

Three places the conversion needed help, each a stated exception rather than a mapping
table: `cv.boolean` infers as a string because it also accepts `"on"` and `"yes"`;
`cv.string` inside a `vol.Any` comes out as an anything-goes branch; and
`active_season` is declared `vol.Any(None, dict, [dict])`, which carries no shape at
all. That last one was found by the gate rather than by reading — the schema said
`object` and the export writes a list.

`tests/unit/test_generate_schema.py` is the gate. It runs the generator as a subprocess
(`tests/conftest.py` promises nothing imports the real package in-process), checks the
field set against `services.yaml` **and** against the records `models.build_task` /
`assets.build_asset` really build, validates a maximal export, and holds seven
documents against both the schema and `plan_import`. It runs in `lint.yml`'s **mypy**
job, which already installs Home Assistant on a Python at its floor — the unit lane
deliberately has none, and a path-filtered docs job would miss a drift caused by a
change to `models.build_task`.

Two things the schema deliberately does **not** do. It never subtracts `EXCLUDED_*`
keys: those name what the *export* omits, and three of them (`last_completed`,
`source`, appliance `device_id`) are real fields an import still takes. And
`additionalProperties` stays open at every level, because an unknown field or section
is a named warning on import and never an error — a schema stricter than the code it
describes sends people to fix files that would have imported.

## Why YAML, and what that costs

`README.md` already showed a worked YAML example one paragraph after calling the export
"a JSON file"; the code is the half that changed. YAML is a superset of JSON, so every
file written by an earlier build still imports — with two caveats now stated in the
README: YAML forbids a **tab** as indentation, and a bare `no` / `on` / `NO` reads as a
boolean. Neither can bite an exported file, because PyYAML's emitter quotes any string
that would resolve to another type.

The dialect is pinned on both sides. The dumper indents a record under its section,
writes a multi-line note as a `|-` block, and never emits an anchor. The loader
**refuses an anchor outright** — that is what closes a billion-laughs expansion bomb,
which `MAX_IMPORT_RECORDS` cannot see because a document is already expanded in memory
by the time there are records to count — and **drops the implicit timestamp resolver**,
so a hand-written `completed_at: 2026-03-04` stays the string it was when the document
was JSON. Stock `SafeLoader` returns a `date` there, which reaches
`datetime.fromisoformat` as a non-string and raises, and which the store cannot
serialize. `MAX_IMPORT_BYTES` (8 MiB) is checked before the parser runs.

`TRANSFER_FORMAT` stays **1**. Serialization is not shape.

## One parser, on the backend

`transfer.py` had no parser at all: every text-to-object step happened in the browser
as `JSON.parse`, which is why the module's first stated property — "export emits
exactly what import accepts" — was only half true. `parse_document` is now the partner
of `document_to_yaml`, both pure and both mutation-scored, and the panel sends the
pasted text rather than a parsed object.

That keeps the bundle free of runtime dependencies (it has none today, and `js-yaml`
would have been the first, in a file Home Assistant serves before authentication), and
it means a syntax error comes back as an ordinary problem row naming **a line and a
column** — which a parse in the browser could never have produced. `import_data` takes
a mapping or a string on the service and its websocket twin alike.

## Upsert, on a stated primary key

Three steps, first hit wins, checked in order:

| Step | Field | Notes |
| --- | --- | --- |
| 1 | `id` | Home Keeper's own uuid. Every export carries it, so re-importing an export is an exact upsert. |
| 2 | `external_id` | **New field.** The author's own key — what makes a re-run of a migration script update rather than duplicate. |
| 3 | `name` | Exact, then trimmed and case-folded: the ladder `resolve.py` already uses for the `*_id` service fields. |

The steps are **independent**: a stated id that names nothing in the store is a
*create*, never a fall-through to a name match that would overwrite an unrelated
record. Ambiguity raises and names every candidate rather than guessing. `match: none`
turns matching off for a backfill whose names legitimately repeat.

## How this stays covered as Home Keeper grows

This was the part worth designing carefully, and it is mostly *not* code.

1. **The export is a denylist, not an allowlist.** `transfer.py` names the fields it
   *excludes* (each with a mandatory reason) and passes everything else through;
   import feeds a record straight back into `models.normalize_fields`, which already
   knows every field. **Add a field to `build_task` tomorrow and it travels in both
   directions with no edit here.**
2. **`test_transfer_roundtrip.py`** builds a maximal record per recurrence type, sends
   it out and back, and compares field for field. It now goes **through the serialized
   file** rather than the dict, so all seven assertions cover the serializer too: a
   value YAML resolves to a date or a boolean on the way back is a lost field, and a
   test handing the mapping straight to `plan_import` could never see it.
3. **`test_transfer_coverage.py`** `ast`-parses `store._save` and fails when a new
   *storage section* appears in neither the document nor `EXCLUDED_STORE_KEYS` — the
   gate for a whole new feature, which is how `declarative_companions` arrived.
4. A third check keeps "a record's fields are the service's fields" true rather than
   claimed, by comparing against `services.yaml`.
5. **`test_generate_schema.py`** extends the same idea to the published contract.
6. `api_surface.py` gains a **Data portability** surface row, and the convention is
   written into `.amazonq/rules/architecture-and-code.md` + `AGENTS.md`.

Forward compatibility rides on one distinction: an unknown field or section is a
**named warning** and the import proceeds (an older Home Keeper must read a newer
document), but it is named, because a field nobody read is data that did not arrive. A
newer `format` is a hard error.

## Six bugs the gates and the review caught

- **The key probe only built a floating task**, so `freq` / `anchor` / `sensor` would
  have been greeted with "not a field Home Keeper reads" and dropped. Found by the
  services.yaml comparison; fixed by probing every shape rather than one.
- **The asset update path planned the updates mapping, not the merged record.** The
  applier writes a payload verbatim, so an updated appliance landed as a fragment with
  no `id` and the next device reconcile died on `KeyError: 'id'`. Found in a browser;
  now gated by a test that asserts every planned payload is a whole record.
- **`parent_asset_id` was passed through unresolved**, so a nested appliance tree
  carried the author's own key as an id and the store's link cleanup nulled it on the
  next load — losing the tree with nothing said.
- **An appliance could be made its own parent, or two made each other's.**
  `store._validate_parent` refuses a loop on every other write path; an import writes
  built records straight in, so it was the one path that could break the invariant.
  The check runs against the *projected* graph, because one document can re-parent
  both halves of a loop in the same run and each edge looks harmless alone.
- **A `source` that was not a mapping was stored verbatim**, which
  `store.async_repoint_device_ids` walks unguarded. A probe of every typed field in
  `ADD_TASK_SCHEMA` showed `source` was the only one `build_task` did not already
  reject, so the fix sits in `models` beside `validate_managed_by`.
- **The published schema said `active_season` was an object**, while every export
  writes a list. `ADD_TASK_SCHEMA` declares it `vol.Any(None, dict, [dict])`, which
  carries no shape at all through the converter. Found by validating a real export
  against the generated schema — which is the whole reason that test exists.

## Panel

Settings gets an **Import and export** section. Export downloads the file. Import is
deliberately two steps: **Preview** runs the same call with `dry_run` set, and the
Import button stays shut until a preview of *exactly this text* came back clean.
Editing the box takes that permission away again, so a preview of one document can
never license importing another. A picked file is read in the browser and put in the
same box you could have typed into — one input, and no upload endpoint to guard.

The card's help line is now a single link to the format documentation. That moved the
markup out of the locale and into the template, which removed the card's one unescaped
interpolation — a security simplification as well as a shorter sentence.

## Notes

- **An import fires one event per record, never one per completion.** Backfilling a
  decade is not a decade of completions happening now; firing each one would replay
  through notifications and to-do sync as if it were.
- **Validation is all-or-nothing.** Every record is checked before anything is
  written, so a document that is wrong anywhere writes nothing.
- **No HTTP view.** The document is text and rides the websocket, so the "a binary
  can't ride a service call" exception stays a two-route exception.
- Both services are admin-only, gated in the service *and* its websocket twin, with
  rows added to `docs/SECURITY.md` and `tests/integration/test_admin_gates.py`.
- **Home Assistant must be imported before the component in `ci/generate_schema.py`.**
  Since 2026.3, importing `homeassistant` swaps probatio in for voluptuous, and the
  integration's first third-party import is `import voluptuous as vol` — so a
  standalone generator ended up with two different `Schema` classes and
  `voluptuous_openapi` died on an unhashable dict lookup. `ci/install-schema-deps.sh`
  then installs the converter under Home Assistant's own `package_constraints.txt`,
  because which version of it works is Home Assistant's choice, not ours. CI found
  both; a local Home Assistant old enough to still use voluptuous cannot.
- Only 3 PNGs are committed. The version footer moved every other panel shot, and
  re-committing 66 unrelated binaries would bury the diff.

## One-way doors

1. **`external_id`** — the name, that it lives on tasks *and* appliances, its 128-char
   cap, and that it is the step-2 primary key.
2. **The envelope** — `home_keeper.format: 1`, and the section names `tasks` /
   `appliances` (not `assets`). Format 1 leaves the top level open for a `recipes:`
   section later.
3. **Record field names** — `history[].completed_at`, `skips[].skipped_at`,
   `appliance`, `area`, `archived`, and that ids win over names.
4. **The match ladder** — id → `external_id` → name, first hit wins, ambiguity raises.
5. **Service surface** — `export_data` / `import_data`, their fields (`include`,
   `document`, `dry_run`, `match`), and the report keys (`ok`, `dry_run`, `counts`,
   `records`, `problems`, and each problem's `severity` / `path`).
   - `export_data` returns **`{document, yaml}`**. `export_inventory` already returns
     `{inventory, csv}`, so a format-named key is the house shape.
   - `import_data.document` accepts a mapping **or the whole file as one string**, on
     the service and the websocket twin. Widening only.
6. **The export's YAML dialect** — block style, insertion order, sequences indented
   under their key, multi-line strings as `|-`, and every date-like or boolean-like
   string single-quoted. Anyone diffing exports across versions depends on it.
7. **The importer's dialect** — `SafeLoader` minus the implicit timestamp resolver,
   with anchors and merge keys refused. `2026-03-04` stays text; `no` / `on` / `NO` are
   booleans. **`MAX_IMPORT_BYTES` = 8 MiB.**
8. **The download** — `home-keeper-<YYYY-MM-DD>.yaml`, `application/yaml`.
9. **The schema URL and `$id`** — `…/schema/home-keeper-1.schema.json`, JSON Schema
   draft 2020-12, named on the first line of every exported file.
   ⚠️ **A future `TRANSFER_FORMAT` bump must ADD a file, never replace one.** The
   production deploy runs `clean: true` with only the two preview umbrellas excluded,
   so the moment the generator stops writing `home-keeper-1.schema.json` that file
   leaves the site and every document exported this year stops validating. The bump
   has to add `schema/` to `clean-exclude` or emit the frozen older formats too. This
   is written into the generator's docstring as well.
10. **What is excluded** — reconciler-owned and managed tasks, and uploaded file blobs.
    The schema also withholds `source` and `managed_by`: both are accepted on import,
    but a task with either is one `is_portable_task` then declines to export, so
    offering them in an editor's completions would be a trap.
11. **`recurrence.record_skip`** and **`resolve.match_by_name`** are new public
    functions in modules other integrations read. `models.validate_source` joins them,
    and it tightens `build_task`: a non-mapping `source` now raises rather than being
    stored.
12. **Not a one-way door, said so review does not re-litigate it:** `TRANSFER_FORMAT`
    stays `1`. A format-1 document is format 1 whether it is written as JSON or YAML,
    and every existing export still imports.

## Checklist

- [x] Tests run locally — `pytest tests/unit` (2194 passed, 3 skipped, 1 xfailed),
      `npx vitest` (1314 passed, 41 files), `tsc --noEmit`, `ruff check` +
      `ruff format --check`, `mypy` (clean bar one pre-existing stale-HA artefact),
      `vale` (zero new hits vs the base, compared as a per-file *set* and cross-checked
      against the raw `ai-tells` regexes)
- [x] e2e — the full Playwright suite, 298 passed, including the new YAML
      syntax-error case
- [x] Schema gate — `pytest tests/unit/test_generate_schema.py` (16 passed) against a
      real Home Assistant; `python3 ci/generate_schema.py` writes both files
- [x] `CHANGELOG.md` — folded into the still-unreleased `## [0.23.0b1]`, with
      `(Fixes #308)` in the bullet's first paragraph. No `b2`: `v0.23.0b1` is not a
      published tag, so `changelog-release-gap` stays green
- [x] Panel UI changed → screenshots at both widths, committed and embedded above
- [x] New user-facing UI surface → `walkthrough.capture.ts` extended, both tours
- [x] `preview-release` label applied
- [ ] Mutation gate — re-run pending for the new `transfer.py` code (84.93% on an
      earlier head)

https://claude.ai/code/session_015TCBfZvwSmnqqxvVNrKfK5


prestomation:
/q review Be critical and skeptical. This adds a portable JSON document that `export_data` writes and `import_data` reads, upserting on an id → `external_id` → name ladder. Surface the most serious issues first, and do not withhold minor ones.

Please scrutinise, roughly in this order:

**Correctness of the match ladder** (`transfer.py` `_Matcher.match`, `_plan_asset`, `_plan_task`). Can a stated `id` that names nothing ever fall through and overwrite an unrelated record? Can two document records upsert onto the same stored one? Is `claimed` consulted everywhere it must be? Is `external_id: ""` genuinely unmatchable? A wrong answer here silently destroys somebody's data during a migration, so treat any path that could pick the wrong record as critical.

**Id remapping.** `_claim_id` keeps a stated uuid when it is free, else mints one. Are all cross-references rewritten through the result — `parent_asset_id`, `card_links[].asset_id`, a task's `device_id` via the `_PLANNED_ASSET_PREFIX` placeholder? What happens when a document names a `parent_asset_id` that is only planned, not stored? I believe that case is unhandled; please confirm and say how bad it is.

**The applier's ordering** (`transfer_runner.async_import_document`). Appliances → one `async_reconcile_assets` → tasks → one entry reload. Is the placeholder resolution safe when provisioning fails or an asset ends up without a `device_id`? Is a partial write possible between the two `async_import_records` calls, and does the all-or-nothing validation actually hold end to end?

**Timezone and date handling.** `_iso` qualifies a naive value with the caller's zone; `apply_history` replays chronologically because `apply_completion` stamps `last_completed` unconditionally and the history cap keeps the list *tail*. Is the sort stable and total? Any DST or ordering case where the resulting `next_due` / `last_completed` is wrong? Does a duplicate `ts` behave sensibly?

**Security.** Both services are `_verify_admin` plus `@require_admin` on the twin. Is the gate the first statement, and is there any path (the websocket schema, the panel) that reaches a write without it? The document carries serials, costs and notes — anything here that leaks to a non-admin?

**Resource bounds.** `MAX_IMPORT_RECORDS` is 2000 per section and history is capped at 500 per task, but a document could still carry 2000 tasks × 500 completions in one websocket message. Is that a real event-loop or memory problem, and should the cap be on total work rather than per section?

**Maintainability.** The "derive, never restate" inversion (export excludes rather than includes) is the load-bearing idea — is it actually watertight, or are there fields that would silently fail to travel without the round-trip test noticing? Are `transfer.py` / `transfer_runner.py` split at the right seam? Is `_known_task_keys`' probe-based derivation robust, or will a future field shape break it?

**HA best practices.** `async_reconcile_assets` + `async_config_entries.async_reload` called from inside a websocket handler; firing bus events after the save rather than before; `SupportsResponse.ONLY` on a mutating service.

---

prestomation:
Thanks — the must-fix is real and is now fixed in 2518101.

**Failed provisioning silently orphaned the task.** The applier resolved its placeholder to `None` and carried on, so the task imported standalone with nothing said. Provisioning is the one step that can still fail after validation has passed, and by then the appliances are already written, so there is nothing to roll back to — the honest answer is to import the task and report it. It now lands as a named warning in the report saying which task, why, and what to do:

> "Flush the boiler" was imported without its appliance: Home Assistant did not give that appliance a device. Open the appliance, then set the task's appliance again.

Two notes on the rest:

**`parent_asset_id` was a real gap**, though not the one you looked at — you couldn't see it because it did not exist when you reviewed. A document naming a parent in the *same document* carried the author's own key straight through as an id, and `store._clean_relationship_links` nulled it on the next load, losing a nested appliance tree with nothing said. Fixed in 8c5063f: it resolves against the document first and then the store, and a parent nobody has is now an error naming what to do (list the parent before its children). Three tests cover it.

**`card_links[].asset_id` you flagged as "would need similar treatment"** — it doesn't, deliberately. A card link points at an appliance *document or metadata entry* the dashboard card shows on a task's row, and those entry ids are minted per appliance. A document that carries one either re-imports onto the same install (where the id is still valid, since the id ladder preserves it) or lands on a different one (where it names nothing and the card simply shows no link). Remapping it would mean matching sub-records across installs, which is a whole second ladder for a cosmetic field. Worth revisiting if it turns out to matter; not worth the surface now.

On **resource bounds**: agreed it is theoretical rather than practical, and I'd rather not add a second cap that has to be explained. 2000 tasks × 500 completions is not a document anyone hand-writes, and the per-section cap already bounds the common accident (a runaway generator). If it ever bites, the right fix is a cap on total entries, not a smaller per-section one.

---

custom_components/home_keeper/transfer_runner.py
```diff
@@ -0,0 +1,110 @@
...
+        await devices.async_reconcile_assets(hass, coord.entry, coord.store)
+
+    planned_tasks = plan.for_section("tasks")
+    if planned_tasks:
+        for record in planned_tasks:
+            asset_id = transfer.planned_asset_id(record.payload.get("device_id"))
+            if asset_id is None:
+                continue
+            asset = coord.store.get_asset(asset_id)
+            record.payload["device_id"] = (asset or {}).get("device_id")
```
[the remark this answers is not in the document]
prestomation: Fixed — the applier now builds its write list instead of editing the plan.

The "data corruption" half of this is not right, though, and it's worth saying why so the reasoning is on the record. Payloads are never shared: `_plan_task` sets `payload` to either `models.build_task(...)` or `models.merge_update(...)`, each of which returns a fresh dict, so one record's mutation could not reach another. `frozen=True` also never promised the dict was immutable — it stops attribute rebinding, not mutation of what an attribute points at.

The underlying critique stands on its own merits, which is why I took it: a frozen record whose dict gets rewritten underneath it is a half-promise, and the plan should stay the record of what was *planned* so anything reading it back — the report, a log, a future retry — sees what the preview showed. It now copies:

```python
payload = record.payload
asset_id = transfer.planned_asset_id(payload.get("device_id"))
if asset_id is not None:
    device_id = (coord.store.get_asset(asset_id) or {}).get("device_id")
    payload = {**payload, "device_id": device_id}
    ...
tasks_to_write.append((record.record_id, payload, record.action == "create"))
```

---

prestomation:
/q review Second pass, on the four commits since your first review (`8c5063f`, `2518101`, `f2e1b4f`, `e353fbc`). Be critical and skeptical, and please re-check the earlier ground too, because the fixes moved code you already read. Most serious first; do not withhold minor ones.

**The parent-appliance resolver** (`transfer.py`, `_parent_id` and its callers). A document can name a parent appliance that is only *planned*, not yet stored. The resolver now tries the document first, then the store, and errors when nobody has the parent. Does that ordering ever pick the wrong appliance when a planned key and a stored name collide? Is a cycle possible — two appliances naming each other as parent — and what does the store do with one? Is a self-parent caught? A wrong answer here builds a bogus appliance tree during a migration, so treat a bad link as serious.

**The orphaned-task warning** (`transfer_runner.py`). A task whose appliance never got a device is now imported standalone with a named warning rather than a silent `None`. Two things to attack: is the report the caller sees still coherent (`ok` stays true, counts still add up) when a warning is appended after `as_report`? And is "import it and say so" the right call at all, given the appliances are already written by then and there is nothing to roll back to?

**The frozen-plan refactor.** `async_import_document` now builds a separate write list instead of mutating `record.payload` in place. Confirm no path still writes through the frozen `PlannedRecord`, and that the dict identity change (`{**payload, "device_id": ...}`) cannot drop or alias anything the store then persists.

**The new assertions** (`e353fbc`). They pin `Problem.section`, `.index` and `.path`, not only the message, and add a case for a task carrying a *non-reserved* `source`. Are the paths asserted actually the paths the panel renders? Is the reserved-source check now airtight, or is there still a set operation that a mutant would survive?

**Re-check the ladder and the id remapping under the new code.** `_plan_asset` now returns the merged record rather than the payload fragment. Does every cross-reference still resolve through `_claim_id`, and can an update path now carry a field it did not carry before?

Also worth a fresh look: whether `_known_task_keys`' per-shape probes cover every recurrence type `models.build_task` can produce, and whether a future shape would fail loudly or silently.

---

prestomation:
Both findings were real. Fixed in `e820000`, with 12 tests each verified to fail against the previous commit.

**The parent loop.** You were right that it is not caught, and I reproduced both shapes: an appliance made its own parent, and two made each other's. Two corrections to the consequences, because they change how bad it is:

- It is **not** nulled by `_clean_relationship_links` — that only nulls a parent that does not *exist*, so a loop between two real appliances is kept.
- It does **not** hang: `devices._ancestor_depth` carries a `seen` set and logs `Cyclic parent chain detected`.

So the damage is quieter than either of us said: the loop persists, the depth sort that provisions parents before children is meaningless for those appliances, and every later reconcile logs an error against a tree the user cannot see is broken. `store._validate_parent` refuses this on every other write path; an import writes built records straight in, so it was the one path that could break the invariant.

The check has to run against the **projected** graph rather than the stored one — a single document can re-parent both halves of a loop in the same run, and taken one at a time each edge looks harmless. Both ends are reported, since changing either one opens the loop. I reused `assets.would_create_cycle` rather than writing a second walker, so the store and the planner cannot come to disagree about what a loop is.

Note the pure-create case was never reachable: a parent must already be planned (an earlier index) or stored, so creates cannot point forward into a loop. It is the update path that gets there.

**The non-dict `source`.** Also real, and it does reach storage — but not for the reason given. Every reconciler guards with `isinstance(source, dict)`, so a string cannot smuggle a task into a reconciler's ownership. What it does is break `store.async_repoint_device_ids`, which walks `(task.get("source") or {}).values()` with no such guard: `'str' object has no attribute 'values'`, long after the write that stored it.

Rather than patch the import path, I probed every typed field in `ADD_TASK_SCHEMA` against `build_task`. `managed_by`, `labels`, `card_links`, `active_season`, `task_chips`, `interval` and `require_tag_scan` are all already rejected; `source` was the only one missed. So the fix went into `models.validate_source`, next to `validate_managed_by` that set the precedent — one field wide, not a class, and now closed for every path rather than for import alone. (`merge_update` never needed it: it does not carry `source` from updates at all, so it keeps the existing value.)

**On the three you cleared** — the report stays coherent, no path writes through the frozen record, and the remapping is complete: agreed, and thanks for checking the reasoning rather than just the shape. On `_known_task_keys`, your read is right that a new recurrence type without a probe would warn-and-drop rather than fail silently; `test_transfer_coverage.py` is what makes that loud, since it compares the probed key set against `services.yaml` and is how the original floating-only probe was caught.

---

/q review Third pass, on `e820000`. Be critical and skeptical; most serious first, and do not withhold minor ones.

**The loop check itself** (`transfer._looping_parents`). It builds a projected graph from the stored parents overlaid with every planned one, then asks `assets.would_create_cycle` per record. Is the projection actually complete — does anything else set `parent_asset_id` after this point? Can a rejected appliance leave a dangling reference behind, given I strip it from `planned`, `planned_assets` and `asset_refs` but leave `matcher.claimed` alone? Is there a loop shape the walk misses, such as one closed through an appliance the document *deletes* rather than re-parents?

**Ordering.** The check runs after the appliance loop and before the task loop. Is that the right seam, or can a task planned later change a parent link? Does stripping `asset_refs` make any task's `appliance` resolve to a *different* appliance rather than to nothing — a silent re-attachment would be worse than the error it replaces.

**`validate_source` placement.** It runs inside `build_task` only. Is there a live path that persists a task without going through `build_task` or `merge_update` — a reconciler building a dict itself, or a storage migration — where a non-dict `source` could still arrive? And is rejecting a non-dict `source` a behaviour change for any existing caller, given voluptuous already typed it `dict` at both services?

**The claim I am making from a probe.** I concluded "`source` is the only typed field `build_task` does not enforce" from probing `ADD_TASK_SCHEMA`'s fields with wrong-typed values. Attack that method: fields where the schema's constraint is a *value* rather than a type, `UPDATE_TASK_SCHEMA` fields that differ, and the asset side (`ADD_ASSET_SCHEMA` vs `build_asset`), which I did not probe.

---

prestomation:
Rebased onto #311 (`8878056 release: 0.22.0`). Force-pushed; head is now `2d0477c`.

**The version had to change, and not by one step.** This PR was cutting `0.22.0b5`, but `v0.22.0b5` is already a published tag — #310 shipped it and owns that changelog section on `main`. Reusing the number would have collided with a released section, and `changelog-release-gap` would have failed the PR the moment it sat on current `main`. Bumping to `b6` would have been worse: #311 makes `0.22.0` stable, and PEP 440 sorts every `0.22.0bN` *below* it, so HACS would offer beta testers the stable as an upgrade that is really a downgrade. Per the rule in `AGENTS.md`, a feature landing after a stable cut opens the next minor. So this is **`0.23.0b1`**, with a fresh section above `## [0.22.0]`:

```
## [0.23.0b1]
### Added
- **Import and export.** …(Fixes #308)
- **A stable key of your own.** …
```

Three conflicts, all in the version files (`manifest.json`, `const.py`, `CHANGELOG.md`); every code and locale file auto-merged, including `panel-settings.ts`, which #310 also touched.

Re-ran the whole local gate on the rebased tree rather than trusting the pre-rebase run: **2159 unit tests**, **1312 frontend tests** across 41 files, `tsc --noEmit`, `ruff check` + `ruff format --check`, and `vale` compared as a per-file hit *set* against the new base — zero added hits.

Two things to note:

- **This branch carries #311's commit until #311 merges.** Since the repo squash-merges, I will re-rebase onto `main` once #311 lands, so the squashed release commit is not duplicated here.
- **The 2 committed screenshots still show the pre-rebase version footer.** They document the Import and export card, which has not changed; re-shooting both widths to update a footer string did not seem worth a full e2e cycle. Say the word if you would rather have them re-captured.

---

/q review Fourth pass, and specifically on the rebase — the previous three covered the feature itself, so please concentrate on what moving to a new base could have broken rather than re-reviewing `transfer.py` from scratch.

**The auto-merges.** `panel-settings.ts` was touched by both this PR (the new Import and export section) and #310 (the phone companion row fix, and the profile delete guard), and git merged them with no conflict. Is that merge actually correct, or has it produced a section list, a wire-up, or a narrow-layout branch that is internally inconsistent? The 16 locale files also auto-merged — check for a key added twice, a key orphaned, or a locale that drifted from `en.json`.

**The version change.** `0.22.0b5` → `0.23.0b1` across `manifest.json`, `const.py` and `CHANGELOG.md`. Is anything else in the tree pinned to the old string — a test fixture, a snapshot, a doc, `quality_scale.yaml`, the upgrade-test pins? And is the new `## [0.23.0b1]` section shaped the way `ci/release-issues.py` expects, given it parses the shipped version's section and quotes only the bold lead?

**What the rebase silently kept.** My commits were replayed onto a tree that gained #310's changes. Is there anywhere the two changes are individually right but jointly wrong — the settings section ordering, the phone layout with one more row in it, or a test that asserted a section count that is now off by one?

---

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+
+Home Keeper reads and writes its data as one file. Use it to move to another Home
+Assistant, or to keep a copy of your own. Use it also to bring in records you keep
+in a spreadsheet or an old app.
+
+**Settings → Import and export** has both halves.
+
+- **Export** saves every task and appliance, with the history, as a JSON file.
+- **Import** reads such a file. Paste it, or choose it from disk, then press
+  **Preview**. Preview reports what would change and writes nothing. **Import** stays
```
prestomation: "and writes nothing" is not the style we want. Instead  "Preview only reports what would..."

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+steps, and stops at the first that matches:
+
+1. **`id`**, Home Keeper's own id. Every export includes it, so re-importing an
+   export updates the same records.
+2. **`external_id`**, the name you choose. Set this one in a file you write
+   yourself, because it makes a second run update the same records instead of making
+   a copy of everything.
+3. **`name`**, an exact match first, then one that ignores case and spaces.
+
+Home Keeper creates a record that matches nothing. When a name matches 2 records you
```
prestomation: "that matches nothing" > ”That matches no record " add this to style guide

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+get an error that names both, because Home Keeper does not guess which one you meant.
+To create every record and match nothing, pass `match: none` to
+`home_keeper.import_data`.
+
+An update only changes the fields the file states. Fields it leaves out keep the
+value they have.
+
+### What the file does not hold
```
prestomation: "Import/Export Limitations"

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+
+The export says so in its own `home_keeper` block. The preview reports anything in a
```
prestomation: This is awful reword this

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+file that Home Keeper did not read.
+
+- **Uploaded manuals and receipts.** A JSON document has no room for a file, so
+  upload those again after an import. A link to a document is only text, so it stays.
+- **Tasks that another part of Home Keeper owns**, such as a wear part's replacement
+  reminder, a buy reminder, a problem-sensor mirror, or a recipe's task. Home Keeper
+  builds these again from the appliance and its parts, which the file does include.
+- **Tasks that another integration owns.**
```
prestomation: "tasks created by companions"

README.md
```diff
@@ -1516,6 +1518,130 @@ Delete button. On a phone the row stacks, and the buttons take a line of their o
...
+- **Settings, profiles, notifications and recipes.** These stay in the config entry.
+
+### Ask an assistant to write one
```
prestomation: Say "AI Agent" everywhere instead of assistant

README.md
prestomation: Done in `b3355ca` — "Preview only reports what would change."

Written into the style guide as a general rule, since it is not just this line:

> **Say what a thing does, not what it does not do.** A trailing negative reads as an afterthought and leaves the reader to work out the point. Use a negative only when the absence *is* the point, such as a limitation.

---

README.md
prestomation: Done in `b3355ca`, in both places on that paragraph — the second one ("To create every record and match nothing") had the same fault.

> Home Keeper creates a record that matches no stored record. When a name matches 2 records you get an error that names both, because Home Keeper does not guess which one you meant. To create every record and match no stored record, pass `match: none`.

Added to the style guide under **Words**, as you asked:

> **Name the noun. Do not write "nothing", "anything", or "something"** where a real noun fits. Write "a record that matches no stored record", not "a record that matches nothing". A vague pronoun makes the reader guess which thing is meant.

I used "no stored record" rather than "no record", because the sentence is already about a record — "a record that matches no record" reads as a riddle. Say the word if you want the shorter form anyway.

---

README.md
prestomation: Changed to **`### Import and export limitations`** in `b3355ca` — I took the substance ("limitations", not the oblique "What the file does not hold") but not the exact string, on two of this repo's own rules, so flagging it rather than quietly deviating:

- **One approved name per thing.** The feature is "Import and export" in the Settings section, the `## ` heading, and the glossary. "Import/Export" is a second name for it.
- Every other heading in the file is sentence case — "How a record finds its match", "Ask an AI agent to write one".

If you want the literal `Import/Export Limitations` anyway, say so and I will change it — it is your style call, and I would rather ask once than have you re-flag it. Leaving this thread open for that reason.

No anchor fallout either way: nothing links to `#what-the-file-does-not-hold`, and `doc-anchors.test.js` passes (13 tests).

---

README.md
prestomation: Agreed — it opened by naming the reporting mechanism instead of saying what the section was about. Reworded in `b3355ca`:

> The file does not hold every record. An export lists what it left out in its own `home_keeper` block. An import reports every field in a file that it did not read.

Worth recording: my first attempt opened "Some records do not travel", and **vale rejected it** — `HomeKeeper.Personification: 'travel'. Software does not do this.` A rule this project wrote for itself caught the replacement for the sentence you flagged. `README.md` is now at zero vale hits.

---

README.md
prestomation: Done in `b3355ca` — **Tasks created by companions.** Better on the glossary's own terms too: "companion" is the approved name and "integration" was doing double duty for both Home Assistant integrations in general and the ones that register with Home Keeper.

---

README.md
prestomation: Done in `b3355ca`. Both uses in this section are now "AI agent" — the heading ("Ask an AI agent to write one") and the sentence under it. I checked the rest of `README.md`, `docs/*.md`, `strings.json`, `services.yaml` and `locales/en.json`: every other hit is a line-wrapped **Home** Assistant, so there was nothing else to change.

Added as a glossary row rather than a one-off edit, so the next writer does not reintroduce it:

| Use this | Not this | Note |
| --- | --- | --- |
| AI agent | assistant, AI assistant, chatbot, LLM, model | Use it for any tool a user asks to write or read Home Keeper data on their behalf. "Home Assistant" is unrelated and keeps its name. |

I used lower-case "agent" to match every other glossary entry; the row is what governs, so tell me if you want it capitalised as "AI Agent" and I will change both.

---

prestomation:
Rebased onto `main` (#311's squash) and added property-based testing. Head is `a4d2914`.

**The rebase.** #311 merged as a squash, so this branch carried a duplicate of it and conflicted with `main` on the three version files — GitHub could not build `refs/pull/309/merge`, so **no CI ran at all** on the previous head. `git rebase origin/main` dropped the duplicate automatically (identical patch-ids), 12 commits replayed with no conflicts, `0.23.0b1` intact.

**Property tests.** The suite had 2,159 unit tests and zero generated inputs — strong on structural drift (`ast` gates, parity checks, mutation scoring), no coverage of the value space at all. 14 properties now cover `recurrence.py` and `transfer.py`, the two surfaces where a shape bug is expensive.

Each transfer property names a mutant that was alive on the previous head, and **every one of those claims was checked by mutating the line and watching the test go red** rather than asserted in a comment:

| Property | Mutant it kills |
| --- | --- |
| T1 `problems == ()` over a generated store | the `EXTERNAL_ID` rename in `_task_out` |
| T2b same instant, differing metadata | `events.sort(key=None)` |
| T3 generated parent graphs | `continue` → `break` in `_looping_parents` |
| T4 reserved-source subsets | `&` → `|` in `is_portable_task` |

Two things worth reporting, because they change what I would have claimed:

**The DST comment at `recurrence.py:207` is correct.** I expected to find a defect there — every recurrence test runs on a fixed offset and no test in this repository had ever built a `ZoneInfo`, so the claim was entirely untested. It holds: 2,000 draws across 7 real zones, including a 30-minute DST shift and a 45-minute base offset, found no counterexample. Keeping 02:30 across a spring-forward makes that step 23 hours rather than 24, which stays positive only because the smallest schedule step is a day and the largest shift is an hour. That is a margin nobody had written down. R5a and R5b write it down.

**A season boundary on February 29 disagrees with itself.** `_next_season_start` clamps the day to Feb 28 in a non-leap year, exactly as its docstring says it does. `in_season` applies no such clamp when it checks. So `{"start": "02-29", "end": "03-05"}` — an ordinary 6-day season, not a degenerate single-day one — clamps a floating task to a date **outside the season it was just clamped into**, in 3 years out of 4. Pinned as `xfail(strict=True)` with the reproducer baked in as an `@example`, so it cannot start passing unnoticed. I have not fixed it here: it changes user-visible scheduling and wants its own change, its own bullet and its own beta. Say the word if you would rather I fold the fix in.

Also corrected: `AGENTS.md` claimed the unit tests need only `pip install pytest PyYAML`. `Babel` has been required for a while — a bare install fails on an import, not a skip — so that line was already wrong before this branch. It now lists all four and says each one skips cleanly on its own, which I verified in a clean venv: **2159 passed / 4 skipped** without hypothesis, **2173 passed** with it.

Runtime, measured rather than estimated: whole unit suite **7.5 s** under the `ci` profile, slowest single property **0.29 s**. Two consecutive `ci` runs and an explicitly-seeded third all agree, so `derandomize` is doing its job and a red run will mean a real defect rather than an unlucky seed.

---

/q review Fifth pass, on the property-based tests in `d2f4b38` and `a4d2914`. Be critical and skeptical; most serious first, and do not withhold minor ones.

**Are the properties true, or merely passing?** This is the main question. A property that is too weak passes forever and proves nothing; one that is too strong gets quietly widened until it does. Please attack each of the 14 for both failure modes. In particular: is `test_r1`'s reference implementation genuinely independent of the code under test, or does it share enough of `_step` to be a tautology? Does `test_r3`'s month-index arithmetic just restate `add_months`? Is `test_t3`'s `has_cycle` helper an independent oracle or a copy of `would_create_cycle`?

**The scoping decisions.** Three properties are deliberately narrowed: R4 excludes February 29, R1 bounds the reference walk with `assume()`, and T2a uses distinct timestamps. Each is defended in a docstring. Is each narrowing honest, or is one of them hiding a defect the way the Feb 29 case would have been hidden if I had widened R4 instead of splitting it?

**The mutation-gate interaction.** Property tests are scored rather than deselected. The `mutmut` profile drops `Phase.shrink`, `Phase.target` and `Phase.explain` and sets `database=None`. Is that reasoning sound — specifically, can a mutant still be *reported* usefully with no shrinking, and is there a path where the forked children still share Hypothesis state despite `database=None`? Also: `HK_HYPOTHESIS_PROFILE` is not part of mutmut's `config_fingerprint`, so a locally cached `mutants/` survives a profile switch. Is that a trap worth guarding rather than documenting?

**The strategies.** `property_strategies.py` builds through `models.build_task` and `assets.build_asset` on purpose, so it cannot generate an invalid record. Does that leave a class of bug untested that matters more than the drift it avoids? And is `parent_graphs()`'s shape (one parent edge per appliance, `unique_by` the child) capable of producing every cycle topology the real code can see?

**The February 29 finding.** Is my reading right that `_next_season_start` clamps and `in_season` does not? If so, which is the correct behaviour to change — and is the same asymmetry present anywhere else that takes a `MM-DD` boundary?

---

prestomation:
/q review Be critical and sceptical. Six commits since the last review turned the document into YAML and added a published JSON Schema; surface the most serious problems first and do not hold back minor ones.

**Correctness — the YAML dialect.** `transfer.py` now has `document_to_yaml` and `parse_document`. The loader is a `SafeLoader` subclass with the implicit `timestamp` resolver removed and `compose_node` raising on an alias. Attack the round trip: is there a value `build_document` can produce that the dumper writes and the loader reads back as something else? I claim PyYAML's emitter quotes every string that would resolve to another type — find a counterexample if one exists (sexagesimals like `1:30`, `0o17`, `.inf`, `~`, a key that is not a string, a very long line under `width=100`). Is dropping only the timestamp resolver enough, or does removing it from `yaml_implicit_resolvers` per first-character bucket miss a path? Is keeping the YAML 1.1 bool resolver the right call given the exporter quotes but a hand-written file does not?

**Security.** `parse_document` refuses anchors to close billion-laughs, and caps input at `MAX_IMPORT_BYTES` (8 MiB) before parsing. Is refusing at `compose_node` actually reached before expansion? Is 8 MiB under Home Assistant's websocket message limit, or does an oversized paste fail opaquely at the transport instead of returning a `Problem`? Anything else a hostile document can do to `yaml.load` with this Loader — deep nesting (I catch `RecursionError`), huge keys, quadratic blowup without aliases?

**The schema's fidelity.** `ci/generate_schema.py` converts `TRANSFER_DOCUMENT_SCHEMA` with `voluptuous_openapi`. Is there a place where the generated schema is *stricter* than `plan_import`, so a document that would import cleanly gets flagged in someone's editor? I know of one (`interval: "3"` coerces but fails `{"type": "number"}`) and argue it is acceptable because no export writes it — push back if you disagree, and look for others. Conversely: anything the schema accepts that the importer rejects? Is `additionalProperties` open everywhere it needs to be?

**Maintainability.** `TRANSFER_TASK_RECORD_SCHEMA` is a new declaration in `__init__.py` built by filtering `ADD_TASK_SCHEMA.schema` — is that the right place, and does `test_generate_schema.py`'s coverage check really stop it drifting from `models.build_task`? `_entry_schema` lifts field types out of `COMPLETE_TASK_SCHEMA` by name; is that too clever? `_SEASON_WINDOW_SCHEMA` is the one shape spelled out rather than inherited — is that restatement I should have avoided?

**HA best practice.** PyYAML is imported lazily in `transfer.py` and deliberately **not** added to `manifest.json`, on the grounds that Home Assistant cannot boot without it (same class as `voluptuous`). Is that right for a custom integration, or should it be declared? `ci/install-schema-deps.sh` installs `voluptuous-openapi` under HA's `package_constraints.txt` — is there a better way to track the version HA is built against?

**Performance.** `plan_import` now parses inside itself when handed a string, and the panel sends text on every Preview. Any concern about doing that on the event loop for a large document?

---

prestomation:
Thanks — the one substantive question was the right one to ask, and the answer was worse than I'd assumed.

> verify no other coercion validators exist in `ADD_TASK_SCHEMA`/`ADD_ASSET_SCHEMA`

**Four cases, not one**, and two of them mattered. `vol.All(cv.ensure_list, [X])` accepts a bare X *or* a list of them, and `labels: kitchen` is the obvious thing to write by hand — but `voluptuous_openapi` renders only the list form, so an editor would have flagged a document that imports cleanly. That is the one failure a published schema must not have, and it hit `labels` and `completion_required_fields`. `_serialize` now renders that pattern as `anyOf [X, array of X]`, matching what the action really takes. Pushed in 3cca803.

The other two are the coercions (`interval`, a completion's `cost`), and those stay stricter on purpose: no export writes a number as a string, so a squiggle on `interval: "3"` is a service rather than a false alarm.

Closing the `ensure_list` gap surfaced a **pre-existing inconsistency** rather than causing one. `ADD_TASK_SCHEMA` wraps `card_links` in `cv.ensure_list`, so the action accepts a single object — and `models.normalize_card_links` then refuses it with `card_links must be a list`. So `home_keeper.add_task` with `card_links: {…}` passes voluptuous and fails in the normalizer today, independently of this PR. The schema follows the action, which is what the README promises a record is; being lax there costs a warning the preview still gives with a path, while being strict would flag documents that import.

All three divergences are now a parametrized table that asserts each one is *still* a divergence, so a fourth is a decision somebody made rather than something a user finds.

Two corrections to your read, both my fault for what the code said at the time:

- **"the unit lane deliberately does not install Home Assistant"** — I wrote that in a comment and it is wrong. `ci/install-deps.sh` installs `pytest-homeassistant-custom-component`, so HA *is* importable there, on whatever release pip backtracked to for that job's Python — and it was old enough that `LOVELACE_DATA` was gone and the integration would not import. Sixteen tests failed for a reason unrelated to the schema. The gate is now an explicit `HK_SCHEMA_GATE` opt-in set only in the mypy job, which pins a Python at HA's floor and runs `check-ha-version.py`; that step greps the pytest output so an opt-in that stops running fails instead of passing quietly.
- **8 MiB vs the websocket limit** — you flagged this as "worth verifying in practice" and you were right to. Still unverified; I have it as a follow-up rather than a claim.

Mutation gate: **85.45%** (1809/2117), threshold 80%.

---

prestomation:
The walkthrough sticky comment says the capture failed on this push. **It is not the tour** — I ran both tours locally against the same seeded container and they pass:

```
✓ 1 [desktop] › walkthrough.capture.ts:1055 › record Home Keeper panel walkthrough (desktop) (3.5m)
✓ 2 [desktop] › walkthrough.capture.ts:1055 › record Home Keeper panel walkthrough (phone)   (34.8s)
2 passed (4.2m)
```

Three things point the same way. The failing run reported `HAS_DESKTOP: no` and `HAS_PHONE: no`, so nothing was captured at all rather than a beat going wrong partway. The only edit I made to `walkthrough.capture.ts` was the pasted document, and both copies of it parse and plan cleanly through the real `plan_import` (`ok=True`, no problems). And the same transfer flow is covered by the e2e suite, which is 298/298 locally on this head.

Leaving it rather than pushing an empty commit to re-run it — it is a soft gate and the next real push will re-capture. Flagging it so the stale "capture failed" note is not read as unverified.

---

prestomation:
## Exploratory testing pass (second author)

`main` is merged in (the branch was conflicted). Then a pass over the planner and a
pass with a real browser against a real Home Assistant. Eight defects, each of which
imported cleanly and said nothing.

### Found by probing `plan_import`

- **Two records in one document could claim the same `external_id`.** Nothing checked
  it, so the import wrote two records answering to one key, and every later run of the
  same document then failed with "matches several" — unrepairable by re-running the
  script `external_id` exists to make re-runnable. Now an error naming the key and both
  records. `_Matcher.claimed` already stopped two records upserting onto one *stored*
  record; this is the other half.
- **An unknown key inside a `history` or `skips` entry was dropped in silence**, while
  the same key on the record drew a named warning. A spreadsheet migration writing
  `notes` for `note` lost the column with nothing said. README already promised "the
  preview names every field that Home Keeper did not read".
- **An `area` name matching no area was stored as a literal area_id, with no warning.**
  The panel renders an unknown area id as its own text, so the record reads as filed in
  that room while Home Assistant has no such area. The name is still kept. It is now
  reported, the way an unresolvable `device_id` already was.
- **`format: yes` passed for format 1.** A bool is an int in Python and YAML 1.1 reads
  a bare `yes` as one. A non-integer format now gets its own message, because "Update
  Home Keeper" answers a later version and answers nothing for a value that is not one.
- **Every syntax error ended "Check the indentation."**, so an anchor refusal read
  "cannot use an anchor or an alias. Check the indentation."

### Found by clicking through the panel

- **A document over 4 MiB dropped the websocket connection.** Home Assistant builds its
  `WebSocketResponse` without passing `max_msg_size`, so aiohttp's 4 MiB default applies
  and the check is on the *decompressed* frame. Going over it is not a failed command:
  the connection closes (`Decompressed message exceeds size limit 4194304`), the card
  could only show "Something went wrong. Please try again.", and pressing the button
  again dropped the connection again.

  `MAX_IMPORT_BYTES` is 8 MiB and its comment said 8 MiB "stays under Home Assistant's
  own websocket message limit". It is twice that limit, so the ceiling and its "split
  the migration" message were unreachable from the panel. Measured on the running
  container: 3.9 MiB imports, 4.2 MiB kills the socket.

  The service is unaffected and the 8 MiB number is right there — measured the same
  way, 7.9 MiB imports and 8.2 MiB gets the correct message. So the two ceilings are
  now distinct and both are mirrored in `limits.ts`, with the panel refusing the send
  in the browser and naming the way in that still works.

#### The refusal, at both widths

The Import and export card showing a red alert that the document is larger than the panel 4 MB limit and naming the import_data service, with Import greyed out

The same refusal at phone width, the message wrapping to four lines inside the card above the Preview and Import row

- **Import stayed enabled after a completed import.** A finished import leaves a clean
  report behind, and that was read as permission. A second press ran the document
  again. Harmless for a document keyed by id or `external_id`, which upserts — but a
  document carrying two records of one name creates both on the first press and cannot
  tell them apart on the second, so pressing a lit button twice returned an ambiguity
  error about records the first press had just made. Import now also requires the
  report to be a dry run, which is what the PR already claimed.

- **The error alert outlived its text.** It is drawn beside the report rather than
  inside it, and only the report was cleared when the document was edited. A refusal
  stayed on screen while the reader pasted a smaller document.

### Checked and correct

Problem rows escape markup (an `img` tag with an `onerror` handler, in a name and in a
field key, both came back inert). The two-step gate holds — editing one character
revokes Import. Export downloads `home-keeper-<date>.yaml` with the schema modeline,
and pasting that file straight back reports every record as an update and no problems.
Folding history is idempotent across three runs. The record cap is exact at 2000/2001.
Unicode, RTL and emoji names, 3000-deep flow nesting, a null name, a negative interval,
and a section that is a mapping rather than a list all come back named. Forty problem
rows render with no horizontal overflow at 390px.

### Reported, not fixed

An unquoted `name: no` creates a task named **`False`** — YAML 1.1 reads it as a
boolean and the record keeps Python's repr of it. README documents the quoting rule, so
whether a boolean in a text field should be refused outright is a call for the
maintainer rather than a bug I should decide.

### Gates

`pytest tests/unit` 2207 passed (the 7 `test_config_flow` / `test_companions_registry`
failures are this sandbox having no real Home Assistant, and fail identically on
`origin/main`). `npx vitest` 1341 passed. `tsc --noEmit`, `ruff check`, `ruff format`
clean. `vale` clean on `README.md`, and the merge added no new `CHANGELOG.md` prose.
Python mutation gate **86.88%** against an 80% threshold, so the checklist's last
unticked box is now green.

No `CHANGELOG.md` entry: none of this ever shipped, so the feature's own `## [0.23.0b1]`
bullets still describe what a user gets.

---

prestomation:
/q review Be critical and skeptical, and surface the most serious problems first without withholding the minor ones. Two new commits sit on top of the import/export work: a backend pass on `transfer.py` (an `external_id` collision check, named warnings for unknown keys inside a `history`/`skips` entry and for an unresolvable `area`, a stricter `format` check, a reworded syntax error) and a panel pass (`limits.ts` + `panel.ts` refusing a document larger than Home Assistant's websocket frame limit, `panel-settings.ts` requiring a *dry-run* report before Import is enabled and clearing a stale error alert).

Please scrutinise:

- **Correctness of `_colliding_external_ids`.** It runs after each section is planned and drops the offending records. Does dropping an appliance leave anything dangling that `_looping_parents` handles and this does not — a task reference, a parent link, a `planned_assets` entry? Is there a case where two records legitimately share an `external_id` and this now refuses a document that used to work? Consider `match: none`, and a record that inherits an `external_id` from the stored record it matched rather than stating one.
- **The two size ceilings.** `MAX_IMPORT_BYTES` (8 MiB, service) and `MAX_IMPORT_WS_BYTES` (4 MiB, panel) are now different numbers for the same feature. Is that split defensible, or should the service come down to match so there is one number a user has to know? `importFrameBytes` adds a flat 128 bytes for the command envelope and measures `JSON.stringify` output via `Blob` — is that a sound proxy for what aiohttp weighs, given permessage-deflate compresses the frame and the limit is on the *decompressed* size? Am I refusing documents that would actually have fit?
- **HA best practices.** Is depending on aiohttp's 4 MiB default the right call, or is there a supported way to learn the limit at runtime? If Home Assistant ever passes its own `max_msg_size`, does this fail safe?
- **The `ready` gate change** (`report.dry_run` now required). Does that strand any legitimate flow — a user who previews, imports, and genuinely wants to re-apply the same document?
- **Maintainability.** `transfer.py` gained a fourth post-pass concern; is `plan_import` getting too long, and should the collision and loop passes share a shape? Is `limits.ts` the right home for `importFrameBytes`?
- **Anything the new tests assert that does not actually pin the behaviour** — I strengthened several assertions to kill surviving mutants, and I would rather know if any of them are now over-fitted to a message string.

---

prestomation:
### Gate results on the final head

Full Playwright suite: **310 passed**, on a container rebuilt from the committed
fixture. Worth stating how that number was arrived at, because an earlier run of mine
was not trustworthy: I killed a suite mid-flight to free CPU, its teardown never ran,
and the next run then failed `documents.spec.ts` (3 file rows where it expects 2) and
`list-honesty.spec.ts` (the shopping pill not empty, against a leftover *Buy Anode
rod*). Both were residue from the killed run rather than regressions. Resetting
`tests/integration/ha_config/` and rebuilding the container cleared both, and the
committed fixture is back to its 16 tasks and 3 appliances.

TypeScript mutation gate: **100.00%** (12/12). It was 83.33% on the previous head, with
two `ArithmeticOperator` mutants alive on `MAX_IMPORT_BYTES` — nothing in the frontend
read its value, because the panel only ever divides it into a message string.
`test_upload_limit_parity.py` pins it against `const.py`, but it cannot see the number a
reader actually gets, so `limits.test.js` now pins both ceilings and their ordering
(`cf1b5d5`).

Python mutation gate **86.88%**, `pytest tests/unit` 2207 passed, `npx vitest` 1353
passed, `tsc --noEmit` / `ruff` / `vale` clean.

---

prestomation:
### A YAML boolean is not text (`b49aa1e`)

The one I had left for you to decide is now fixed. Where it belonged turned out to be
the interesting part, because the two obvious places were both already right:

**Not the loader.** `transfer.py` keeps YAML's boolean resolver deliberately, and has
to. `enabled`, `archived` and `require_tag_scan` are real booleans and `enabled: no` is
what a writer of Home Assistant YAML expects. A resolver is global rather than
per-field, so dropping it would make `"no"` a truthy string, which fails in the more
surprising direction.

**Not the schema.** `ci/generate_schema.py` maps `cv.string` to `{"type": "string"}`
and `cv.boolean` to `{"type": "boolean"}`, so the published JSON Schema has always
typed `name` as text and `enabled` as a boolean. An editor validating the document
already flagged `name: no`.

The runtime was the half that disagreed with the contract it publishes:
`str(False)` is `"False"`, and that is what landed. So `_reject_boolean` sits in the
pure core beside the other normalizers, and the text fields refuse a boolean rather
than coercing it — a task's `name` / `notes` / `external_id`, an appliance's `name` /
`manufacturer` / `model` / `serial_number` / `notes` / `external_id`, a part's `name` /
`part_number` / `vendor` / `notes` / `stock_unit`, and a document's name. The message
names the field and the fix, because the file looks right to whoever wrote it.

Only booleans. A number is still coerced, so an unquoted `part_number: 4711` keeps
working. It is private rather than public, so it adds nothing to the surface the
one-way-doors list already covers.

**The gate this belonged in.** `test_generate_schema.py` has a table asserting the
schema and `plan_import` agree, and a second naming every place they deliberately
disagree. `name: false` was in neither: an unlisted divergence, which is exactly the
shape of bug that table exists to catch. It now has three rows — a task name, an
appliance's `manufacturer`, and `enabled: false` for the other half of the split.

25 tests, including that `enabled: no` / `archived: yes` / `require_tag_scan: no` still
read as booleans, and that an export of a task named `no` round-trips (the exporter
quotes it, so Home Keeper's own files never trip the new rule). README updated.

Two caveats worth stating rather than burying. `test_generate_schema.py` needs a real
Home Assistant, so it skips on this sandbox's Python 3.11 — I verified the
`plan_import` half of all three new rows directly and the `lint.yml` mypy job runs the
schema half. And the Python mutation gate is still running locally at the time of
writing (84% killed against an 80% threshold, but not finished); CI runs the same gate.

---

prestomation:
/q review Critical and skeptical please, most serious first, and do not withhold the minor ones. New commit `b49aa1e` adds `_reject_boolean` to `models.py` and `assets.py`: a text field now refuses a YAML boolean instead of coercing it through `str()`, which was storing `"False"` as a task name.

Scrutinise in particular:

- **Field coverage.** I guarded a task's `name`/`notes`/`external_id`, an appliance's `name` and its `_ALL_TEXT_FIELDS` loop plus `external_id`, a part's `name`/`part_number`/`vendor`/`notes`/`stock_unit`, and a document's name. What did I miss? `metadata` entries (label/value), `card_links`, `task_chips`, `unit`, `freq`, `anchor`, `due`, `completion_detail`, `recurrence_type`, `area_id`, `device_id` all still take `str()` or a bare `.get()`. Which of those can a hand-written YAML document reach with a bare `no`/`on`, and does leaving them coerce create the same silent-garbage class of bug?
- **Is refusing right, or should it warn?** Every other unknown/unusable value in an import is a *named warning* and the record still lands. This makes a whole document fail. Is a hard error the right severity given `plan_import` is all-or-nothing, or would a warning that drops the field be truer to the module's stated forward-compatibility posture?
- **Blast radius beyond import.** `models.build_task` and `assets.build_asset` are also the service path and are called by other integrations (Pawsistant, Battery Notes). On the service path `cv.string` coerces a bool to `"True"` *before* models sees it, so the guard should be unreachable there — is that actually true for every caller, including the websocket commands and `merge_update`? Have I broken a contributor who passes a real bool today?
- **`isinstance(value, bool)` placement.** Note `_require` treats `False` as present (it only rejects `None` and `""`). Is there an ordering bug between `_require`, `_reject_boolean` and the `or ""` fallbacks I added in `assets.py`, e.g. somewhere `False` now becomes `""` silently instead of raising?
- **The schema/importer agreement table.** I added three rows to `test_generate_schema.py`. Are they correct — does `voluptuous_openapi` really emit `{"type": "string"}` for `name` such that `false` is invalid, and `{"type": "boolean"}` for `enabled` such that `false` is valid? I could not execute that file locally (no Home Assistant on this Python), so this is the assertion I am least able to vouch for.
- **Message wording** in an exception a user reads in the import preview.

---

prestomation:
Python mutation gate on the final head: **86.63%** (2541 detected / 2933 scored, threshold 80%), with nothing left alive in `_reject_boolean`.

Three mutants did survive the first run of it, all on the second half of the refusal message. The assertions stopped at `"quotation marks"`, which a mutant that rewrote the tail still satisfied — so the part of the string whose entire job is to teach the fix was the part nothing checked. `70962f1` pins the whole sentence in both modules, and the re-run killed all three.

That closes the last unticked box on the checklist in the description.

---

prestomation:
CI was red on the previous head and I put it there. Both causes were local gates that were quietly weaker than CI's, so this is worth recording rather than just fixing.

**The property test found a real defect, not a test artifact.** `test_t1` generated a store holding two tasks under one `external_id` — which nothing upstream prevents — exported it, and could not import the result. My collision check refused the *shape* rather than the *damage*, so an export of such a store became impossible to restore. Restoring a backup is the job this format exists for, so that was the more serious bug of the two.

The rule now refuses a shared key only when at least one record in the group would actually fall through to it. A record stating its own uuid is matched at step 1 of the ladder and never reaches step 2; an export writes an `id` on every record, so it round-trips. A migration script that writes `external_id` and no `id` — the case the check was added for — is still refused, and so is a mixed group, because the record without an id is the one that would go missing.

It skipped here because `hypothesis` and `jsonschema` were never installed in this sandbox. `AGENTS.md` names both; I had installed three of the five packages and did not notice two test groups going quiet. They are in now, and the local suite matches CI's shape.

**vale** flagged the new locale string under `ai-tells.MicDrop`: "This document is larger than the panel's {mb} MB limit" matches its comparative-sentence pattern. I had run vale on `README.md` and never on `en.json`, which is also in the linted set — the trap `AGENTS.md` describes. The string now reads "is over the panel's limit", checked against all 77 rule files' regexes directly and with a vale pass over the whole linted set rather than the one file I edited.

Gates on this head: `pytest tests/unit` 2251 passed (the 7 `test_config_flow` / `test_companions_registry` failures are this sandbox having no real Home Assistant and fail identically on `origin/main`), property tests green under the `ci` profile, `npx vitest` 1347 passed, `tsc`, `ruff`, `vale` clean, Python mutation **87.12%** (2564/2943, threshold 80%) with nothing left alive in `_colliding_external_ids`.

---

PR #310, opened 2026-09-08
fix(panel): keep companion row buttons on screen, and guard profile delete

prestomation:
Fixes #

## What changed

Two independent Settings fixes, from three reports.

### 1. Companion rows on a phone

> "add a delete button for the declarative companions"
> "the edit button for the declarative companions is covered by the preset chip on mobile"

These are one bug. **The declarative-companion row has had a Delete button since #231** (`panel-declarative.ts:266`, wired to `openConfirmDialog` + `api.deleteDeclarativeCompanion`, with an e2e test and a committed shot). It was simply off the screen.

`.hk-companion` kept its desktop single-line flex layout below 700px. The body box shrinks (`min-width: 0`) but nothing clips its content, and an `ha-assist-chip` will not go below its own label — so the Enabled and `Preset: device_pulse` chips painted out over the action column, and the buttons (`flex: 0 0 auto`) ran off the right edge with nothing to scroll to reach them.

The row now stacks into a grid inside the 700px block: icon and text on top, buttons on a full-width line beneath, at `--hk-tap` height. Applied to `.hk-companion` generally rather than `.hk-decl-row` — a *suggested* companion carries three buttons (Install, Dismiss, Docs) and overflowed worse. `.hk-companion-name` also wraps at every width, since the Settings column is narrow below 1000px too.

Two notes for review:

- This finally makes the `flex-wrap: wrap` already on `.hk-companion-actions` do something. It could never engage while the row was one flex line — `flex: 0 0 auto` pins the container at max-content.
- **Desktop is untouched.** Every new rule is inside `@media (max-width: 700px)`; the desktop shot below is there to prove nothing moved.

A recipe row in Settings, Companions on a desktop: name with Enabled and Preset chips, Edit and Delete on the right

The same recipe row on a phone, with the preset chip wrapped onto its own line and Edit and Delete on a line beneath

### 2. A guard on deleting a profile

> "delete profile needs to validate that there are no notifications using it"

A notification names a profile by `profile_id`. Deleting the profile used to succeed silently; `notifier._notification_profile` then read the dangling id back as "send nothing", and the notification went quiet with nothing on screen to say why.

`options.async_set_options` now refuses a save that removes a profile while a notification surviving the **same save** still names it — so the panel, the `home_keeper.set_options` service and the websocket command are all gated by one check. `ServiceValidationError` for the service, the eager `_err` helper for the websocket, both resolving `exceptions.profile_in_use`.

The guard is on the **write path only**. `_normalize` / `current_options` are also the *read* path, and a dangling `profile_id` stays reachable by design — a backup restore, or a service call that writes `notifications` alone. `notifier.py`, `todo_list.py` and `docs/PROFILES_REFACTOR_PLAN.md` all depend on that state being readable. Only a profile *this save removes* counts, and the references are read from the merged document, so deleting a profile together with its notifications is still allowed (three e2e specs tear down exactly that way).

In the panel, Delete now opens a dialog instead of deleting on the press — profile delete had no confirmation at all before. `openConfirmDialog` keeps its signature; the new `openBlockedDialog` shares the scrim with `onConfirm: null`, which drops the Delete button and relabels Cancel to Close.

The panel refusing to delete a Profile and naming the notification that uses it, with only a Close button

The ordinary confirmation for a Profile nothing uses, with Cancel and Delete

The same blocking dialog on a phone

## One-way doors

| Surface | Shape | Where it appears | What relies on it |
| --- | --- | --- | --- |
| `exceptions.profile_in_use` | translation key | `strings.json` + 16 `translations/*.json` | HA's lazy `translation_key` resolver and `backend_i18n.resolve_exception`. Renaming it breaks every translation. |
| `{profiles}` / `{notifications}` | placeholder names | the same message | `test_translations_parity.py` pins the token set per key, so a rename is a 17-file change and re-resolves on every translation platform. |
| `profile_in_use` | websocket error **code** on `home_keeper/set_options` | the wire | `api.ts` reads only `.message` today, but a third-party client can branch on the code. Note `api_surface.py` has no registry for websocket error codes — a pre-existing gap, not introduced here. |
| **`home_keeper.set_options` now rejects calls it used to accept** | behaviour | the service, the websocket command | **The real door.** An automation or restore script that rewrites the whole `profiles` list will start raising. It fires only when the same call leaves a notification naming a profile that call removed; sending `notifications` in the same call clears it. |

One edge worth calling out inside that last row: a call that sends `profiles` **without the `id` fields** is now refused. `profiles.normalize_profile` mints a fresh uuid when `id` is absent, so such a call reads as remove-then-add — and that is precisely the call that used to strand a notification silently. Intended, but it is a narrowing.

Not one-way: `openBlockedDialog`, `notificationsUsingProfile`, the `persistOptionList` options bag, `ProfileInUseError` (never crosses the process boundary — both callers convert it), and every CSS rule in change 1.

## Notes for review

- **`tests/unit/test_options_write.py` is new and load-bearing.** `async_set_options` had **zero** unit coverage, and `ci/mutation_scope.py` pulls the whole enclosing function into scope — so every mutant in it (the no-op short-circuit, the `_CALLER_RELOADING` add/discard, the `try/finally`) would have survived the 80% gate. It was more work than the guard itself.
- **`profile_removals_in_use` indexes rather than defends.** Both its arguments come from `current_options` / `_normalize`, which fill every key and give every profile an `id` and a `name`. The first draft used `.get()` defaults and `or` fallbacks; they were unreachable branches, mutmut scored them as surviving, and a missing key silently meant "nothing is in the way" — which would write the save through. It now raises `KeyError` instead. That took the score from 80.72% to 100%.
- **The profile-dialog shots live in a new standalone harness** (`screenshots-profile-delete.capture.ts`), not in `screenshots.capture.ts`. That harness walks every surface in one test and its shot 0 asserts the first-run banner, which the committed fixture has dismissed — so a step appended to it can never be captured. Same reasoning `screenshots-declarative.capture.ts` already gives.
- **Two shots were renamed to get past the PR-body `` mangling.** The API backtick-wrapped the src of the two phone images, twice running, while their neighbours went through clean. That is the #298 failure AGENTS.md records, where the mangling follows the file and only a rename clears it.
- **`mypy` was not run locally.** HA requires Python ≥3.14.2 and this container maxes at 3.13, so `pip install homeassistant` fails. `mypy --strict` reports **0 errors in `options.py`**, and no error lands on any line changed in `__init__.py` or `websocket_api.py`. CI is the real check here.
- Local runs: 1997 pytest unit, 1298 vitest, 289 e2e (all three projects), ruff clean, mutation 100%, `vale` adds zero new hits to `CHANGELOG.md` or `README.md`.

## Checklist

      step through it — **n/a**: one bug fix and one guard, no new surface in the tour
      `const.py`), `preview-release` label applied — bumped to `0.22.0b5` (`0.22.0b4`
      is already a published tag, so `changelog-release-gap` forces a new section)

https://claude.ai/code/session_01MhMamDcznsqv1ecEdACoVk


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first without withholding the minor ones. This PR carries two unrelated changes; scrutinize both.

**Correctness of the profile-in-use guard (`options.py`).** `profile_removals_in_use` diffs `base[OPTION_PROFILES]` against `merged[OPTION_PROFILES]` for removals, then reads references out of `merged[OPTION_NOTIFICATIONS]`. Try to break it: a save that sends `profiles` but not `notifications` (and the reverse); a profile re-sent without its `id` (which `normalize_profile` re-mints, so it reads as remove-then-add — intended, but is anything else caught in that net?); `profile_id` of `None`, `""`, or a non-string; duplicate profile ids in one list; a profile whose id collides with one being added in the same save. It deliberately indexes rather than using `.get()`, on the argument that both documents always come from `current_options`/`_normalize` — is that reachability claim actually true for **every** caller, including `merge_flow_input` and anything in the options flow?

**The read/write split.** The guard sits in `async_set_options` only, after the `merged == base` short-circuit and before `async_update_entry`, so an options document that *already* holds a dangling `profile_id` stays readable and writable (`notifier.py`, `todo_list.py` and `docs/PROFILES_REFACTOR_PLAN.md` all depend on that). Is there any path that reaches a write without passing this function? Can a refused save leave `_CALLER_RELOADING` dirty, or the entry half-updated?

**HA best practices.** `ServiceValidationError` vs `HomeAssistantError` for the service path; catching `ProfileInUseError` inside `_with_coordinator` beside the two store exceptions rather than inline in `ws_set_options`; whether the websocket error **code** `profile_in_use` should be modelled anywhere in `api_surface.py` (it currently is not, and neither are the existing codes).

**Frontend.** `persistOptionList` grew a `rollbackOnFailure` option that restores a snapshot under an `isNewest()` guard — is that snapshot correct against concurrent debounced saves of the same list, and is the deliberate *no*-rollback for edits still intact? The panel pre-check `notificationsUsingProfile` duplicates the backend predicate in TypeScript; is that duplication a liability worth a shared fixture? `openConfirmDialog` and `openBlockedDialog` now share one scrim renderer keyed on `onConfirm === null` — check the Escape/teardown paths and `_syncDrawerModality` for either mode.

**Performance and maintainability.** The guard runs on every options save; is it a hot path worth caring about? And the CSS: `.hk-companion` becomes a grid below 700px while its children keep flex properties that are inert inside a grid — is leaving them a maintenance trap, and does the change reach any surface beyond Settings → Companions?

---

prestomation:
Thanks — three of the five landed, pushed in b1ed674. Two I'm pushing back on, with reasoning.

**Fixed: the `profile_id: None` fall-through.** Agreed, and the framing was the useful part: `None` is a *valid* value (the notification covers every due task), not malformed input, so the "we index because everything is normalized" justification doesn't cover it. Anyone hardening the lookup into a string check would turn every all-tasks notification into a blocker. I kept the `removed.get(...)` form rather than taking the suggested `is not None and ... in removed` — the extra branch is the thing a future reader would be tempted to "simplify" back — and instead made the comment say why it must fall through, and name `test_a_notification_with_no_profile_is_never_a_blocker` as the test that pins it.

**Fixed: the inert flex declarations.** Real trap, documented in place. I didn't scope them into a `@media (min-width: 701px)` block as suggested: that duplicates the rule and puts the desktop layout one specificity change away from breaking, which trades a comment-shaped problem for a cascade-shaped one.

**Not a defect: the same-row rollback race.** `rollbackOnFailure` is set at exactly two call sites — `deleteProfile` (`panel-settings.ts:1238`) and `deleteNotification` (`:1465`) — and both call `persistOptionList` **directly**. A typed edit goes through `persistDebounced`, which passes no options, so `rollbackOnFailure` is `undefined` on every edit path. The scenario described ("the user typed into one row twice in under 600ms") therefore cannot reach the rollback at all; the deliberate no-rollback-for-edits behaviour is intact, which is what the last test in `profile-delete.test.js` guards.

Two overlapping *deletes* are also safe, for a reason worth stating: `previous` is captured after the preceding delete's optimistic write, not before it, because `persistOptionList` writes `p._options` synchronously before it awaits. So the second delete's snapshot already has the first row gone. `isNewest()` then stops a superseded save writing anything. That was an argument rather than a test, so it is now a test — "does not roll a second delete back over a first one that already landed".

**Not a defect: the test-name nitpick.** `test_the_caller_reloading_flag_is_set_during_the_reload_and_cleared_after` asserts both halves. The callback proves *set-during* (which can only be observed from inside the reload), and the line after `_set(...)` returns proves *cleared-after*. The name is accurate.

**Agreed and not actionable here:** the websocket error code not being modelled in `api_surface.py` is pre-existing for every code, and `_CALLER_RELOADING` is correct because a refusal happens before the id is ever added.

Local validation on the new head: 1299 vitest, 1997 pytest, ruff and tsc clean, mypy 0 errors in `options.py`, and the responsive-layout specs re-run against a real browser — the CSS change lives inside a template literal, and the first draft of that comment used backticks and silently broke the whole stylesheet, so it needed a real render rather than a typecheck.

---

PR #314, opened 2026-09-09
fix(panel): say what a notification sends, and head its switches Triggers

### LINKED ISSUE #313: Non-overdue tasks on notification digesy
Hi again !

I setup a notification digest to see my overdue tasks in a nutshell, but I notice that tasks that are not due for a long time are included in the message 

Image
Image

prestomation:
Fixes #313

## What changed

The reporter turned on **Auto-send when overdue** and received a digest of tasks that were not due for months. The switch is only a trigger: `async_send_auto` (`notifier.py:311-326`) sends the notification once when a task crosses into overdue, and never decides the contents. The tasks come from the **profile's** Include field (`profiles.matches_filter`), which lives in a different Settings card. The notification editor said nothing that told the two apart, and the switch had no helper text at all.

Settings → Notifications now answers both halves where the question is asked.

**What this profile sends.** Under the Profile picker, a line states what the chosen profile selects, in the same words the Include field's own options use — `My chores includes overdue tasks only.` Beside it, **Edit this profile** opens that profile in the Profiles card, expanded.

**Triggers.** The two switches moved into a group headed **Triggers**, captioned *"Profiles determine which tasks a notification includes. Triggers determine when that notification is sent."* Their labels shorten from `Auto-send when overdue` / `Auto-send when due soon` to **Send when overdue** / **Send when due soon**, because the heading now carries the "auto". Under them, a link to the docs' automation examples, for a trigger the two switches cannot express.

**One free fix found on the way.** Nothing in the panel said that *due soon* means 3 days — `DUE_SOON_WINDOW` is a constant with no setting — so `notify.status_help` now names the number.

### Implementation notes

The editor builds as three `ha-form`s rather than one, so text can sit between them: `notificationProfileSchema` + `notificationDeliverySchema` + `notificationTriggerSchema`, concatenated by the existing `notificationSchema`, the same shape `problemSyncSchema` already has. Each half is seeded and merged back **by schema** via `pickFormData`, so one form cannot re-assert a stale snapshot of the others.

`openProfile` empties the Profiles host before re-rendering — `renderProfiles` appends its card, which costs nothing on a full render because the host is new each time, but out of band would leave a folded duplicate under the same id. It also navigates with `replace`, the way the rail's own entries do: pushing makes Home Assistant treat it as a fresh panel navigation and build a new element, which folds every row again including the one just opened.

### One-way doors

**None.** No option key, service field, event payload, entity attribute, or stored value changed. `auto: { overdue, due_soon }` and `filter.status` keep their names and values; the schema split is internal `ha-form` data. The 8 new locale keys and 3 edited ones are UI strings.

## Screenshots

Desktop — the scope line under the Profile picker, and the Triggers group:

Settings Notifications on a desktop, with the scope of each profile stated under its picker and the two switches under a Triggers heading

Phone — the two halves cannot share a frame at this width, so they are two shots. The scope line:

The same notification on a phone, showing the profile scope line under the Profile picker

…and the Triggers group, where the caption stacks under the heading:

The same notification on a phone, scrolled to the Triggers group with its caption and the link to the automation examples

## Tests

- `custom_components/home_keeper/frontend/test/notify-profile-scope.test.js` — new, 9 cases: the scope line for each stored status, its live repaint when the picker changes, silence when the profile is gone, the Edit link's navigation and expansion, the Triggers heading and caption, the docs link, and a trigger save that does not drop the other two forms' fields.
- `forms-schema.test.js` — the three-way split adds up to `notificationSchema`, each half owns a distinct complete set, both required fields stay in the first half, and `profileScopeKey` covers all 3 statuses plus the missing/empty/unknown fallback.
- `tests/e2e/tests/settings.spec.ts` — a new `@responsive` case asserts the scope line, the Triggers group, the docs href and the Edit link's navigation against the real panel, at all 3 widths. A screenshot cannot tell a rendered sentence from a missing one.
- `screenshots.capture.ts` — the phone block gains the 2 shots above.

Local runs: `bash ci/test-frontend.sh` 1315 passed · `pytest tests/unit -q` 1994 passed · `settings.spec.ts` 23 passed across desktop/phone/tablet · `bash ci/test-mutation-frontend.sh` **100.00%** on the changed `forms.ts` lines (threshold 80%) · `mypy` and `ruff` clean.

## Notes for the maintainer

- The CHANGELOG entry went into a new `## [Unreleased]` section. `main` is still on `0.22.0` — the post-release bump to `0.23.0b1` has not happened yet, and editing the released `0.22.0` section would fail `changelog-release-gap`. Rename `[Unreleased]` when that bump is cut.
- `i18n-parity.test.js` requires every English key in all 15 other locales and rejects an English-identical value, so the translations are part of this PR rather than a follow-up. Dutch `notify.triggers_heading` is genuinely "Triggers" — that locale already keeps the English loanword ("Trigger", "Triggermodus") — so it is added to `COGNATE_IDENTICAL` with a reason.
- `tests/unit/test_release_issues.py` pins the set of issues the CHANGELOG closes; 313 is added to it.

## Checklist

- [x] `CHANGELOG.md` updated — with `(Fixes #313)`
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML `` tag, desktop and phone
- [ ] New user-facing UI surface → not applicable, this is a fix to an existing surface, so the walkthrough is unchanged
- [ ] New user-facing feature → not applicable, bug-fix-only PR, no version bump and no `preview-release` label

https://claude.ai/code/session_013RhYFfGC4zKWCaAzE39rnM

---
_Generated by [Claude Code](https://claude.ai/code/session_013RhYFfGC4zKWCaAzE39rnM)_


prestomation:
/q review Please be critical and skeptical, and rank the most serious findings first without withholding minor ones.

This PR splits the Settings → Notifications editor from one `ha-form` into three (`notificationProfileSchema` / `notificationDeliverySchema` / `notificationTriggerSchema`), adds a computed line stating what the selected profile sends, and adds a link that navigates to that profile. Please scrutinise:

**Correctness of the three-form save path.** `notificationEditor` in `panel-settings.ts` keeps three `let` slices and merges them with `pickFormData(slice, schema)` on every change, through one `persistDebounced` key. Can a field be lost or resurrected? Consider: a user clearing a field so `ha-form` omits it from the emitted value; two forms changing inside one debounce window; a rename racing a trigger toggle; and whether `notifyFormToNotification`'s defaults can silently overwrite a real stored value when a key is absent from the merged record.

**The navigation side effects of `openProfile`.** It mutates `_settingsSectionCollapsed` and `_itemExpanded`, calls `host.replaceChildren()` then `renderProfiles`, then `_navigate(..., replace = true)`, then `scrollIntoView`. Is re-rendering one card out of band safe here — event listeners, `_liveHassEls` registration, the autosave status slot, a profile row that was mid-edit with a pending debounce? Is `replace` right, and does clearing the section's collapsed state without restoring it surprise anyone? What happens if the profile was deleted between render and click?

**`profileScopeKey` and the scope line.** `forms.ts`. Does the fallback to `overdue` match `profiles.normalize_filter` and `profileFormToProfile` in every case? The line is built with `escapeHTML` into `innerHTML` alongside a button — check the escaping and that repeated `repaint()` calls cannot leak listeners.

**HA best practices and a11y.** The Edit affordance is a `<button class="hk-linkish">` rather than an anchor, inside an `ha-form`-adjacent div. The Triggers group reuses `hk-indent` / `hk-eyebrow` markup built inline rather than through the `dependent` scaffold that `settingsToggleCard` uses — is that duplication worth factoring out?

**Maintainability.** Is the three-way schema split the right seam, or should the panel have kept one form and solved the text placement differently? Are the new locale keys named consistently with the rest of the `notify.*` namespace?

---

PR #315, opened 2026-09-10
ci: install every CI dependency at session start

prestomation:
## What changed

`ci/setup-ci-deps.sh` is a new idempotent installer for everything the CI workflows need. The Claude Code SessionStart hook now starts it in the background instead of `ci/setup-browser-env.sh`, which the new script calls as its last step. The log is `/tmp/setup-ci-deps.log`.

What it covers, and the lane each part serves:

| Dependency | Lane |
| --- | --- |
| `requirements-test.txt`, `pytest-homeassistant-custom-component`, `homeassistant` | `test.yml`, `pytest_coverage.yml` |
| `ruff`, `mypy` | `lint.yml` |
| `mutmut`, npm root (Stryker + vitest) | `mutation.yml`, `test.yml` |
| npm in `custom_components/home_keeper/frontend`, `tests/e2e`, `website` | panel build, `e2e.yml`, docs site |
| `vale` + the `ai-tells` styles | `lint.yml` vale job |
| `ffmpeg` | `ci/capture-video.sh` / `walkthrough-preview.yml` |
| Docker daemon + Playwright Chromium | `integration.yml`, `e2e.yml` |

Every step checks first and skips what is already installed, and no step can stop the run. The script always prints an installed/skipped/failed summary, so running it again retries only what failed. The exit status is 1 when a step failed, so a caller need not read the log.

```bash
bash ci/setup-ci-deps.sh          # install what is missing
FORCE=1 bash ci/setup-ci-deps.sh  # install everything again
# Leave one part alone:
SKIP_PYTHON=1  SKIP_NPM=1  SKIP_VALE=1  SKIP_FFMPEG=1  SKIP_BROWSER=1
```

Only one run at a time holds `.setup-ci-deps.lock`, since the hook starts the script in the background and two sessions can open together. The lock records its process id, so a run killed with `-9` or lost to a container restart does not disable setup for every session after it.

Pins are read from where CI already keeps them — `mutmut` from `mutation.yml`, the Python floor from `pyproject.toml` — so the script cannot go stale on its own. The `ai-tells` style needs no pin here: `.vale.ini` holds the release URL and `vale sync` reads it.

### Why the Python packages go in `.venv`

Home Assistant sets a Python floor that the system interpreter is usually below, and pip then resolves an old Home Assistant that does not always build or run. So the script makes `.venv` (already in `.gitignore`) and picks its interpreter **by test, not by name**: it tries `3.14` (the floor), then `3.13`, then `3.12`, and keeps the first that can run a unit test file. A stale Home Assistant imports fine on a new Python and then breaks at the first fixture, so a version-name check is not enough — on this machine 3.14 dies on `typing.ByteString` and 3.13 dies in the event-loop fixture, while 3.12 runs the suite green. When no interpreter meets the floor, the script says the resolved Home Assistant is old, in the log and again in the summary, and points at `ci/check-ha-version.py`.

`rm -rf` runs on the `.venv` path, so an overridden `VENV` is resolved with `realpath` and refused unless it lands inside the repository.

`AGENTS.md` gets a **Session setup** section covering all of this.

### Review

Amazon Q reviewed the first version and then the hardening. Both passes are triaged in the comments: the valid findings are fixed in `28b253b` and `58bd1c8` (the lock, the path guard, portable pin reading, a real vale version check, npm staleness against npm's own marker, the skip flags, the smoke-test fallback), and four are declined with reasons — chiefly the vale tarball checksum, which would be published in the same release as the tarball and so proves nothing.

### Verified locally

- Clean run from no `.venv`: 4 installed, 8 skipped, 0 failed. Second run: 0 installed, 12 skipped, 0 failed.
- `pytest tests/unit` — 2001 passed. `npx vitest run` — 1315 passed.
- `ruff check` and `ruff format --check` — clean. `vale README.md` — 0 errors.
- The lock refuses a live holder and reclaims a dead one; each skip flag works; a touched `package-lock.json` causes exactly one reinstall; `VENV=/`, `VENV=$ROOT/../elsewhere` and `VENV=$ROOT` are all refused, while `.venv` and a trailing slash are accepted.
- `mypy custom_components/home_keeper` — 1 error locally, and it is the stale-Home-Assistant artifact the script warns about (`LOVELACE_DATA` is absent from the older release), not a code problem. The `mypy` job passes in CI, where the floor is met.

## One-way doors

None. This adds a developer setup script and a local hook. It changes no service, event, storage or entity contract.

## Checklist

- [x] `CHANGELOG.md` updated — not needed, this is a developer-only change
- [x] Panel UI changed → not applicable
- [x] New user-facing UI surface → not applicable
- [x] New user-facing feature → not applicable

https://claude.ai/code/session_012o4F8BJzxo5kGZ6AiWASqC

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "nohup bash ci/setup-browser-env.sh >/tmp/setup-browser-env.log 2>&1 & disown; echo 'Home Keeper: preparing Docker + Playwright Chromium in the background (see /tmp/setup-browser-env.log). Run e2e with: bash ci/e2e-up.sh'"
+            "command": "nohup bash ci/setup-ci-deps.sh >/tmp/setup-ci-deps.log 2>&1 & disown; echo 'Home Keeper: installing the CI dependencies in the background (see /tmp/setup-ci-deps.log). This includes Docker and Playwright Chromium. Run e2e with: bash ci/e2e-up.sh'"
           }
         ]
       }
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ mutants/
 stryker.log
 reports/
 .venv/
+.setup-ci-deps.lock/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -518,8 +518,38 @@ siblings instead of their fakes.
   `/home-keeper` panel.
 - Run locally / in a session: `bash ci/e2e-up.sh` (builds the panel, starts HA, runs
   Playwright, tears down). `KEEP_UP=1` leaves HA running.
-- Env prep: `ci/setup-browser-env.sh` (wired to a Claude Code SessionStart hook).
+- Env prep: `ci/setup-browser-env.sh` (Docker plus the browser). `ci/setup-ci-deps.sh`
+  calls it and also installs the other CI dependencies — see "Session setup" below.
 - Auth: `tests/e2e/global-setup.ts` completes onboarding and performs a real login.
+
+## Session setup
+
+`ci/setup-ci-deps.sh` installs every dependency the CI workflows need: the Python
+packages, the npm packages for each of the four projects, vale and its styles,
+ffmpeg for the walkthrough capture, and (through `ci/setup-browser-env.sh`) the
+Docker daemon and Playwright Chromium. A Claude Code SessionStart hook starts it
+in the background; the log is `/tmp/setup-ci-deps.log`.
+
+The script is idempotent. Each step looks first and skips what is already there,
+so it is safe to run again — and it is the way to try a step that failed. No step
+can stop the script: it always prints a summary of what it installed, skipped and
+failed.
+
+```bash
+bash ci/setup-ci-deps.sh                  # install what is missing
+FORCE=1 bash ci/setup-ci-deps.sh          # install everything again
+SKIP_BROWSER=1 bash ci/setup-ci-deps.sh   # leave Docker and Playwright alone
+```
+
+**The Python packages go in `.venv`** (git ignores it), not in the system Python.
+Activate it before you run a Python lane: `source .venv/bin/activate`. The script
+picks the interpreter for that virtual environment by test, not by name: it tries
+`3.14` (the Home Assistant floor), then `3.13`, then `3.12`, and keeps the first
+one that can run a unit test file. An old Home Assistant imports on a new Python
+and then breaks at the first fixture, so a name check is not enough. When the
+machine has no interpreter at the floor, pip resolves an older Home Assistant —
+the script says so, and `python ci/check-ha-version.py` gives the detail. mypy and
+the Home Assistant unit lane then test an older API than CI does.
 
 ## Typing & quality scale
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -536,10 +536,17 @@ can stop the script: it always prints a summary of what it installed, skipped an
 failed.
 
 ```bash
-bash ci/setup-ci-deps.sh                  # install what is missing
-FORCE=1 bash ci/setup-ci-deps.sh          # install everything again
-SKIP_BROWSER=1 bash ci/setup-ci-deps.sh   # leave Docker and Playwright alone
+bash ci/setup-ci-deps.sh          # install what is missing
+FORCE=1 bash ci/setup-ci-deps.sh  # install everything again
+# Leave one part alone:
+SKIP_PYTHON=1  SKIP_NPM=1  SKIP_VALE=1  SKIP_FFMPEG=1  SKIP_BROWSER=1
 ```
+
+Only one run can hold the lock directory, because the hook starts the script in
+the background and two sessions can open together. A second run says so and stops.
+The exit status is 1 when a step failed, so a caller does not have to read the log.
+The `mutmut` pin comes from `mutation.yml` and the Python floor from `pyproject.toml`,
+so the script cannot go stale on its own when CI moves a pin.
 
 **The Python packages go in `.venv`** (git ignores it), not in the system Python.
 Activate it before you run a Python lane: `source .venv/bin/activate`. The script
--- a/ci/setup-ci-deps.sh
+++ b/ci/setup-ci-deps.sh
@@ -1,0 +1,244 @@
+#!/usr/bin/env bash
+# Install every dependency the CI workflows need, so a session can run the same
+# commands locally (lint, unit tests, frontend tests, mutation, e2e, docs).
+#
+# Idempotent: each step checks first and skips what is already there. Safe to run
+# repeatedly (it is wired to a Claude Code SessionStart hook). No step is fatal —
+# the script always finishes and prints a summary, so one blocked download does
+# not stop a session.
+#
+# Usage:
+#   bash ci/setup-ci-deps.sh                  # install what is missing
+#   FORCE=1 bash ci/setup-ci-deps.sh          # install everything again
+#   SKIP_BROWSER=1 bash ci/setup-ci-deps.sh   # do not touch Docker or Playwright
+#
+# The Python packages go in .venv (git ignores it). Activate it before you run the
+# Python lanes: source .venv/bin/activate
+set -uo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$ROOT" || exit 1
+
+FORCE="${FORCE:-0}"
+SKIP_BROWSER="${SKIP_BROWSER:-0}"
+VALE_VERSION="${VALE_VERSION:-3.9.1}"
+MUTMUT_VERSION="${MUTMUT_VERSION:-3.7.0}"
+BIN_DIR="${BIN_DIR:-$HOME/.local/bin}"
+VENV="${VENV:-$ROOT/.venv}"
+PY_FLOOR="${PY_FLOOR:-3.14}"
+# Newest first. CI runs the floor; this machine often can only run an older one.
+PY_CANDIDATES="${PY_CANDIDATES:-$PY_FLOOR 3.13 3.12}"
+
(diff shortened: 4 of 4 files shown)


prestomation:
/q review Please be critical and skeptical of this new `ci/setup-ci-deps.sh` and the SessionStart hook change. Surface the most serious issues first, and do not hold back the minor ones.

Focus on:

- **Correctness of the shell**: `set -uo pipefail` without `-e` combined with `&&`/`||` chains — where can a failure be recorded as OK, or an OK recorded as failed? Unquoted `$PY_CANDIDATES` word splitting, the `rm -rf "$VENV"` paths (can `VENV` ever be empty or point outside the repo?), `mktemp -d` cleanup on the vale download path, the `install -m 0755` target, and `${VENV#$ROOT/}` in the final log line.
- **Idempotency claims**: each skip check is a proxy for "installed". Which proxies are wrong? A `node_modules` directory that exists but is stale or partial, `styles/ai-tells` present but empty, a `vale` on PATH at the wrong version, a `.venv` that passes the smoke test but has a partial package set.
- **The interpreter-by-test choice**: trying 3.14, then 3.13, then 3.12 and keeping the first that runs one test file. Is one test file a sound proxy for the whole suite? Should a failed candidate leave `.venv` deleted rather than half-built? What happens when every candidate fails?
- **Drift**: the script duplicates knowledge from `ci/install-deps.sh`, `requirements-test.txt`, `mutation.yml` (`mutmut==3.7.0`), and `.vale.ini` (the pinned ai-tells release). Where should that live in one place instead, and what breaks quietly when CI changes and this script does not?
- **Hook safety**: it now runs on every session start, in the background, and can rebuild `.venv`. Can it race a second session, or fight a virtual environment the maintainer made themselves? Is replacing the browser-only hook the right call?
- **Security**: it downloads a vale tarball over HTTPS with no checksum and runs `apt-get install` under `sudo -n`.
- **Home Assistant / maintainability**: the pinned `PY_FLOOR` of 3.14 and the fallback list will go stale. Is the warning enough, or should a below-floor environment be a harder signal?

---

prestomation:
Thanks — triaged. Pushed `28b253b` with the fixes. Where I disagreed I say why.

## Fixed

| # | Finding | Fix |
| --- | --- | --- |
| 1 | Unquoted `$PY_CANDIDATES` | Now an array, read from the env with `read -a` when overridden. |
| 2a | Stale `node_modules` | A `package-lock.json` newer than `node_modules` triggers a reinstall. |
| 2b | Empty `styles/ai-tells` passes | Now checks the directory is not empty. An interrupted sync leaves one. |
| 2c | `vale` presence, not version | `vale --version` is compared to the pin. |
| 3 | No locking | Takes a lock with `mkdir` (atomic), released on a trap. A second run says so and stops. |
| 5 | `rm -rf "$VENV"` safety | A `VENV` outside `$ROOT` exits 2 before any step runs. Verified with `VENV=/`. |
| 6 | Exit 0 after a failed `.venv` | Exits 1 when anything failed. No step stops the run; the status now reports. |
| 7 | Hook now installs everything | Added `SKIP_PYTHON`, `SKIP_NPM`, `SKIP_VALE`, `SKIP_FFMPEG` beside `SKIP_BROWSER`. |
| 8 | Duplicated pins | `mutmut` is read from `mutation.yml`, the Python floor from `pyproject.toml`. |
| 10 | `${VENV#$ROOT/}` on a path outside the repo | Cannot happen now — the guard in 5 rejects it. |
| 11 | `mktemp -d` cleanup hides a failed download | The directory stays on failure, and the log names it. |

Verified after the push: clean run 4 installed / 8 skipped, re-run 0 / 12; the lock refuses a second run; the skip flags each work; a touched `package-lock.json` triggers the reinstall.

## Not changed, with reasons

**2d — the `.venv` smoke test as a proxy.** Your example is a run that got `homeassistant` but missed `ruff` or `mutmut`. Those are not covered by the smoke test at all: each has its own check further down (`vhas ruff`, and a `mutmut` check that compares `importlib.metadata.version`), and a missing one is installed on the next run. The smoke test only decides which interpreter can host the environment, which is what it is good at.

**4 — no checksum on the vale tarball.** A checksum from the same GitHub release adds nothing: whoever can serve a bad tarball can serve a matching checksum. Real integrity needs a key that does not travel with the artifact, and vale publishes no signature. A hardcoded SHA256 would also have to be edited by hand on every version bump, which is the drift you flag in 8. The trust model here is the same one `pip`, `npm ci` and `apt-get` already use in this script.

**9 — `set -uo pipefail` without `-e`.** Deliberate, as you note. Every step is a single `cmd && ok … || fail …` chain, so a failure is recorded rather than swallowed.

**12 — `install` overwrites an existing `$BIN_DIR/vale`.** That is now the point: with the version check in 2c, an overwrite only happens when the vale on PATH is not the pinned one. Leaving a wrong version in place is the worse outcome.

**13 — the below-floor warning does not fail.** Nothing the script can install would fix it: the environment has no interpreter at the floor, and failing would leave the session with no Python tools at all instead of ones that mostly work. The warning names the effect (mypy and the Home Assistant unit lane test an older API) and points at `ci/check-ha-version.py`, which is the check that does fail in CI, where the floor is real.

---

prestomation:
/q review Please re-review after `28b253b`, which acts on your first pass. Be critical and skeptical again, most serious first, and do not hold back the minor ones.

Look hardest at what the new code introduced:

- **The lock.** `mkdir "$LOCK_DIR"` with a `trap 'rm -rf "$LOCK_DIR"' EXIT INT TERM`. What happens to a stale lock after a SIGKILL or a container restart — the next session then skips setup silently and forever. Is the "stop and say so" behavior right for a background hook, or should it wait? Is `${TMPDIR:-/tmp}` the right home for it when the repo can be checked out more than once?
- **The `VENV` guard.** `case "$VENV" in "$ROOT"/?*)` — does it hold for a relative path, a trailing slash, a symlinked `$ROOT`, or `$ROOT/../elsewhere`, which is textually inside `$ROOT` but resolves outside it?
- **The pins read from CI.** `read_pin` uses `grep -oP`, which is a GNU extension and absent on BSD or busybox. The regexes are `mutmut==\K[0-9.]+` against `mutation.yml` and `python_version\s*=\s*"\K[0-9.]+` against `pyproject.toml`. What makes them silently match the wrong line, and is falling back to a hardcoded value the right failure mode, or does it just hide the drift the change was meant to remove?
- **The npm staleness check.** `[ "$dir/package-lock.json" -nt "$dir/node_modules" ]`. Does a fresh `npm ci` reliably move the `node_modules` mtime, or can this reinstall on every run — or never fire after a checkout that rewrites both mtimes?
- **The vale version check.** `vale --version | grep -qF "$VALE_VERSION"` — substring, so `3.9.1` matches `3.9.10`. Worth anchoring?
- **The exit status change.** The script now exits 1 when any step failed. It runs under `nohup … &` in the hook, so nothing reads it there. Does the new status break any other caller, and is a partial failure really an error for a best-effort installer?
- **The new skip flags.** `SKIP_PYTHON` wraps its whole section in an `if … else … fi`. Check that the helper functions and variables that later sections use are still defined when that branch is taken.

Also say whether anything in my previous reply that I declined to change (the vale checksum, the smoke-test proxy, the below-floor warning that only warns) is wrong on the merits.

---

prestomation:
This pass reviewed `28b253b`, so items 1 and 5 were already fixed in `fa140f8` before it landed. The rest is in `58bd1c8`.

## Already fixed in `fa140f8`

**1 — stale lock.** The lock holds the process id. A holder that is gone loses the lock, a holder that is live keeps it. Verified both ways. `58bd1c8` also moves the lock to `$ROOT/.setup-ci-deps.lock`, so two checkouts no longer share one — good catch.

**5 — `grep -qF` substring.** Anchored on a word boundary, so `3.9.1` no longer accepts `3.9.10`.

## Fixed in `58bd1c8`

**2 — the `VENV` guard.** Now resolves with `realpath -m` (falling back to `os.path.realpath`) before the containment test. Verified: `/` refused, `$ROOT/../elsewhere` refused, `$ROOT` itself refused, `.venv` and `$ROOT/.venv/` both accepted.

**3 — `grep -oP` and the loose regexes.** Replaced with `sed` and `awk`. `python_version` is read from the `[tool.mypy]` table only, so another table carrying the same key cannot win. A pin that cannot be read now says so on stderr instead of falling back silently — which was the drift you point out.

**4 — the npm staleness check.** Now compares against `node_modules/.package-lock.json`, the marker npm writes on every install, with the directory as a fallback for a pre-existing tree. Verified: touching the lock file triggers exactly one reinstall, and the run after it skips.

**SMOKE_TEST is a hardcoded path.** You are right, and this was the best finding of the pass: a rename would fail every candidate and leave no `.venv` at all. It now falls back to the first `tests/unit/test_*.py` and says which one it used.

**The below-floor warning is easy to miss.** It now repeats in the end-of-run summary. I did not add `$ROOT/.ci-setup-warnings`: an untracked file that no tooling reads is another thing to go stale, and the summary is what a reader of the log sees last.

## Not changed

**6 — the exit status.** Keeping `exit 1` on a failed step. The design is "no step stops the run", which is about the *run*, not about staying quiet afterwards — a caller that installs 12 things and gets 11 should be able to tell without parsing a log. Nothing calls this script synchronously today; `AGENTS.md` documents the status.

**7 — the helpers defined inside the `SKIP_PYTHON` branch.** As you say, currently safe. They are Python-section internals (`vpip`, `vhas`, `venv_smoke`, `make_venv`) and moving them out would suggest they are usable elsewhere, which they are not.

**Vale checksum.** Still declining, same reason: a checksum published in the same GitHub release as the tarball is signed by nobody, so whoever can serve a bad binary can serve a matching hash. It is not integrity, it is a second copy of the thing being attacked. vale publishes no signature to check against, and a hardcoded hash reintroduces exactly the hand-edited pin that finding 8 asked me to remove. `pip`, `npm ci` and `apt-get` in this same script rest on the same TLS-plus-pinned-version trust, so singling out this one download would not change the script's real exposure.

---

PR #316, opened 2026-09-10
test: fail when a test dependency is missing, never skip quietly

prestomation:
## Why

On #309 the local run looked green while CI was red: this sandbox had no `hypothesis`, so both property files skipped at collection and nobody noticed. A skip reads as "this lane does not cover that", which makes a broken environment look the same as a deliberate exclusion.

The session hook **does** install `hypothesis` — it is in `requirements-test.txt`, which `ci/setup-ci-deps.sh` installs into `.venv`. That run predated the hook. `jsonschema` and `voluptuous-openapi` are held back on purpose (`voluptuous-openapi` pulls `voluptuous` in, and `voluptuous` decides what other modules skip), so `ci/install-schema-deps.sh` installs those in the one job that needs them.

## What changed

A plain import is the whole fix: an `ImportError` at collection is a collection error, and a collection error fails the run.

- `test_recurrence_properties.py`, `test_transfer_properties.py`, `test_api_surface.py`, `test_generate_api_docs.py`, `test_transfer_coverage.py` — `pytest.importorskip` (and one `try: import yaml`) becomes a plain import of a package `requirements-test.txt` names.
- `test_generate_schema.py` imports `jsonschema` and `voluptuous_openapi` plainly, after its `HK_SCHEMA_GATE` skip. The gate is what says the lane opted in, so a missing package there is a broken job. The gate skip itself is unchanged.
- `importorskip` stays for `homeassistant` and `voluptuous` — a bare install really may not have those, and `voluptuous` is held back on purpose.
- `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md` record the rule, including what not to do: a session-start check that reads the requirements file fails the mypy job, which installs its own three packages and nothing else.

An earlier revision of this PR added a `required_deps` helper and that session-start check. CI failed the mypy job on it, which is the argument against the helper, so both are gone.

## Gates

`pytest tests/unit` 2258 passed, 1 skipped (the schema gate, no `HK_SCHEMA_GATE`), 1 xfailed. `ruff check` and `ruff format --check` clean. Behaviour checked by hand: no `HK_SCHEMA_GATE` still skips the schema gate; `HK_SCHEMA_GATE=1` without `jsonschema` errors at collection; a hidden `hypothesis` fails the property files instead of skipping them.

## One-way doors

None. No user-facing surface, no service, event, storage or attribute changes.

Developer-only, so no CHANGELOG entry, no version bump, and no screenshots.

https://claude.ai/code/session_01PQ7kAi2YbNYRjc35KRiCX6


prestomation:
/q review Be critical and skeptical. This turns test-dependency skips into failures. Please scrutinize, most serious first, and do not withhold minor findings:

- **Correctness of the rule**: `tests/unit/required_deps.py` treats `requirements-test.txt` as the source of truth. Is the parser right for every line shape (extras, markers, comments, `-r` includes)? Is the dist-name to import-name mapping (`_IMPORT_NAME`) a drift risk when a new requirement lands?
- **Does it fail where it should?** `pytest.fail(allow_module_level=...)` at import time, and `pytest.UsageError` from `pytest_sessionstart` — do both really abort under `-p no:cacheprovider`, `-x`, `-n auto`, and inside mutmut's sandbox, or can either be swallowed?
- **Does it fail where it should not?** Lanes that legitimately have fewer packages: mutation (mutmut/Stryker), integration, the docs build, `ci/test-python-unit.sh`. `HK_ALLOW_MISSING_DEPS` is the escape hatch — is it reachable from every one of those?
- **The E402 noqa** added to the property files: is that hiding a real import-order problem now that `pytest.importorskip` no longer opens the file?
- **Maintainability**: is the helper's split between `require()` and `optional()` clear enough that the next contributor picks the right one, or would a single function with the list doing the deciding be better?

---

prestomation:
/q review This head replaces the `required_deps` helper with plain imports, after the helper's session-start check failed `lint.yml`'s mypy job (that job installs only homeassistant, voluptuous-openapi and jsonschema). Be critical and skeptical, most serious first, and do not withhold minor findings:

- **Does a plain import really fail every lane?** Collection errors under `-n auto`, mutmut's sandbox, Stryker, and the coverage workflow — can any of them turn a collection error into a silent zero-test pass?
- **Did removing the guards weaken any lane?** `homeassistant` and `voluptuous` keep `importorskip`. Is that line drawn in the right place, or does a lane exist where those skips now hide the same class of defect #309 shipped?
- **`test_generate_schema.py`**: plain imports after a module-level `pytest.skip`. Is the order right, and does the mypy job's grep for "the tests really ran" still hold?
- **Import ordering**: the property files now import `hypothesis` at the top. Any circular-import or `conftest` ordering risk with `property_strategies` and the Hypothesis profile registration?
- **Docs**: `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md` now say what not to build. Is that accurate about which packages each CI job installs?

---

PR #317, opened 2026-09-10
Feature/pull forward task

joss94:
Fixes #

## Checklist

- [ ] Tests run locally (`pytest tests/unit -v`, `bash ci/test-frontend.sh`)
- [ ] `CHANGELOG.md` updated — user-facing changes only, with `(Fixes #N)` for each
      issue this fixes (developer-only changes don't need an entry)
- [ ] Panel UI changed → current screenshots committed under `docs/images/` and
      embedded above with an HTML `` tag
- [ ] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended to
      step through it
- [ ] New user-facing feature → version bumped to the next beta (`manifest.json` +
      `const.py`), `preview-release` label applied

PR #318, opened 2026-09-10
feat(panel): Due today, a task's due date moved to now whatever its schedule

### LINKED ISSUE #312: Force the next due date
Hi !
I just installed your add-on after spending quite some time trying to configure a custom dashboard working with Home Maintenance. You have done a fantastic job and I found with Home Keeper almost all the features I was missing in Home Maintenance. Thank you for that !

One feature I woul enjoy having:

Being able to force the next due date of a "backlog" task to today, independently of a schedule. 

Use case example:

I scheduled a task to do a monthly check-up on my bike. This task is not overdue according to this schedule. But I have noticed that I have a flat tire, and I need to take care of it now -> I'd like to click a "schedule for today" button to switch the due date to today, so that I remember to do it.

Many other examples come to mind: 

- Clean bathroom asap because guests are coming soon
- Wash clothes asap before coming trip
- ...

Thanks in advance for considering this :)

prestomation:
Fixes #312

Supersedes #317 by @joss94, whose implementation this carries. He reported #312 and wrote the feature; this branch finishes the naming, the gates and three defects found in review. Credit is in the CHANGELOG bullet.

## Rebased onto main (2026-09-12)

The branch is now a single commit on top of `main`, and 3 things changed in the rebase:

- **The version is `0.24.0b2`, not `0.23.0b2`.** `0.23.0` shipped as a stable while this
  PR sat open, and `main` moved to `0.24.0b1`. PEP 440 sorts `0.23.0bN` below the shipped
  stable, so a beta in that series reads as a downgrade in HACS. The CHANGELOG bullet
  moved to a new `## [0.24.0b2]` section for the same reason.
- **The prose moved out of `README.md`.** #333 split the User Guide into one authored
  file per page, so the Due today text now lands in
  `docs/guide/tasks/snooze-and-skip.md` (retitled *Snooze, skip and due today*),
  `docs/guide/automation/services.md` and `docs/guide/automation/events.md`, with the
  matching `USER_SECTIONS` title in `website/scripts/doc-map.mjs`. `README.md` is
  untouched.
- **The CHANGELOG doc-link rule is no longer duplicated.** #329 landed that rule on
  `main` while this PR was open, so this branch keeps only the part the landed rule was
  missing: the bold sits **outside** the link, because `_BOLD_LEAD` anchors on `^\*\*`.
  The 2 tests pinning both forms are unchanged.

### The name

The feature shipped in #317 as **Pull forward**, but the button already said **Due today** and everything else — README, settings switch, service, event, option key — said Pull forward. Picking one mattered beyond tidiness: for a date, moving it *forward* reads as **later**, which is the opposite of what this does, so the docs had to gloss the term at every mention.

It is now **Due today** everywhere, in the UI and in the contracts alike. None of these had shipped (#317 is unmerged and no release predating this PR carries them), so this was the last moment the rename was free rather than a one-way door.

### Three defects

**The action was offered on a task that is already due or overdue.** One tap there pushed the due date *later* and silently dropped the overdue state, with no dialog, no toast and no undo, immediately beside Done. #312 asked for this on a task explicitly *not* overdue ("This task is not overdue according to this schedule. But I have noticed that I have a flat tire"), and `notifications.py` already refused the verb on notifications for the same reason — a rule the panel and card contradicted. `deferVerbs` now withholds it, and takes an injectable `now` so the rule is testable rather than racing the wall clock.

**The card row was over-stuffed.** Four buttons — three 36px actions plus Done at HA's default 48px — take 156px of a ~330px dashboard column, leaving 134px for the name. Task titles wrapped to four lines and appliance chips clipped under the Overdue badge. Below 400px the actions now take a line of their own, via a **container query rather than a media query**, because a card's width comes from its dashboard column and not the viewport: a 1280px screen holds a 330px card.

**The Settings card was titled for three verbs and summarised two.** It read "Snooze and skip are available" under a heading naming three, and the rail dot stayed green with the third turned off. The summary is now composed from the verbs on offer rather than enumerated, so a fourth verb costs one row instead of doubling the strings.

### Also

- The long-press on a card row now says **"Make this task due today"** instead of joining the label and hint into "Due today — Move the due date to today", which just echoed itself.
- The capture's defer locators are scoped to `.hk-detail-actions`, as `walkthrough.capture.ts` already does — an unscoped `.hk-defer-skip` picks whichever closed menu comes first in the DOM.
- `store.set_due_today`'s docstring claimed the schedule "resumes exactly where it would have without this call". That is false for a floating task, which measures from its last completion: completing on the earlier date shifts the whole future schedule earlier. `snooze_task` never made that claim.
- A glossary row for the new name in `.amazonq/rules/writing-style.md`, per "Add a row when a new name appears".
- `tsconfig` `lib` moves to ES2021 for the type of `Intl.ListFormat`. `target` stays ES2020.
- Two water-heater PDFs that local e2e runs left behind, and a previous commit swept in, are dropped. `main` tracks four seeded documents and so does this branch again.

### Known issue this PR does not fix

**The seeded fixture uses absolute dates, and the captures that assert on this button depend on one.** "Replace fridge filter" is seeded due 2026-09-28. Once that passes, the button correctly disappears and the panel and walkthrough captures fail for a reason unrelated to the code.

I tried moving the two future-dated tasks a year out and **reverted it**: a floating task recomputes `next_due` from `last_completed`, so the completions had to move too, which put six completions in the future — not a state Home Keeper can be in. It broke eight e2e assertions across five specs (the shopping scope stopped being empty; documents, measured-stock, problem-sensor and settings went red). The real fix is seeding *relative* dates, which changes the fixture format and belongs in its own change.

## One-way doors

Renamed before first release, so none of these is yet a committed surface — that is the point of doing it now:

| Surface | Was | Is |
| --- | --- | --- |
| Service | `home_keeper.pull_forward_task` | `home_keeper.set_due_today` |
| Event | `home_keeper_task_pulled_forward` | `home_keeper_task_due_today_set` |
| Option key (stored) | `allow_pull_forward` | `allow_due_today` |
| Websocket command | `home_keeper/pull_forward_task` | `home_keeper/set_due_today` |

The event carries `origin` for loop prevention, as snooze and skip do. The option defaults **on**, so an options document written before it existed reads back with the verb available.

The panel and card withhold the verb on an already-due task; the store does not. That asymmetry is deliberate and documented on `set_due_today` — a service caller may still move such a task, which only pushes its due date later.

## Screenshots

Desktop and phone, SHA-pinned to 5d36b0d93c6ef65f166ff86b96b2c85d75adf197, the rebased head. The images are the same ones captured on the pre-rebase head; the rebase changed no panel or card markup.

The Done caret, with the verb named:

A task due in 19 days with its Done caret open, listing Snooze, Skip and Due today with a line each

Settings, desktop then phone. The summary now names all three verbs, and the help text states the already-due rule:

The Snooze, skip and due today settings card with three switches and a summary naming all three

The same settings section on a phone, one screen at a time

The card. Rows carry Due today only where the task is due in the future, and the wrapped layout puts task names back on one line with the appliance chips unclipped:

A dashboard card whose overdue rows show snooze, skip and done, while the two tasks due in the future also show due today

## Verification

Run locally against the rebased head:

| Gate | Result |
| --- | --- |
| `pytest tests/unit` | 2279 passed, 1 skipped, 1 xfailed |
| `bash ci/test-frontend.sh` | 1362 passed, 44 files |
| `ruff` lint + format | clean over `custom_components tests ci` |
| `vale` | 0 hits added vs `main`, compared as a set per file |
| `check-changelog-release-gap.py` | exit 0 |
| `npm run sync` (site) | 24 User Guide pages, 5 Developer Guide pages, API reference |
| `mypy` | no new errors (see note) |

Playwright e2e, the walkthrough tour and both mutation lanes were green on the
pre-rebase head (310 e2e passed; Python and TypeScript mutation at 100% on the changed
surface). CI re-runs all three here.

**mypy note:** this container's Python is 3.12, below Home Assistant's 3.14 floor, so pip backtracks to an old HA release — the #199 trap. Two errors appear (`card.py` `LOVELACE_DATA`, `transfer_runner.py`), both in files this branch does not touch and both present without it. CI checks it on a Python at the floor.

## Checklist

- [x] `CHANGELOG.md` updated — with `(Fixes #312)` and the contributor credit
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML `` tag
- [ ] New user-facing UI surface → `tests/e2e/walkthrough.capture.ts` extended to step through it — the tour already steps through this menu and the card row; the verb is renamed in place rather than adding a surface

https://claude.ai/code/session_01XokgrL1RxPq3WFqYDZGi2w


prestomation:
/q review Please be critical and skeptical, and surface the most serious problems first without withholding the minor ones. This takes over #317 (feature by @joss94), renames the verb to "Due today" across the UI and the service/event/option contracts, and fixes three defects. Areas I want scrutinised:

**Correctness.** `deferVerbs` in `frontend/src/defer.ts` now withholds the verb when `isOverdue(task, now)` — check the boundary (`isOverdue` is `<=`), timezone/DST behaviour, and whether a task due later *today* is handled the way a user would expect. `store.set_due_today` writes `dt_util.now().isoformat()`; consider a seasonal task, a fixed-recurrence task, and a task whose season is inactive. Is there a case where moving `next_due` to now produces a state the coordinator's edge-triggered transitions handle badly, or double-fires an overdue event?

**The UI/store asymmetry.** The panel and card hide the action on an already-due task, but the store still permits it for service callers. Is that split defensible, or should the store reject it too? Argue the other side.

**The rename.** Verify nothing still references `pull_forward`, and that `api_surface.py`, `services.yaml`, `strings.json` and the 15 translations agree. The stored option key changed from `allow_pull_forward` to `allow_due_today` — confirm that is genuinely safe given the feature has never been released, and say so if you think it is not.

**CSS.** `card.ts` uses a container query (`container-type: inline-size` on `.hk-body`, wrapping below 400px). Check the breakpoint, whether `.hk-acts:empty` covers every case, and whether the wrapped layout breaks any row variant — a dormant monitored task, a completed one-off, a blocked Done.

**i18n.** `tlist()` wraps `Intl.ListFormat` and `settings.skipsnooze_available` is a frame with a `{verbs}` placeholder. Check the composed sentence reads correctly in German, Russian, Polish and Chinese, and that reusing the `btn.*` labels mid-sentence is right.

**HA best practices** for the new service and event, and anything about the websocket command's error handling.

---

prestomation:
/q review Follow-up after a force of changes since the last request — please re-review at head `10973a2` and be critical. Two things changed that are worth your attention specifically:

**1. A fixture change I made and then reverted.** I moved the two future-dated seed tasks a year out so the captures asserting on the Due today button would not expire on 2026-09-28. A floating task recomputes `next_due` from `last_completed`, so the completions had to move too — which put six completions in the *future* and broke eight e2e assertions across five specs. I reverted it and documented the horizon as a known issue instead. Do you agree that future-dated completions are the actual cause, and is there a way to push the seeded due date out that I missed which keeps history in the past? Also: should `models`/`store` reject or normalise a completion timestamp in the future at all, given nothing currently does?

**2. Why does the same seeded `next_due` survive on one task and not the other?** "Replace furnace filter" kept a committed `next_due` of 2027-11-28 with `last_completed` in 2026, while "Replace fridge filter" was recomputed back to `last_completed + 1 month`. If reconcile only recomputes under some conditions, I would like to know the rule — it looks like something a user could hit.

Everything from the first request still stands, in particular the `deferVerbs` overdue boundary (`isOverdue` is `<=`), the panel/store asymmetry, the container query in `card.ts`, and the `Intl.ListFormat` sentence in German, Russian, Polish and Chinese. Most serious issues first, and please do not withhold the minor ones.

---

prestomation:
Thanks — taking these in turn. CI is green on `37b3eb6`.

## The `isOverdue` boundary: not changing it

You quote it correctly (`<=`, "at or before now"), but I don't think `<` is right, for two reasons.

**A task due exactly at `now` has nothing to move.** `set_due_today` writes `next_due = dt_util.now()`. On a task whose `next_due` already equals now, that is a no-op. Under `<` the button would appear and do nothing. Withholding it is the more honest answer, and a boundary that resolves to "no visible effect" is the one case where offering the control is strictly worse.

**`isOverdue` is not local to this feature.** It has five call sites — the card row's overdue rail (`card.ts:893`), the panel's overdue filter scope (`panel-controls.ts:289`), the task list (`panel-lists.ts:238`), the status chip (`utils.ts:736`), and `deferVerbs`. Moving the boundary to `<` would relabel a task due exactly now across all of them, changing what "Overdue" means app-wide to buy one no-op button. If the boundary were ever revisited it should be for the chip's sake, not this verb's.

The comment on `deferVerbs` already says "already due or overdue", which matches `<=`. So the code and the comment agree; it is the PR description's "not yet due" that was loose, and I'll tighten that wording rather than the code.

## Two things you could not see

`card.ts` and the `Intl.ListFormat` helper **are** in the diff — you reported not finding either, so the review may have run against a partial view. For a re-review:

- Container query: `custom_components/home_keeper/frontend/src/card.ts`, in `STYLES` — `.hk-body { container-type: inline-size; container-name: hk-card }`, an `@container hk-card (max-width: 400px)` block that wraps `.hk-row`, and `.hk-acts:empty { display: none }`.
- List formatting: `custom_components/home_keeper/frontend/src/i18n.ts` — `tlist()` wrapping `Intl.ListFormat`, consumed by `deferVerbSummary` in `panel-settings.ts` against `settings.skipsnooze_available` (`"A task offers {verbs}"`).

## Future-dated completions

Agreed on the diagnosis, and agreed that store-level rejection is the wrong place — a user back-dating a completion and mis-picking tomorrow deserves a gentler failure than an exception. I have not added validation here; it is out of this PR's scope, and worth its own change if the maintainer wants it.

On the seed asymmetry: your reading describes the reverted seed correctly. What I saw was against my own edited seed, which I have since reverted, so there is nothing left to explain and I am not pursuing it.

---

prestomation:
/q review This branch was rebased onto `main` as a single commit after `0.23.0` shipped. Please be critical and skeptical, and put the most serious findings first without withholding the minor ones.

Focus on the rebase itself, since that is what is new since your last look:

- **Correctness of the rebase.** `main` gained the fixed-schedule change (#332), the User Guide split (#333) and the CHANGELOG doc-link rule (#329) while this PR was open. Did any of the Due today code lose an edit, or contradict the new fixed-schedule advance in `recurrence.py`? `store.set_due_today` and the fixed-schedule path both move `next_due`.
- **Version and release gates.** The bump is `0.24.0b2` with a new CHANGELOG section. Check that `manifest.json`, `const.py` `PANEL_VERSION` and the section agree, and that `(Fixes #312)` sits in the bullet's first paragraph.
- **Docs completeness.** The prose moved from `README.md` into `docs/guide/tasks/snooze-and-skip.md`, `docs/guide/automation/services.md` and `docs/guide/automation/events.md`. Is anything the old README section said now missing, or is any surface documented that the code does not have?
- **Edge cases in `deferVerbs`.** It withholds Due today on an already-due task and takes an injectable `now`. Timezone and DST, the boundary at exactly now, and a task with no `next_due` at all.
- **Error paths and the privilege boundary.** `set_due_today` ships as a service, a websocket command and a notification path. Is the admin gate on both the websocket command and its service twin, per `docs/SECURITY.md`?
- **HA best practices** and **performance**: the event firing at the `store.py` chokepoint, and whether the card's container query costs a reflow per row.

---

prestomation:
The `walkthrough` check was red on `5d36b0d`, and it was this PR's own defect rather than a flake. Fixed in `4344cd6`.

**What failed.** The tour asserted `.hk-defer-due-today` in the water filter's menu. That task is seeded due 2026-05-20, so it is overdue, and the rule this PR adds is exactly what withholds the verb there. The tour was asserting the opposite of the feature.

**The fix.** The overdue beat now asserts the entry is *absent*, which pins the rule instead of tripping over it, and a new beat opens the furnace filter — due in 78 days — to show all 3 entries. Both tours carry it, since below 700px the menu is a different layout. I also added the matching assertion to `tests/e2e/tests/skip-snooze.spec.ts`: no spec named `.hk-defer-due-today` before this, so a capture was documenting a surface nothing tested.

Verified on a clean container: both tours pass, and the gif shows Snooze, Skip and Due today on the 78-day task. 48 defer specs pass under the CI browser with `CHROMIUM_EXEC` unset.

---

On the Cue review above:

1. **Version bump consistency** — the CHANGELOG does carry a `## [0.24.0b2]` section, at the top of the file, with the Due today bullet. It was added in the rebase.
2. **`snooze_task` `hours` vs `until`** — a real suggestion, but it changes a shipped service's schema, so it is not this PR's to make. `snooze_task` predates this change and nothing here touches it.
3. **Websocket error handling** — the snooze and skip handlers already resolved the coordinator by hand before this PR; `set_due_today` follows its 2 siblings rather than splitting the trio across 2 patterns. Moving all 3 onto `@_with_coordinator` is worth doing on its own.
4. **`origin` in the docs** — it is documented, in the 2 places an integration author reads: the field description in `services.yaml` (which the Home Assistant UI and the generated API reference both render) and `docs/INTEGRATING.md`, which gives the loop-prevention pattern with an example. `docs/guide/tasks/snooze-and-skip.md` is the household-facing page, and `origin` is not a household concept.

---

prestomation:
/q review Since your last look the branch gained `4344cd6`, which fixes the red `walkthrough` check. Please be critical and skeptical, most serious findings first, and do not withhold the minor ones.

Review that commit closely:

- **`tests/e2e/walkthrough.capture.ts`** — the overdue beat now asserts `toHaveCount(0)` for `.hk-defer-due-today`. Is that assertion able to pass vacuously (a menu that never opened also has 0)? The `.hk-defer-skip` `toBeVisible()` above it is meant to be the guard; say so if it is not enough.
- **The new future-due beats** open the furnace filter in both tours. Does step 3 of the desktop tour leave `require_tag_scan` or `tag_id` set on that task in any path, which would change what its menu offers?
- **Date fragility.** The beats depend on the furnace filter being seeded due 2026-11-28. Is there a way to express "a task due in the future" that does not hard-code a task, without changing the fixture format?
- **`tests/e2e/tests/skip-snooze.spec.ts`** — the new spec navigates twice with `gotoPanel`. Is reusing `let panel` / `let caret` across 2 navigations sound in Playwright, or should the second half be its own test?
- Anything in the rest of the diff you would still push back on.

---

prestomation:
The `walkthrough` check went red again on `4344cd6`, for a different reason than the first time. Addressed in `f01d600`.

**What the run says.** Three attempts, three different failure points, none of them near the beats this branch adds:

| Attempt | Died at | Elapsed |
| --- | --- | --- |
| 1 | `seedBuyReminder`, `page.evaluate` lost its execution context to a navigation | 7.3s |
| 2 | the card's `.hk-defer-snooze`, after the test hit its 240s cap | 4.0m |
| 3 | the sensor form's summary read `Every 300 of use, or every 6 months` where the step expects `When it changes to on` | 1.7m |

Attempts 2 and 3 both ran *past* the new beat at step 2a3b, so the beat itself works. Every other check on that head is green, including `e2e`, which covers the same card buttons attempt 2 could not find. The same job passed on #321 five minutes after the first of these runs, so the runner was not simply down.

That is the desktop tour running out of budget, and `walkthrough.config.ts` already names the trap:

> Read a future timeout that way round: suspect the margin first, and only go looking for CI-specific slowness if a green run ever reports a length the container cannot reproduce.

The tour measured 168s on CI against a 240s cap. The desktop beat spent 2 navigations and 5 `BEAT`s of that margin to show a menu the tour already photographs elsewhere.

**The fix.** The desktop walk gives the budget back. What stays there is the overdue assertion, which swapped `toBeVisible` for `toHaveCount(0)` and costs nothing, and step 8a, which already frames Due today on the card rows of the 2 tasks due in the future — so the desktop gif still carries the verb. The phone tour keeps the detail-menu beat: it passed on all 4 attempts, it runs 37s against the same cap, and below 700px that menu is a different layout rather than a narrower one.

I did not raise the timeout. The config says to move that number against a fresh measurement rather than a hunch, and one overloaded run is not a measurement.

Locally on a clean container: desktop back to 3.3m, the length it ran before the beat, phone 37s, both green. The coverage that matters is unchanged — `tests/e2e/tests/skip-snooze.spec.ts` asserts the rule in both directions, at every viewport the `@responsive` tag runs.

---

PR #319, opened 2026-09-10
Withhold Edit when a companion locks every field

prestomation:
Fixes #

## What changed

`managed_by.locked_fields` is a partial contract: `taskSchemaSections` drops the fields a companion names and keeps the rest, so a glue that owns the name and the device still leaves a form worth opening. Claim every field the kind offers, though, and every section comes back empty — and the task page still drew an **Edit** button that opened a drawer with no rows in it.

Came out of asking why Battery Notes and Pawsistant tasks still have a working Edit. They do, and that is correct — they claim only what they write. Both glues are being tightened in their own repos (prestomation/ha-home-keeper-battery-notes#24, prestomation/Pawsistant#84), and neither goes as far as locking everything. This is the guard for when a companion does.

**New `taskFormIsEmpty` in `forms.ts`**, built on `taskSchemaSections` so the guard and the form it guards cannot disagree. `panel-detail.ts` asks it inside the existing `if (!sourceOwned)` branch; when true, Edit gives way to a caption naming the owner — the courtesy a source-owned task already gets, per the rule already written into the comment above it ("say why Edit and Delete are missing rather than just omitting them").

**Delete and the "Edit in X" deep link are untouched.** Deletion protection is a separate declaration a companion makes, and the integration's own UI is the honest destination once the panel has nothing left to offer — the analogue of the `d-edit-recipe` button on a recipe's task.

Every option list on `taskFormIsEmpty` is **required**, unlike on `taskSchema`. Two of them decide whether a field exists at all (the consumable and card-links pickers appear only when there is something to pick), so a caller that let one default to empty would be asking about a shorter form than the drawer is about to build, and could withhold Edit from a task that still had a picker to offer.

### A note on ordering

Each caption stands where its button would have been — the allLocked caption in the Edit slot, the delete-blocked caption in the Delete slot, greyed Duplicate between them. That is positionally consistent with the buttons it replaces, but it does mean two captions bracket one button. Happy to move them together if you would rather they read as one block.

## Screenshots

The seeded "Replace battery: Thermostat" task is widened to lock all 8 fields the `triggered` kind offers, which is what makes this surface photographable. It is deliberately *not* what the shipped glue sends — Home Keeper has to handle whatever `locked_fields` an integrator writes.

Desktop:

Task page with no Edit button, a caption reading Battery Notes sets every field on this task, a greyed Duplicate and the delete-blocked caption

Phone:

The same task page below 700px, where the two captions take lines of their own around the greyed Duplicate

I reverted the ~80 other PNGs the capture run rewrote: their only change was the version footer, and one of them (`9-panel-managed-detail.png`) had caught the page mid-load. Neither belongs in this diff.

## One-way doors

None. `managed.allLocked` is an internal locale key, and the guard reads a `managed_by` field that is already part of the published contract. The `docs/INTEGRATING.md` table row is documentation of a consequence, not a new surface.

## Testing

- `npx vitest run` — 44 files, **1360 passed**, i18n parity included.
- `npx tsc --noEmit` — clean.
- `bash ci/test-mutation-frontend.sh` — **100%** on the changed surface (4/4). The first run scored 57% on three surviving default-argument mutants; making the lists required removed them and improved the API, rather than papering over them.
- `vale` — compared the hit *set* per file against `origin/main` rather than filtering by line number: 81 on both sides, **zero new**.
- Playwright: `list-honesty`, `task-duplicate`, `battery`, `detail-edit` — **26 passed** across desktop/phone/tablet. `screenshots.capture.ts` passed end to end.

New coverage: `taskFormIsEmpty` unit tests in `forms-schema.test.js` (including a loop dropping each field in turn, so the predicate cannot degrade into "is this task managed at all"), a `managed-fully-locked.test.js` render test, and an e2e case in `list-honesty.spec.ts` beside the existing source-owned one. I verified the render test genuinely bites: 3 of its 7 fail with the guard stashed out.

One assertion started out weak — `.hk-managed-info` also matches the delete-blocked caption, so it passed without the guard. It now matches the caption's own text.

No e2e `.d-open-in` assertion: the seeded glue tasks record Home Keeper's own `home_keeper_test_entry`, so the deep link is correctly suppressed for HK's domain. The unit test covers it with a config entry that really belongs to a glue.

## Checklist

- [x] Tests run locally — frontend suite, typecheck, mutation, vale, Playwright (see above). Python untouched by this change.
- [x] `CHANGELOG.md` updated — new `## [0.23.0b2]` section. No `(Fixes #N)`: no issue filed for this.
- [x] Panel UI changed → desktop and phone screenshots committed under `docs/images/` and embedded above with HTML `` tags
- [ ] New user-facing UI surface → walkthrough. **N/A** — a fix to an existing surface, no new one.
- [x] Version bumped to `0.23.0b2` (`manifest.json` + `const.py`). Forced rather than optional: `v0.23.0b1` is already a published tag, so `changelog-release-gap` fails any edit to that section without a bump. No `preview-release` label — this is a fix, not a feature.

https://claude.ai/code/session_019sb4tequsceQBQT5R4CySf

---
_Generated by [Claude Code](https://claude.ai/code/session_019sb4tequsceQBQT5R4CySf)_


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first. Do not withhold minor ones.

Focus on:

1. **Correctness of the predicate.** `taskFormIsEmpty` asks `taskSchemaSections(...).every(s => !s.fields.length)`. Is there any task kind or state where that returns true but the drawer would still have rendered something useful — a footer Delete, a History button, a section the guard cannot see? `renderTaskForm` adds chrome around the sections; I only counted the sections. If the drawer is never truly empty, the caption is a lie.

2. **The reverse.** Any case where the form really is empty but the guard says otherwise, so Edit still opens an empty drawer? Consider the non-`triggered` kinds especially — I derived the 8-field list from the `triggered` branch and did not enumerate the others.

3. **Orphan behaviour.** I keep Edit withheld when the owning integration is unloaded, on the grounds that `locked_fields` is stored on the task so the form stays empty either way. Delete comes back for cleanup. Is that the right split, or should an orphaned task become fully editable since nobody is left to enforce ownership?

4. **The single entry point claim.** I gate only the `.d-edit` button and deliberately did *not* also guard `_openEdit`, arguing `.d-edit` is the only way into the task drawer (`_pendingEdit` is set by `_openEdit` itself; no URL route reaches it). Please verify that against `panel.ts` routing rather than trusting me — if there is a second path, the guard is cosmetic.

5. **Making the option lists required.** I removed the `= []` defaults so callers cannot ask about a shorter form than the drawer will build. It diverges from `taskSchema`/`taskSchemaSections`, which keep theirs. Is the inconsistency worth the safety, or should all three agree?

6. **Test honesty.** Do the new tests assert behaviour or shape? `test_locks_every_field`-style derivation is easy to make tautological. Also whether widening the seeded thermostat fixture to a `locked_fields` set no real glue sends is a fair fixture or a misleading one.

7. **HA best practices**, the `panel-detail.ts` → `panel-task-form.ts` import (I exported two previously-private helpers), and anything about the caption ordering I raised in the PR body.


---

PR #321, opened 2026-09-10
feat: counted wear items — repeat every so many uses, or so many months

### LINKED ISSUE #306: Feature request: Manual usage counters
Thanks again for the positive response my other feat request. :) Another feature that I think could fit Home Keeper particularly well would be a simple Home Keeper-managed usage counter.

Usage/Meter tasks are already very flexible when a suitable numeric sensor exists. But for many real-world things there simply is no native meter:

- wearing a rain jacket
- hikes with a pair of boots
- charging cycles
- uses of a tool
- washing cycles
- trips with camping gear
- filter uses
- any arbitrary event that Home Assistant can detect

Today I could build this myself with an HA counter or `input_number`, but having it directly associated with an appliance/asset would make this much easier and more discoverable. For example:

```
Rain jacket
Usage counter: Wears
Current: 17
```

and then a task:

```
Renew DWR
Every 25 wears
OR every 12 months
```

The really useful part is that Home Assistant could increment that counter from anything: 
a dashboard button, NFC/tag scan, automation, device event, state change, script/service call

So Home Keeper would not need to know what a `wear`, `trip` or `cycle` actually is. HA would decide when usage happened, and Home Keeper would only track the number and use its existing Usage/Meter logic.

Ideally the counter could be exposed as an HA entity and support something simple like:

```
+1
-1
set value
reset
```

This would make Usage/Meter tasks useful for a lot more objects without needing a dedicated sensor for each use case.

I think it would be a nice example of Home Keeper making deeper use of Home Assistant rather than reimplementing the automation side itself.

prestomation:
Fixes #306

## Rebased onto main (2026-09-12)

The branch is now a single commit on top of `main`, and 5 things changed in the rebase:

- **The version is `0.24.0b3`, not `0.23.0b3`.** `0.23.0` shipped as a stable while this
  PR sat open, and `main` moved to `0.24.0b1`. PEP 440 sorts `0.23.0bN` below the shipped
  stable, so a beta in that series reads as a downgrade in HACS. `b2` is left to #318,
  which merges first.
- **`recurrence.py` takes both sides of 3 conflicts with the fixed-schedule fix (#332).**
  A completion and a skip on a fixed task now call `_advance_fixed_schedule` *and* accept
  `REC_USE`. `move_completion` excludes `REC_FIXED` **and** `REC_USE`: a fixed task keeps
  the date it advanced to, and a use task has no date to compute. Either side alone would
  have been a regression — dropping `REC_FIXED` reintroduces #331, dropping `REC_USE`
  makes the first completion of a use task raise.
- **The prose moved out of `README.md`.** #333 split the User Guide into one authored
  file per page, so the counted-wear-item section now lands in
  `docs/guide/appliances/appliances.md` as *Count uses instead of months*, at the same
  heading depth as its neighbours and with `../../images/` paths. `README.md` is
  untouched.
- **The walkthrough's 2 notify-row locator fixes are already on `main`.** #333 landed
  them in a stricter direct-child form (`> .hk-item-body ha-form`, and the fold
  assertion on the body itself rather than on a form inside it), so `main`'s version
  stands and this branch drops its own.
- **The CHANGELOG bullets now link their doc page**, per the rule #329 landed while this
  PR was open.

## What changed

A wear item can now measure its interval in **uses** instead of in time. Such a part
generates 2 tasks: a **use task** the household completes once per use, and a
**replacement task** that arms once the count reaches the target or an optional time
backstop elapses, whichever comes first.

The count is not a stored number. It is the use task's completion log, filtered to the
entries later than the replacement task's `last_completed` — so completing the
replacement resets the count by construction, there is nothing to keep in sync, and the
household gets a usage history rather than an integer. Anything in Home Assistant can
record a use, because a use is an ordinary completion: the Done button, a to-do item, an
NFC scan, a device-page button, a notification action, or `home_keeper.complete_task`
from an automation.

A wear item is also no longer always *replaced*. **Action** names its task Replace,
Clean, Service, Renew, Sharpen, Rotate or Inspect.

Design and rationale: `docs/COUNTED_WEAR_ITEMS_PLAN.md`.

### The plan needed correcting first

This implements the plan on `704753a`. Reviewed against the code on `main`, that plan
would have shipped broken, so the first commit rewrites it. The substantive corrections:

- **`recurrence.py` was absent from its file list**, and it has 5 branch points a use
  task reaches. Two `raise ValueError` on an unknown type outright; 3 more read
  `not in (REC_TRIGGERED, REC_SENSOR)` and would have called `compute_next_due`, hitting
  the same raise. The *first* completion of a use task would have raised before
  recording anything.
- **`store._stamp_part_replacement` fires on any `source.part`.** A use task carries
  one, so without a gate every wear of the jacket would have stamped `last_replaced`
  and drawn a spare out of inventory.
- **The reconciler's index was keyed `(asset_id, part_id)`.** One part now owns 2 tasks,
  which collide there.
- **`build_task` creates a triggered task armed, not dormant**, so every counted wear
  item would have shipped overdue the moment its part was saved.
- **`replace_unit` validated against `UNITS`** — the same list `models.normalize_fields`
  reads for a *task's* unit, so adding `uses` there would have made an uncomputable
  floating task valid.
- **The `card-filter` insertion point was unreachable**: `if (!task.next_due) return
  'none'` sits above the `shopping` line the plan modelled the bucket on, and a use task
  is dateless for its whole life.
- **The completion cap could eat the count.** `MAX_INTERVAL` is 10,000 but
  `recurrence._record_entry` trims every completion list to 500, blindly. A `uses` target
  is capped at 250, so `2 × target` lands exactly on that cap.

Three questions the plan left open were settled before implementation: the **time
backstop** the reporter asked for in all 3 of their comments is in scope; the status
bucket is **`counted`**, because `SENSOR_MODE_USAGE` already claims "usage"; and the
`AGENTS.md` rule against commenting on issues stands, so the plan's §10 is gone.

### Two defects found by building it

Both were invisible to the unit tier, and both are now covered.

**`_PART_SCHEMA` is a strict voluptuous schema.** Every new part field had to be
declared there or `add_asset` and `update_asset` answer `400`. Every unit test passed
throughout — they call `assets._normalize_part` directly and never cross the service
boundary — while the whole feature was unreachable from an automation. The integration
tier found it, which is the #183 lesson exactly.

**`mergePartForm` cleared `replace_unit` whenever the interval was empty.** Harmless
while every unit measured time (the field re-seeds to `months`), and a dead control the
moment `uses` existed: picking it before typing a target discarded the choice, so the
counting fields it reveals never appeared and the form looked broken.

### Also

- `ci/test-python-integration.sh` could not run locally at all. It passes
  `-p no:pytest_socket`, but the plugin that blocks the socket is registered as
  `homeassistant`. CI never noticed because `requirements-test.txt` deliberately keeps
  that package out of the integration lane, so the flag is inert there — but in the
  `.venv` `ci/setup-ci-deps.sh` builds, all 210 integration tests errored with
  `SocketBlockedError`. Running the tier the way `AGENTS.md` prescribes was impossible.
- The seeded `.storage/` files were restored from `main` first. A local integration run
  writes back to those tracked files and an earlier `git add -A` swept the residue in —
  57 leftover appliances and 2 duplicate onboarding entries. Same trap #318 hit with 2
  stray PDFs. The seed now carries **one deliberate** counted wear item: a rain jacket
  17 wears into a 25-wear cycle.
- Glossary rows for *wear item*, *counted wear item* and *use task* in
  `.amazonq/rules/writing-style.md`.
- Three defects in `walkthrough.capture.ts`, found by running the tour rather than
  reading it. The count read used the row's `innerText`, which never carries the number
  — `ha-assist-chip` paints its label in its own shadow root, and Playwright's text
  matching pierces that root while `innerText` does not, so a guard assertion passed
  over a read that matched nothing. The other 2 are strict-mode violations on
  `.hk-item-body ha-form`, and `main` has since fixed those 2 in a stricter form, so
  only the count read remains here.

## One-way doors

| Surface | Committed shape |
| --- | --- |
| Stored asset | `replace_unit: "uses"`, `replace_also_every: {interval, unit}`, `action`, `use_noun`, `use_task_name` |
| Stored asset | A `uses` target is capped at **250**, so a document carrying a larger one is refused on import |
| Stored task | `recurrence_type: "use"` |
| Stored task | `source.part.role` of `"use"`; **absent means `"replace"`**, and the role is written only for a use task, so no wear part in the wild has its source rewritten |
| Service input | `add_task` / `update_task` accept `recurrence_type: "use"`; `add_asset` / `update_asset` accept the 4 new part keys |
| Event payload | Nothing new. A use is an ordinary `home_keeper_task_completed` — but the *meaning* of a completion on this task type is now "1 use" |
| Panel | The `counted` status-bucket string |
| Behaviour | The backstop is always "whichever comes first". No `any`/`all` combinator, unlike a usage sensor task |
| Behaviour | `use_noun` is stored and rendered as a **plural**, verbatim |
| Behaviour | The completion trim. Once trimmed, history is gone and no export brings it back |

**Not** a one-way door, and cut after being drafted: an `exclude_counted` Profile
option. `profiles.matches_filter` opens with `if task.get("next_due") is None: return
False`, so a use task cannot match a Profile at all — not even a `status: all` one. The
toggle would have been stored state that could never change an outcome.

## Screenshots

Every changed surface at both widths, SHA-pinned to d04031f9800a49b445419642a9990dc84c08dcce, the rebased head. Below 700px the panel is a different layout, not a narrower one, so each pair is shot twice. The rebase changed no panel markup, so these are the images captured before it.

**The Tasks list.** A use task has no due date, so the count chip is the only thing on
the row that says where it stands. Its own scope pill sits beside Shopping, and on a
phone the pills come apart into wrapping chips, which puts Counted on a row of its own.

The Tasks tab scoped to Counted, with the use task reading 17 of 25 wears

The same Counted list on a phone, the pills wrapped and the count chip stacked under the task name

**The part editor.** Setting the unit to uses is what reveals the 3 controls below it:
what the uses are called, what the use task is called, and the time backstop. A wear
item measured in months shows none of them.

The part editor with the unit set to uses, showing the action picker, the use noun, the use task name and the time limit switch

The same counting fields on a phone, stacked to one column

**The appliance part row.** The cadence chip says every 25 uses; the count chip and the
meter beside it say how close that is.

The appliance Parts tab, the coating showing 17 of 25 wears above a progress meter

The same part row on a phone, the row stacked and the meter under the count

They come from `tests/e2e/screenshots-counted.capture.ts`, a harness of its own rather
than steps inside the main walk — the way the NFC, card, declarative and 5 other
captures already work, and what `capture-config.ts` was extracted for. The main walk
photographs about 60 surfaces in one sequence, so a shot near its end costs a 15-minute
run to re-take; these 6 re-take in 13 seconds.

The walkthrough comment above carries the motion: the tour steps through the Counted
section and taps Done, so the count moves on camera.

## Verification

Run locally against the rebased head:

| Gate | Result |
| --- | --- |
| `pytest tests/unit` | 2381 passed, 1 skipped, 1 xfailed |
| `bash ci/test-frontend.sh` | 1425 passed, 45 files |
| `ruff` lint + format | clean over `custom_components tests ci` |
| `vale` | 0 hits added vs `main`, compared as a set per file |
| `check-changelog-release-gap.py` | exit 0 |
| `npm run sync` (site) | 24 User Guide pages, 5 Developer Guide pages, API reference |
| `mypy` | no new errors (see note) |

Playwright e2e (339 passed, the whole suite), the Docker integration tier (10/10 on
`test_use_tasks.py`), both walkthrough tours and both mutation lanes (Python 90.43%,
TypeScript 89.12%, threshold 80%) were green on the pre-rebase head. CI re-runs all of
them here.

**mypy note:** this container's Python is 3.12, below Home Assistant's 3.14 floor, so
pip backtracks to an old release — the #199 trap. Two errors appear (`card.py`
`LOVELACE_DATA`, `transfer_runner.py`), both present without this branch. CI checks it
on a Python at the floor.

## Checklist

- [x] `CHANGELOG.md` updated — with `(Fixes #306)` and a doc link on each lead
      embedded above with an HTML `` tag, desktop and phone for all 3 surfaces
      step through it, and the regenerated comment shows it

https://claude.ai/code/session_01XokgrL1RxPq3WFqYDZGi2w


prestomation:
/q review Be critical and sceptical, and put the most serious findings first — but do not withhold the minor ones. This adds a new recurrence type (`use`) whose whole contract is "never due, nothing arms it", plus a second reconciler-owned task per wear part. Please scrutinise:

**Correctness**
- The `next_due is None` invariant on a `use` task. Is there *any* path that can give one a due date? `recurrence.py` (6 branch points), `models.merge_update`, `store.trigger_task`/`snooze_task`/`skip_task`, `transfer.py` import replay, the sensor watcher, the coordinator's purge.
- `reconcile.uses_since_replacement` — the boundary is `ts > replace_task.last_completed`. Timezone-mixed histories, a completion exactly at the boundary, an unparseable `ts`, a back-dated completion inserted before the last replacement, and undo/edit of either half.
- `trim_use_completions` — can the protected window ever be trimmed? Interaction with `recurrence._record_entry`'s blind 500-entry cap, and with a target that changes after entries accumulate.
- `replacement_backstop_due` — the anchor ladder (`last_completed` → part `last_replaced` → `created`), DST and month-end arithmetic via `add_interval`, and whether count and backstop can ever disagree about which cycle they are in.
- The role gate in `store._stamp_part_replacement`. Is `is_use_task` the right discriminator, and is there a path where a use completion still consumes stock or stamps `last_replaced`?
- Switching a part between `uses` and a time unit in both directions, and deleting the part or the appliance mid-cycle.

**Maintainability**
- `reconcile.py` grew a lot. Are the pure helpers at the right altitude, and is `settle_use_tasks` returning `(to_arm, trimmed)` the right split between pure decision and store-side effect?
- The `role` key: written only for a use task, absent meaning `replace`. Is that asymmetry worth the storage saving, or is it a trap?

**Performance**
- `counted_part_pairs` builds an index over every task on each call, and `settle_use_tasks` runs from both `async_settle_buy_tasks` (12 call sites) and the `SCAN_INTERVAL` tick. Is that O(tasks) per settle acceptable, and are there redundant passes?

**HA best practices**
- `store.settle_use_tasks` calls `trigger_task` in a loop, each of which saves and fires an event. Should that batch?
- The new `_PART_SCHEMA` keys and the `services.yaml` selectors.

**Security** — the new part fields are free text reaching the panel; check the escaping path on the count chip and the part row.

---

prestomation:
/q review Be critical and skeptical. This push adds `tests/e2e/screenshots-counted.capture.ts` (a capture harness of its own, matching the 8 existing ones) and fixes a collapsing-section race in `tests/e2e/tests/counted-wear.spec.ts`. Please look hardest at:

1. **Test robustness.** Both files now use an open-if-shut rule on the `Parts & wear items` `<details>` rather than an unconditional summary click. Is that rule actually race-free, or can the attribute be read before the form settles? The capture also dismisses the first-run banner conditionally on `count() > 0` — is that a TOCTOU against a still-rendering panel?

2. **Coverage vs documentation.** `AGENTS.md` warns that a screenshot is documentation, not verification. Does `counted-wear.spec.ts` really assert on everything the 6 new shots show, or is there a surface that is only photographed?

3. **The main walk.** Steps 61/61c/62/62c/63/63c were removed from `screenshots.capture.ts`. Is anything left orphaned there — an unused import, a state change a later step depended on, a `setViewportSize` that no longer balances?

4. **Correctness of the feature itself**, still: the `next_due is None` invariant on both halves of a counted wear item, the count boundary at exactly the target, the `_stamp_part_replacement` gate, and the time backstop sharing its anchor with the count. Timezone and DST edges on `replacement_backstop_due` especially.

5. **Maintainability and HA best practices** anywhere in the diff.

Most serious first, and please do not withhold the minor ones.

---

prestomation:
Triage of the first review's 5 recommendations. Two were already in the diff, three I am declining with reasoning.

**Escaping — the premise was wrong, but I verified the conclusion.** This panel has no lit-html. It builds strings and assigns `innerHTML`, and `AGENTS.md` requires an explicit `escapeHTML` at every injection point, so "lit-html auto-escapes by default" would have been the wrong thing to rely on. Checked each site the new free text reaches by hand:

- `use_noun` reaches the DOM only through `useCountLabel`, and both renderers escape its output — `chip()` in `utils.statusChipHtml` and `chip()` in `panel-detail`. The meter's `aria-label` and both `aria-value*` attributes are escaped too.
- `use_task_name` becomes the derived task's stored `name` and renders through the existing task-name path.
- `action` never reaches the panel as free text. It is an enum from `PART_ACTIONS`; an unknown value falls back to `replace`, and all it selects is a localized name template.
- No `unsafeHTML` and no markdown pass anywhere on this path.

**3 and 4 are already in the diff.** `card-filter.ts` puts the `counted` check beside the `monitored` one, above `if (!task.next_due) return 'none'` — that unreachable placement is the sixth correction in the PR body. `_PART_SCHEMA` declares all 4 new keys; that omission is the first of the two defects the PR body names, and `test_use_tasks.py` is what caught it.

**1. A guard in `part_role` — declining.** It would convert a benign value into a crash. `role` is absent on every wear part in the wild, and absent means `replace`; an explicit `"replace"` means the same thing and is what a hand-edited export or a future writer would most naturally produce. `part_role` already normalizes anything that is not `"use"` to `replace`, which is the total function this needs. An assertion would make the tolerant reader intolerant and lose the migration story the docstring exists to protect.

**2. Caching the task index — declining as premature.** The review's own condition is "if profiling shows it's slow", and the pass is O(parts x tasks) with parts in the low single digits. A cache here buys nothing measurable and adds an invalidation bug surface.

**5. Decoupling the settle from `reconcile_buy_tasks` — declining, and this one is a user-visible regression.** `SCAN_INTERVAL` is 60 seconds. Moving the settle to the tick alone means tapping Done on the 25th wear leaves the replacement task dormant for up to a minute: the count reads 25 of 25 while nothing is due, and the panel looks broken to the person who just tapped. The coordinator calls the settle in **both** places on purpose — `async_settle_buy_tasks` for the immediate answer after a completion, the tick for the time backstop, which no completion can trigger. The redundant passes are the cheap half of that trade.

The trim behaviour the review asked me to document is in the **One-way doors** table: once `_record_entry` caps the log, that history is gone and no export restores it.

---

prestomation:
/q review Be critical and skeptical, most serious first, and do not withhold the minor ones. This push fixes `tests/e2e/walkthrough.capture.ts` so the tour completes; both tours now pass locally and the gif and mp4 are produced. Three changes, all in that one file:

1. **The counted step reads the chip's `label` attribute** instead of the row's `innerText`. The row text does not carry the count, because `ha-assist-chip` paints its label in its own shadow root — Playwright's text matching pierces it, `innerText` does not. Is `toHaveAttribute('label', /^\d+ of 25 wears$/)` followed by `getAttribute` the right shape here, or is there still a window where the chip is re-rendered between the wait and the read? And is reading a rendered attribute the right source of truth for a tour at all, versus asking the store?

2. **Two `.first()` scopings on `.hk-item-body ha-form`** in the notifications step. That body holds 3 forms, so the unscoped locator was a strict-mode violation on both `toBeVisible` and `toBeHidden`. Please check my reasoning that this was **latent** rather than introduced: I claim it previously passed only by racing the second form's render, and that this branch's extra pauses are what expose it. If that is wrong and something in this diff actually adds a third form, I want to know. Also: is `.first()` the right fix, or should the assertions name which form they mean?

3. Does `toBeHidden()` on `.first()` still mean what the step intends when folding may remove all 3 forms from the DOM rather than hide them?

Beyond this file: please re-examine **whether any of these harness fixes are papering over a real product defect** rather than a test bug — that is the failure mode I am most worried about here. And anything you see in the feature itself that the earlier reviews missed.

---

prestomation:
/q review This branch was rebased onto `main` as a single commit after `0.23.0` shipped. Please be critical and skeptical, and put the most serious findings first without withholding the minor ones.

The rebase is what is new since your last look, so weight it there:

- **`recurrence.py` is the risky part.** 3 conflicts with the fixed-schedule fix (#332) were resolved by taking both sides. Check each one: `complete_occurrence` and `skip_occurrence` call `_advance_fixed_schedule` for `REC_FIXED` and route `REC_USE` to the dormant branch, and `move_completion` now excludes `REC_TRIGGERED, REC_SENSOR, REC_FIXED, REC_USE`. Is any combination now wrong — a fixed task that is also counted, a use task reaching a branch that computes a due date, or `remove_completion` needing `REC_FIXED` too (it excludes only `REC_TRIGGERED, REC_SENSOR, REC_USE`)?
- **The count derivation against the new fixed-schedule behaviour.** The count is the use task's completions filtered by the replacement task's `last_completed`. Does the replacement task's schedule advance differently now that a fixed task holds its local hour, and can that shift the filter boundary?
- **Correctness elsewhere:** the 250 cap against the 500-entry completion trim, timezone and DST on the time backstop, the "whichever comes first" arming, and what happens when a user switches a part's unit from `uses` back to `months` with a use task already carrying history.
- **Maintainability:** `store.reconcile_part_tasks` now keys on `(asset_id, part_id, role)`. Is the absent-role-means-replace default safe for every existing wear part?
- **Security and the privilege boundary:** the new part keys reach `add_asset`/`update_asset`. Is the admin gate on both the websocket command and its service twin, per `docs/SECURITY.md`?
- **Performance:** deriving the count per render, and whether `async_settle_use_tasks` walks every task on every coordinator tick.
- **HA best practices** for the 2-tasks-per-part model and the entity unique IDs it creates.

---

prestomation:
Thanks — 2 of these need a correction, and the rest I agree with.

**Finding 2, the `remove_completion` / `move_completion` asymmetry.** Real, but it is not this branch's, and it is already documented. Both lines come from #332 on `main`: `remove_completion` reads `not in (REC_TRIGGERED, REC_SENSOR)` there, and `move_completion` reads `not in (REC_TRIGGERED, REC_SENSOR, REC_FIXED)`. This branch adds `REC_USE` to each and changes nothing else about them.

The "why" is in `move_completion`'s own docstring, last sentence of that paragraph:

> its due date is schedule state that completing or skipping already moved on (`_advance_fixed_schedule`), not a value derived from the log, so recomputing it from the anchor here would rewind the task onto an occurrence it has already dealt with. **Only `remove_completion` rewinds, because an undo is meant to.**

Moving a completion to a different date is a correction to the log; the occurrence still happened. Removing one says it never happened, so rewinding the schedule is the point of the call. They differ because the 2 operations mean different things, and the docstring says so.

**Finding 4, "switching units loses the task's completion history".** This one is wrong, and it is worth being precise because the opposite would be a data-loss bug.

The replacement task is **updated in place**, not orphaned and recreated. `existing_by_key` is keyed `(asset_id, part_id, role)`, and switching a part from `uses` to `months` leaves the `replace` key in `desired` — only the `use` key drops out. So the orphan path at the top of the loop deletes the use task alone, and the replacement task reaches the update branch, where `recurrence_type` is one entry in `updates` and `merge_update` carries `id`, `completions` and `last_completed` through untouched. The task keeps its identity and its whole history across the switch, in both directions.

What is lost is the **use** task's log, which is the count itself — correct, since there is nothing counting any more — and `store` archives it onto the appliance on delete, as it does for any deleted part-task.

**Findings 1, 3, 5, 6, 7 and 9** match my reading. On 5 in particular, `keep = max(cap, protected)` is the line that matters, and you talked yourself to the right answer: the cap is a floor on what is kept, not a ceiling.

No code change from this review. CI is green on `d04031f`: all 20 checks pass, including e2e, integration, the walkthrough capture and both mutation lanes.

---

prestomation:
Rebased onto `main` again now that #318 has merged. Head is `98d0326`; the version stays `0.24.0b3` and the CHANGELOG section now sits above #318's `0.24.0b2`.

31 files conflicted this time, in 4 groups:

- **The 16 frontend locales.** Both branches add keys and #318 also deletes 4 (`settings.skipsnooze_both` and its 3 siblings, which it replaced by composing the summary from the verbs on offer). I merged them per key, 3-way against the merge base, preserving `main`'s key order rather than re-sorting — `en.json` on `main` has 35 out-of-order pairs, so a sort would have churned 185 lines that nobody changed. The result is additive only: **+411 lines across 16 files, 0 deletions, 0 keys where both sides set the same key differently**, and the 4 keys #318 removed stay removed. `i18n-parity.test.js` passes.
- **`screenshots.capture.ts`.** Both branches scoped the same snooze locator to `.hk-detail-actions`, by different routes. `main`'s version wins: it binds `detailActions` once and reuses it, which is what this branch's comment argued for.
- **`const.py`, `manifest.json`, `CHANGELOG.md`.** Version and section ordering, as above.
- **11 PNGs under `docs/images/`.** Both branches re-shot them. `main`'s win, because they were captured against the UI that now ships, Due today included. They are not this PR's changed surfaces — the 6 `6x-panel-counted-*` images are, and those never conflicted, so the shots embedded in the description are still the ones this branch captured. The cost is that a task-list shot on the docs site is short one seeded row until the next full re-shoot.

`recurrence.py` did **not** conflict this round: its 3 resolutions were already in `d04031f`.

Re-run locally against `98d0326`:

| Gate | Result |
| --- | --- |
| `pytest tests/unit` | 2385 passed, 1 skipped, 1 xfailed |
| `bash ci/test-frontend.sh` | 1435 passed, 45 files |
| `ruff` lint + format | clean over `custom_components tests ci` |
| `vale` | 0 hits added vs `main`, compared as a set per file |
| `check-changelog-release-gap.py` | exit 0 |
| `npm run sync` (site) | 24 User Guide pages, 5 Developer Guide pages, API reference |
| `mypy` | the same 2 stale-HA errors `main` has, none new |

Every check was green on the pre-rebase head. CI re-runs them all here.

---

prestomation:
/q review The branch was rebased onto `main` after #318 merged. Head is `98d0326`. Please be critical and skeptical, most serious findings first, and do not withhold the minor ones.

The rebase is what is new since your last pass, so weight it there:

- **The 16 locale files.** I merged them per key, 3-way against the merge base, rather than taking a side. Check for a key this branch needs that the merge dropped, a key #318 deleted that came back, and any English string that now reads as the wrong feature's copy. `btn.dueToday` and the counted keys should both be present in every locale.
- **Did the merge silently drop a translation?** Every locale should have gained the same set of keys. If one locale is short, `i18n-parity.test.js` may not be strict enough to catch it — say so if that is the case.
- **Interaction between the 2 features.** #318 added `deferVerbs` with an injectable `now` and the rule that due today is withheld on an already-due task. A use task has no `next_due` at all, and a counted replacement task is `REC_TRIGGERED` and dormant until it settles. What do snooze, skip and due today offer on each of those, in the panel and on the card? A verb offered on a dateless task is a bug, and so is a crash in `deferVerbs` on `next_due: null`.
- **`screenshots.capture.ts`.** I took `main`'s scoped snooze locator over this branch's. Confirm the surrounding steps still bind what they mean to.
- **The 11 PNGs I resolved to `main`.** Is any of them load-bearing for a doc page that this branch changes?
- Anything else in the rebased diff you would push back on.

---

prestomation:
Verified each of the "needs verifying" items against the code on `98d0326`. All 4 hold, no change needed, and 3 of them are load-bearing enough that it is worth recording where the answer lives.

**`deferVerbs` on a dateless task.** Safe, and by construction rather than by a special case. `defer.ts` reads `const dormant = !task.next_due`, and all 3 verbs depend on it:

```ts
snooze: allowSnooze && !dormant,
skip: allowSkip && !blocked && !dormant,
dueToday: allowDueToday && !dormant && !isOverdue(task, now),
```

A use task carries `next_due: null` for its whole life, so it is dormant and gets none of the 3. A counted replacement task is `REC_TRIGGERED` and dormant until it settles, so it gets none either, until it arms — at which point it has a real date and the normal rules apply. The panel and the card both call this one function, so there is no second answer to keep in sync.

**The `counted` bucket and `next_due: null`.** Correct, and the ordering is the whole point. `statusBucket` puts `if (isUseTask(task)) return 'counted'` **above** `if (!task.next_due) return 'none'`, beside the `monitored` check rather than beside the `shopping` line it is modelled on. Placing it where the plan originally had it would have made the section unreachable — that is one of the plan corrections in the first commit, and the reasoning is in a comment on those lines.

**The trim past 500.** Handled, and this is exactly why the target cap is 250 rather than `MAX_INTERVAL`. Two guards, in order:

- `recurrence._record_entry` trims blindly to `MAX_COMPLETION_HISTORY` (500) on every write, so the protected window has to fit under it. `MAX_USE_TARGET = 250` is chosen so `2 × target` lands on that cap, and `const.py` says so where the constant is defined.
- `reconcile.trim_use_completions` computes `keep = max(cap, protected)`, so an entry inside the live count is never dropped even when it puts the log over the cap.

Your failure mode — the count truncated so the replacement never arms — cannot be reached. Trimming starts at 500, and the largest target is 250, so by the time any entry is dropped the count has been at or above the target for at least 250 uses and the replacement is already armed. Losing entry 501 takes the count from 501 to 500, which is still ≥ any legal target.

**Plural-category parity.** Already stricter than exact key matching. `i18n-parity.test.js` has a `plural-category completeness` suite that resolves each locale's real CLDR categories with `Intl.PluralRules(lang).resolvedOptions().pluralCategories` and requires every plural base key to define every category **that locale** uses. So French carrying `.many` where English carries only `.one`/`.other` is the test working, not drift, and a French locale that *lost* its `.many` would fail. The suite passes on the merged locales.

**`screenshots.capture.ts`.** `main`'s version binds `const detailActions = panel.locator('.hk-detail-actions')` once and the surrounding steps use it, which is what this branch's comment was arguing for; the 2 skip locators below it are untouched on both sides.

**The 11 PNGs.** None is referenced by `docs/guide/appliances/appliances.md`, the page this branch changes. That page embeds `61-`, `62-` and `63-panel-counted-*`, which are new here and never conflicted.

---

PR #328, opened 2026-09-11
chore(deps-dev): combine the safe Dependabot bumps (playwright 1.63.0, panel vitest 5)

prestomation:
## What changed

Combines the Dependabot pull requests that are safe on their own, and holds back the two that are not.

**Included**

- #322 — `@playwright/test` 1.62.1 → 1.63.0 in `tests/e2e`
- #324 — `vitest` 4.1.11 → 5.0.0 in `custom_components/home_keeper/frontend`

**Deliberately left out: #325 (`vitest` 5) and #323 (`@vitest/ui` 5) on the root package.**

Those two cannot land apart — the root package peer-pins them, which is the only reason each one is red on its own (`npm ci` ERESOLVE, `Conflicting peer dependency: vitest@4.1.11`). Taken together they install cleanly and the root suite passes all 1360 tests. But they break the Stryker mutation gate.

Under vitest 5, `@stryker-mutator/vitest-runner@10` (the latest, which declares `vitest: ">=2.0.0"`) mis-analyses per-test coverage: tests report `covered 0`, fewer tests run per mutant, and the score is not reproducible between runs. Same file, same command:

| | score | killed | survived | errors | tests/mutant |
|---|---|---|---|---|---|
| main, vitest 4 | 100% | 14 | 0 | 0 | 32.2 |
| vitest 5, run 1 | 20.0% | 2 | 8 | 4 | 19.9 |
| vitest 5, run 2 | 42.9% | — | — | yes | — |

This branch changes no mutable source, so `mutation.yml` is a no-op on it and the breakage would have landed unseen, then failed every later pull request that touches the mutable surface. Holding the pair until the runner supports vitest 5.

Note: the panel package's own `npm test` does not work standalone and did not before this change either — `vitest.config.js` lives at the repository root and supplies jsdom, the `panel-version` alias, and the repository-root-relative fixture paths. `ci/test-frontend.sh` runs vitest from the root, so the panel package's `vitest` entry is not exercised by CI.

## One-way doors

None. Development dependencies only; no service, event, storage, or entity surface changes.

## Checklist

- [x] Tests run locally — `bash ci/test-frontend.sh` (44 files, 1360 tests), `bash ci/build-panel.sh`, `bash ci/test-mutation-frontend.sh`
- [x] `CHANGELOG.md` — not needed, developer-only change
- [x] Panel UI changed → no
- [x] New user-facing UI surface → no
- [x] New user-facing feature → no

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #329, opened 2026-09-11
release: 0.23.0, and link the docs from every feature bullet

### LINKED ISSUE #308: Feature request: Create a completed task

### LINKED ISSUE #313: Non-overdue tasks on notification digesy

prestomation:
## What changed

Cuts the stable **0.23.0** from the `0.23.0b2` line, and adds the release rule that
every feature bullet links its documentation.

**The release.** `manifest.json` and `const.py` (`PANEL_VERSION`) go to `0.23.0`, with
a matching `## [0.23.0] - 2026-09-11` section. The section is written for someone
upgrading from **0.22.0**, so the import/export work the two betas introduced is
`### Added` rather than beta-to-beta framing. Both issues that a commit since `v0.22.0`
referenced carry a `(Fixes #N)`, so `notify-issues` comments on each and closes it:
**#308** (import and export) and **#313** (notification triggers). The other three
commits in the range — #315, #316, #328 — are developer-only and correctly get no
bullet.

**The new rule.** An `### Added` bullet writes its bold lead as a Markdown link to the
page that documents the feature:

```markdown
- **[Import and export](https://prestomation.github.io/ha-home-keeper/docs/guide/import-export).**
  Settings has a new *Import and export* card that saves every task and appliance to
  one YAML file, and reads one back.
```

The lead is the right place for it. `summarize()` in `ci/release-issues.py` quotes
**only** the bold lead into the comment the issue reporter gets, so a linked lead
carries the docs link into that comment as well, and a link costs nothing against the
three-sentence bullet budget. The rule is recorded in `RELEASE.md` (the long form, with
how to derive a URL from `website/scripts/doc-map.mjs`), `AGENTS.md`, and
`.amazonq/rules/testing-and-workflow.md`.

`### Fixed` and `### Changed` bullets may link the same way and do not have to; the two
fixes in this release stay plain, because one is covered by a developer page and the
other by a page that did not change.

**The three new links return 404 until this release publishes, and that is correct.**
`docs-deploy.yml` deploys the site when a stable GitHub Release is published, so the
`import-export` guide page and the bullets that link it go live in the same step. The
URL shape is verified against pages that exist today (`/docs/guide/notifications` and
`/docs/guide/settings` both return 200).

## One-way doors

None. No service input, event payload, storage shape, or entity attribute changes. The
changelog link convention is prose, reversible at any time.

## Verification

- `python ci/release-issues.py --version 0.23.0 --json` resolves exactly `#308` and
  `#313`, and `--notes` renders the section intact. The `#308` summary comes back as
  the linked lead, so the issue comment will carry a working docs link.
- `python ci/check-changelog-release-gap.py` passes (the version is bumped).
- `pytest tests/unit` — 2260 passed, 1 skipped, 1 xfailed.
- `npx vitest run tests/frontend/doc-anchors.test.js tests/frontend/changed-pages.test.js`
  — 29 passed.
- `vale CHANGELOG.md` — no hits on any added line. The remaining hits are the
  pre-existing backlog below line 400, which the diff-scoped CI job does not read.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`)
- [x] `CHANGELOG.md` updated — `(Fixes #308)` and `(Fixes #313)` both present
- [ ] Panel UI changed → n/a, no panel source is touched
- [ ] New user-facing feature → n/a, this is the stable cut of features already in
      beta, so no `preview-release` label

https://claude.ai/code/session_01JTTLgTDntPbszPCcPBDRH2

---
_Generated by [Claude Code](https://claude.ai/code/session_01JTTLgTDntPbszPCcPBDRH2)_

### DIFF
--- a/.amazonq/rules/testing-and-workflow.md
+++ b/.amazonq/rules/testing-and-workflow.md
@@ -42,9 +42,11 @@
   user-guide page is `.../docs/guide/<slug>`, where the slug is the section's
   `USER_SECTIONS` entry in `website/scripts/doc-map.mjs`; a deeper anchor adds the
   heading slug. `### Fixed` and `### Changed` bullets may link the same way and do not
-  have to. A link to a page the release itself adds is a 404 until the release
-  publishes, which is correct: `docs-deploy.yml` deploys on stable release
-  publication, so page and link go live together.
+  have to. A link to a page the release itself adds resolves only when a **stable**
+  ships — `deploy-docs` in `release.yml` is gated on `prerelease == 'false'`, so a beta
+  never republishes the site and the bullet's link 404s for beta testers until then.
+  Write it in the feature PR regardless; that is the cost of pinning the site to the
+  latest stable. Nothing validates these URLs, so check the shape against a live page.
 - **Credit an outside contributor in the bullet for their change.** End the bullet
   with `(Thanks @user!)`, after `(Fixes #N)` if the bullet has one. The credit does
   not count against the three-sentence budget. An outside contributor is anyone
--- a/.amazonq/rules/testing-and-workflow.md
+++ b/.amazonq/rules/testing-and-workflow.md
@@ -33,6 +33,18 @@
   buttons the feature hides, what it rewrites internally, which surfaces it touches,
   which fields it added. That is `README.md` material. When trimming a big feature
   loses something real, split it into 2 bullets rather than growing 1.
+- **A feature bullet's bold lead links its documentation.** Every `### Added` bullet
+  writes the lead as a Markdown link to the page documenting the feature, e.g.
+  `**[Import and export](https://prestomation.github.io/ha-home-keeper/docs/guide/import-export).**`.
+  It costs nothing against the three-sentence budget, and `summarize()` quotes the
+  lead into the issue reporter's comment, so the reporter gets the docs link too. Use
+  the absolute site URL — the bullet is read on GitHub and in a release body. A
+  user-guide page is `.../docs/guide/<slug>`, where the slug is the section's
+  `USER_SECTIONS` entry in `website/scripts/doc-map.mjs`; a deeper anchor adds the
+  heading slug. `### Fixed` and `### Changed` bullets may link the same way and do not
+  have to. A link to a page the release itself adds is a 404 until the release
+  publishes, which is correct: `docs-deploy.yml` deploys on stable release
+  publication, so page and link go live together.
 - **Credit an outside contributor in the bullet for their change.** End the bullet
   with `(Thanks @user!)`, after `(Fixes #N)` if the bullet has one. The credit does
   not count against the three-sentence budget. An outside contributor is anyone
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -80,10 +80,16 @@
   — and a deeper anchor is that page plus the heading slug. Link the nearest page that
   says what the feature does, not the repository README and not a `docs/*_PLAN.md`.
   A `### Fixed` or `### Changed` bullet may link the same way when a page covers it,
-  and does not have to. **A link to a page that a release adds is a 404 until that
-  release publishes**, which is correct: `docs-deploy.yml` deploys the site on stable
-  release publication, so the page and the link go live together. Check the shape of
-  an existing URL rather than the new one.
+  and does not have to. **A link to a page the release itself adds resolves only when a
+  _stable_ ships**, because `deploy-docs` in `release.yml` is gated on
+  `prerelease == 'false'` and a beta never republishes the site. Write the link in the
+  feature PR anyway, since that is when the author knows which page documents the
+  feature and the stable cut rolls the bullet up unchanged — but the bullet ships first
+  in a `## [X.Y.ZbN]` section whose link 404s for beta testers until the stable
+  publishes. That is the cost of pinning the site to the latest stable, not an
+  oversight. Prefer an existing page when one already covers the feature, and check the
+  shape of a URL that is already live, because nothing validates these links and a typo
+  404s forever.
 - **Credit an outside contributor in the bullet for their change.** End the bullet
   with `(Thanks @user!)`, after `(Fixes #N)` if the bullet has one. The credit does
   not count against the three-sentence budget. An outside contributor is anyone
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -68,6 +68,22 @@
   because `(#N)` is also the squash-merge PR number and the two can't be told apart.
   The job posts a CI warning naming any issue a shipped commit referenced that the
   section forgot.
+- **A feature bullet's bold lead links its documentation.** Every `### Added` bullet
+  writes the lead as a Markdown link to the page that documents the feature, e.g.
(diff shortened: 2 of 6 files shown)


prestomation:
/q review This is the stable 0.23.0 release cut plus a new changelog convention. Please be critical and skeptical, and surface the most serious issues first without withholding minor ones.

Topics I want scrutinized:

**Release correctness (highest value).** Does the `## [0.23.0]` section actually satisfy what `release.yml` and `ci/release-issues.py` require? Specifically: is the bold lead written as a Markdown link (`**[Text](url).**`) still matched by `_BOLD_LEAD` (`^\*\*(.+?)\*\*`) the way I claim, and does the resulting `summary` render correctly inside the issue comment template? Check whether a `]` or `)` in the quoted summary can break the comment's Markdown. Also confirm `manifest.json` and `const.py` agree, and that the section heading format matches what the workflow greps for.

**Issue completeness.** I claim #308 and #313 are the only issues fixed by commits since `v0.22.0`, and that #315, #316 and #328 are developer-only. Verify against the git log — a missed issue is never notified and never closes, which is the expensive failure here.

**Stable-section framing.** `AGENTS.md` says a stable section describes what changed since the last *stable*, not since its betas. Did I roll the b1/b2 work up correctly? Is `**External IDs.**` a better lead than b1's `**A stable key of your own.**` under the "a lead is a heading, drop articles and prepositions" rule, and do the three Added bullets stay inside the three-sentence budget counting the lead?

**The new convention itself.** Is "link from the bold lead" the right call, versus a trailing link or a separate line? Consider how it reads in the GitHub release body, in the HACS changelog view, and in the issue comment. Does linking a page that 404s until the release publishes create a real problem I have waved away — for example, does anything validate changelog links, or would a beta release ever quote a link to a page that never ships?

**Rule consistency.** The rule is now in three files (`RELEASE.md`, `AGENTS.md`, `.amazonq/rules/testing-and-workflow.md`). Do they contradict each other or the surrounding rules? Is the `DOC_ROUTES` reference in `RELEASE.md` accurate for deriving a Developer Guide URL, and is the `/docs/guide/<slug>` derivation from `USER_SECTIONS` correct?

---

prestomation:
Thanks — the beta-404 finding is correct and I have fixed it in `d3282ca`. Triage of the rest below.

### Fixed: the beta case (your "most serious")

You are right, and I was wrong to write "which is correct" without qualifying it. `deploy-docs` in `release.yml` is gated on `prerelease == 'false'` (line 139), so a beta never republishes the site. That matters more than the finding suggests: `AGENTS.md` requires a feature PR to cut a beta, so a feature bullet's **first** appearance is always a `## [X.Y.ZbN]` section — the 404 is the common case, not an edge one.

I kept the convention and stated the cost honestly rather than changing it. Writing the link in the feature PR is still right: that is the moment the author knows which page documents the feature, and the stable cut rolls the bullet up unchanged. The 404 in between is the price of pinning the site to the latest stable so nobody reads docs for an unreleased feature — a deliberate project trade. All three files now say so, and also say to prefer an existing page when one already covers the feature.

I did not take the "temporary preview docs build for betas" option. Every PR already gets a preview at `pr-preview/pr-<n>/`, but that URL is ephemeral and deleted on close, so baking it into a permanent changelog bullet trades a temporary 404 for a permanent dead link.

### Fixed: `DOC_ROUTES` derivation

Split into its own bullet with the value spelled out (`docs/INTEGRATING.md` → `/developer/integrating`), plus a note that the generated API reference comes from `GENERATED_DEV_PAGES`, which carries its `route` directly. Also recorded that nothing validates these URLs, so a typo 404s forever.

### Verified, no change needed: issue completeness

Ran the repo's own cross-check rather than relying on my earlier reading:

```
$ git log --format=%B 86d7bce..HEAD | python ci/release-issues.py --version 0.23.0 --missing
[]
```

Empty, so every issue a commit in the range references is in the section. #308 and #313 confirmed.

### Declining, with reasons

**Markdown injection.** You checked `body()` yourself and found it builds `> ${summary}`, a blockquote, not link syntax — so there is no injection path. Documenting a constraint against a shape the code does not have would be a comment about hypothetical future code, and it would go stale before it ever applied.

**Cross-check is warning-only.** True, and it predates this PR — `notify-issues` runs after the release is already tagged, so it structurally cannot gate anything. Turning it into a pre-tag gate is a real idea, but it is a change to the release workflow, not to this release, and widening a release PR to carry it is how a release PR goes red. Worth its own issue.

**A CI job that curls changelog links.** Same reasoning, and it has a design problem: the links it would check are 404 by design until the stable publishes, so the job can only run post-release, where it warns about something already shipped. The rule now tells the author to check the shape against a live page, which catches the typo at the only point where it is cheap to fix.

**Rule consistency / DRY.** The three files are deliberately parallel, not deduplicated: `AGENTS.md` and `.amazonq/rules/` exist so Claude and Amazon Q each load conventions from a file they read natively, and `AGENTS.md` says to keep them consistent rather than to make one cite the other. You confirmed no contradiction, which is the actual bar.

---

prestomation:
/q review Second pass, please — `d3282ca` responds to your first review. Be critical and skeptical again, most serious first, and do not withhold minor findings.

What changed since your last look: the docs-link rule in `RELEASE.md`, `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md` now states the beta case you flagged (a beta never republishes the site, so the bullet's link 404s until the stable), and the Developer Guide URL derivation moved into its own bullet naming the `DOC_ROUTES` value and `GENERATED_DEV_PAGES`.

Specifically, please check:

**Is the beta framing now actually accurate?** I claim `deploy-docs` is gated on `prerelease == 'false'` and that a beta therefore never republishes the site. Verify that against `release.yml` and `docs-deploy.yml` — is there any other trigger that would republish the production site (a push to `main`, a workflow_dispatch, a schedule) that would make my "404 until the stable" claim wrong in the other direction? An overstated cost is a defect too.

**Did I pick the right resolution?** I kept the convention and documented the cost rather than dropping links from beta bullets or pointing them at the ephemeral `pr-preview/pr-<n>/` build. Argue the other side if you think one of those is better.

**Did the rewrite introduce drift between the three files?** They were consistent before; I edited all three by hand. Check that they still say the same thing and that none of them now contradicts a neighbouring rule in its own file.

**Are my three declines defensible?** I declined documenting the Markdown-injection constraint (no injection path exists in `body()`), a pre-tag issue cross-check gate, and a CI job curling changelog links — the latter two as out-of-scope for a release PR. Push back if any of those is cheap enough that declining it is the wrong call.

**Anything in the release itself you would still block on**, now that the section, the version bump and the issue list have all been checked once.

---

Issue #101, opened 2026-06-28
[FR] Improvements for the Dashboard Task Card


psym88:
Hi,

First of all, thank you for creating this amazing integration. It provides everything I've been looking for in a Home Assistant maintenance tracker. Great work!

I have two feature requests for the Dashboard Task Card:

**Option to disable the inline edit form**
Please add an option to disable the behavior where tapping a task row opens the inline "add/edit/delete" form. For less technical users, this can be confusing, and it's easy to accidentally delete a task instead of simply marking it as completed.

**Display the device documentation link**
It would be great to have an option to display the documentation link of the associated device directly within the Task Card. This would make it much easier to access manuals or maintenance instructions while performing a task.

Finally, I really hope you'll continue maintaining this project. It's exceptionally well thought out, fills a real gap in the Home Assistant ecosystem, and has great potential to become the go-to maintenance tracker for many users.

Thanks again for your work and for considering these suggestions!

prestomation:
Thank you for the kind words. I really appreciate it! Your suggestions make sense, let me see what I can do

prestomation:
I have a preview for you. The PR is in #102, feel free to comment there or here. In HACS if you select Redownload and manually pick v0.6.0.dev102 you can try it out before release.

psym88:
Thanks so much!

**Option to disable the inline edit form**
This looks really great. It provides a much easier experience on mobile devices.

**Display the device documentation link**
Being able to attach the device documentation to a task is a fantastic idea. It keeps it very flexible.

I can add configured links to a task, but unfortunately uploaded documentation files are not available to select.
If you could have a look at that as well, it would be greatly appreciated.

prestomation:
Great catch!  I've added this. If you force a redownload of the same release you should see uploaded files as options

psym88:
Tested and works great.

**Note:** Uploaded documents do not open in the Home Assistant Companion App on mobile. This is acceptable, as opening them in an unauthenticated mobile browser would likely result in a poor user experience.

Thanks so much for your work!

prestomation:
Interesting. Could you tell me more about which phone you're using, if you're using the browser or app? In my testing on my Samsung phone, using chrome opening the pdf works as expected. Happy to dig into this one.

I really do appreciate the kind words and the requests. I just built this in the last week so even I haven't put it through its paces. I look forward to any bug reports or feature requests you have. I do intend on maintaining this as I intend on using it. That's why I built it.

Thanks again!

[...]

PR #333, opened 2026-09-12
Split the User Guide out of README.md and add user quotes

prestomation:
## What changed

Two things, one structural and one editorial.

### 1. The User Guide moves out of `README.md`

The README was **1,952 lines** because it was the canonical source for all 24 User
Guide pages — `sync-docs.mjs` split it by `## ` section. A visitor landing on the
repository got the whole manual instead of "what is this, how do I install it,
where are the docs".

The guide is now **one authored file per page** at `docs/guide/<group>/<slug>.md`,
mirroring the existing `GUIDE_GROUPS`. Content moved byte-for-byte apart from a
one-level heading shift and the link rewrites below.

| Before | After |
|---|---|
| `README.md` split by `## ` section | `docs/guide/**/*.md`, one file per page |
| `USER_SECTIONS[].h` joins on README heading text | entry names a file on disk |
| `ANCHOR_ROUTES` maps cross-section `#anchor`s | relative `../group/page.md#anchor` links |
| `unlistedReadmeSections()` guard | `guideFileDrift()` guard |

- **Every live URL is preserved.** The frontmatter `slug: /guide/<slug>` is
  unchanged, and the build emits the same 24 routes.
- **Cross-page links are relative `.md` paths**, so they work when reading
  `docs/guide/` on GitHub, and `rewriteLinks()` turns them into site routes via a
  new `GUIDE_ROUTES` map. This replaces `ANCHOR_ROUTES` entirely — the resulting
  link set is identical to what that map produced.
- **The drift guard survives.** `guideFileDrift()` fails the site build both ways:
  a file with no `USER_SECTIONS` entry, and an entry naming a file that does not
  exist. `tests/frontend/doc-anchors.test.js` fails first, as before.
- `changed-pages.mjs` loses its README section-diffing; every source is now one
  file mapping to one page, so `docs-preview.yml` no longer needs the
  `git show $BASE_SHA:README.md` plumbing.

The new README is **131 lines**: intro, screenshot, quotes, the feature list,
installation, doc links, Development.

### 2. User quotes on the README and the docs landing page

Three quotes, verbatim, each linked to the post it came from — two from GitHub
issues, one from the Home Assistant community forum.

| Quote | Source |
|---|---|
| "It's exceptionally well thought out, fills a real gap in the Home Assistant ecosystem…" | @psym88, issue #101 |
| "So happy to finally see such a thorough solution to this problem…" | @peng1can, HA Community |
| "It is currently much closer to the maintenance tracker I had been looking for…" | @Tomblarom, issue #305 |

On the site they render as a card band below the feature grid on `/`
(`src/pages/index.tsx` + `index.module.css`), all colours from Infima tokens so
dark mode follows for free. `docs/intro.md` is unchanged.

The @psym88 quote trips `ai-tells.OverusedVocabulary` on "exceptionally". Quotes
are other people's words, so the README disables that rule for the block rather
than editing what someone wrote.

### 3. Policy files that forbade this

`AGENTS.md`, `.amazonq/rules/testing-and-workflow.md` and `website/README.md` all
said the README must stay the comprehensive user doc and must not be slimmed.
They now name `docs/guide/` as the place new feature documentation goes.
`.vale.ini` lints `docs/guide/**/*.md` so the guide keeps its prose gate.

## Verification

- `npm run build` (website) — succeeds with `onBrokenLinks: throw` and
  `onBrokenAnchors: throw`, so every relative link and anchor resolves.
- All 24 `/docs/guide/<slug>` routes emitted, matching the current site.
- `bash ci/test-frontend.sh` — **1350 passed**, 44 files.
- `vale README.md docs/guide` — **clean**. Baselined against `main`'s README
  (also clean) rather than the diff, since moved lines all read as added.
- Landing page inspected at desktop and phone width, in light and dark.

`ci/generate_schema.py` fails in this session's `.venv` (Python 3.12 resolves an
HA without `LOVELACE_DATA`). Confirmed pre-existing by stashing the branch — this
is the #199 trap AGENTS.md documents, not a regression. A stub file was used for
the local build only.

## One-way doors

None. No service field, event payload, storage shape or entity attribute changes.
The `/docs/guide/<slug>` URLs — the only external contract in scope — are
deliberately preserved unchanged.

## Checklist

- [x] Tests run locally (`bash ci/test-frontend.sh`)
- [ ] `CHANGELOG.md` updated — **n/a**: no user-facing integration change, so this
      is a developer/docs change
- [ ] Panel UI changed → **n/a**: `custom_components/home_keeper/frontend/src/` is
      untouched
- [ ] New user-facing UI surface → **n/a**
- [ ] New user-facing feature → **n/a**, no beta bump or `preview-release` label

https://claude.ai/code/session_01WqxWmRdKsap9EDf5226Td2

---
_Generated by [Claude Code](https://claude.ai/code/session_01WqxWmRdKsap9EDf5226Td2)_


prestomation:
/q review Please be critical and skeptical, and surface the most serious issues first without withholding minor ones. This PR flips the User Guide from "generated by splitting README.md" to "one authored file per page under docs/guide/", and adds user quotes to the README and the docs landing page.

Focus on:

**Correctness of the generator change** (`website/scripts/sync-docs.mjs`, `doc-map.mjs`):
- `guideFilesOnDisk()` uses `readdir(recursive, withFileTypes)` and reads `e.parentPath ?? e.path`. Is that stable across the Node versions CI runs? Does it produce repo-relative POSIX paths on every platform?
- `guideFileDrift()` replaced `unlistedReadmeSections()`. Are there drift cases the old guard caught that this one misses — a duplicate slug across two groups, a file moved between groups, a non-`.md` file, a nested subdirectory?
- `rewriteLinks()` now resolves `GUIDE_ROUTES[rel] ?? DOC_ROUTES[rel]`. Can a guide file shadow a developer doc, or vice versa? Is the `sourceDir` change from `''` to `docs/guide/<group>` correct for every link form (images, `../../EVENTS.md`, `custom_components/...`, anchors)?

**Link integrity:** ~20 cross-section anchors became relative `.md` links and `ANCHOR_ROUTES` was deleted. Do all of them resolve both on GitHub (reading `docs/guide/` in-repo) and on the site? Any link form the rewrite misses — reference-style `[ref]: path`, raw HTML `<a href>`, links split across a newline?

**URL preservation:** every `/docs/guide/<slug>` route must be byte-identical to what `main` serves. Please check I did not silently drop, rename or reorder one, and that sidebar positions still match.

**Content fidelity:** the 24 files should be the README sections verbatim apart from a one-level heading demotion. Please spot-check for content lost at section boundaries, mangled fenced code blocks, or headings demoted inside a fence.

**Maintainability:** is `USER_SECTIONS` still the right shape now that the slug is also the filename — is anything now duplicated between the map and the filesystem that could drift?

**Tests:** `tests/frontend/doc-anchors.test.js` was rewritten. Does it actually cover the failure modes the old anchor guard covered, or did coverage shrink?

Also: the README disables `ai-tells.OverusedVocabulary` for the quotes block because the quotes are other people's words. Is a scoped inline disable the right call there, and is the scope tight enough?

---

prestomation:
Thanks — triaged all of it. Pushed `ce12e3a`. Two changes, and one real defect your content-fidelity prompt led me to.

### 🔴 Node compatibility — fixed, though not for the stated reason

The specific claim is wrong: `parentPath` landing in v21.4 would not have broken CI, because the expression was `e.parentPath ?? e.path` and the docs jobs pin `node-version: 20`, where `Dirent.path` exists (v20.12+). It would have fallen back cleanly.

The underlying point is right anyway, for a reason the review didn't name: **`website/package.json` declares `engines.node >= 18`**, and `readdir`'s `recursive` option didn't land until **v20.1** — so on a declared-supported Node the old code returned only top-level entries and the drift guard would have silently passed while the build wrote nothing. That's worse than a crash.

Also agreed on the test/production discrepancy. Both now call one `guideFilesOnDisk()` exported from `doc-map.mjs`, a hand-rolled recursive walk using nothing newer than Node 18:

```js
export function guideFilesOnDisk(repoRoot, dir = 'docs/guide') {
  const files = [];
  for (const entry of readdirSync(posix.join(repoRoot, dir), {withFileTypes: true})) {
    const rel = posix.join(dir, entry.name);
    if (entry.isDirectory()) files.push(...guideFilesOnDisk(repoRoot, rel));
    else if (entry.name.endsWith('.md')) files.push(rel);
  }
  return files.sort();
}
```

### 🔴 Real defect found — orphaned vale directives (not in your list)

Your "manual spot-check recommended" on content fidelity was worth acting on, so I diffed all 24 files against `main`'s README line by line, normalising heading level and link targets. Result: **1332 lines in, 1332 lines out**, identical apart from two intentional edits.

But it surfaced a genuine bug the split caused. `main`'s README suppressed `ai-tells.OverusedVocabulary` around the Notifications heading with an inline comment pair:

```
<!-- vale ai-tells.OverusedVocabulary = NO -->
## Notifications (actionable reminders on your phone)
<!-- vale ai-tells.OverusedVocabulary = YES -->
```

Splitting on `## ` put the `= NO` at the end of `todo-sync.md` and the `= YES` at the top of `notifications.md` — a dangling disable in one file and a re-enable of something never disabled in another. Both removed. The heading is now just `# Notifications`, and the one real use of the word (`services.md`) has its own correctly-paired suppression. Vale is clean.

This is the class of problem worth calling out: **a paired inline directive spanning a section boundary**. I checked the rest — the other pairs (`appliances.md`, `services.md`) are intra-file.

### ⚠️ Drift gaps — two are non-issues, one is now tested

- **Duplicate slugs across groups** — already caught, just not where you looked. The slug *is* the URL, and `USER_SECTIONS` is where slugs come from; `doc-anchors.test.js` asserts uniqueness there. Two files in different groups with the same slug would each need an entry, and the second collides. Added an assertion that the file count matches the entry count, so the two can't diverge.
- **Non-`.md` files** — deliberate. A stray `.DS_Store` should be ignored, not fail the build.
- **Nested subdirectories** — this one is real but already behaves correctly: `guideFile()` only builds two-level paths, so `docs/guide/tasks/sub/nested.md` can never be listed and the walker reports it as `unlisted`, failing the build with a clear message. Added a test pinning that, since "fails loudly" vs "silently ignored" was not previously proven either way.

### ✅ Points I agree with and took no action on

Reference-style `[ref]: path` and raw `<a href>` links aren't matched by `rewriteLinks()` — correct, but that's unchanged pre-existing behaviour, and I verified no guide file uses either form (`grep -rnE '^\[[^]]+\]: '` over `docs/guide/` is empty). Widening the regex is a separate change.

Heading demotion is safe: all 24 files have exactly one `# ` at line 1, and the demotion script was fence-aware, so no `#` inside a code block moved.

### Verification on `ce12e3a`

- `npm run sync && npx docusaurus build` — clean, with `onBrokenLinks: throw` and `onBrokenAnchors: throw`
- 24 `/docs/guide/<slug>` routes, unchanged
- `bash ci/test-frontend.sh` — 1352 passed (up 2)
- `vale README.md docs/guide` — clean

---

prestomation:
### Walkthrough capture — pre-existing, not this PR's

The `walkthrough` soft gate reported a failed capture. It is not caused by this PR, and I am not fixing it here. Evidence and a proposed patch below.

**What fails.** `tests/e2e/walkthrough.capture.ts:907`, in `openNotifyRow`, all three desktop attempts:

```
Error: strict mode violation: locator('home-keeper-panel').first()
  .locator('#hk-notifications').locator('.hk-item-card').first()
  .locator('.hk-item-body ha-form') resolved to 3 elements:
    1) <ha-form> ... filter({ hasText: 'Name* Profile *Household' })
    2) <ha-form> ... filter({ hasText: 'Send to (mobile apps) Mark done ... Snooze ... Open' })
    3) <ha-form> ... filter({ hasText: 'Send when overdue Send when' })
```

**Why it is not this PR's.** The diff touches no file under `tests/e2e/` or `custom_components/` — only `docs/guide/`, `website/`, `tests/frontend/`, `.vale.ini`, `AGENTS.md`, `.amazonq/rules/` and `.github/workflows/docs-preview.yml`. And the identical failure — same locator, same `resolved to 3 elements`, same line 907, same three desktop retries, same `CAPTURE_OUTCOME=failure / HAS_DESKTOP=no` — already occurred on **2026-09-11 17:52 UTC** on the unrelated branch `claude/github-issue-331-he7o0u` at sha `0a75d89` ([run 34630014712](https://github.com/prestomation/ha-home-keeper/actions/runs/34630014712)), before this PR existed. The tour is broken on `main`.

No fix for it exists on any branch to port, so per the "don't widen the PR" rule here is the proposed patch rather than a push:

```diff
-    await expect(notifyRow.locator('.hk-item-body ha-form')).toBeVisible();
+    await expect(notifyRow.locator('.hk-item-body ha-form').first()).toBeVisible();
```

The assertion is a readiness gate — "the row expanded and its body is rendering a form" — not a claim about which form. The notifications row now renders three (identity, delivery actions, send-when), so the un-narrowed locator became ambiguous when the third was added; `.first()` restores the original intent. Happy to push it here if you would rather not carry it separately, but it belongs in its own PR by the repo's own rules.

**One unrelated thing that run also exposed:** the phone tour failed its first attempt on a different locator and only passed on retry:

```
Locator: locator('home-keeper-panel').first().locator('#hk-task-form')
Expected: visible — Timeout: 15000ms — element(s) not found
```

at `walkthrough.capture.ts:992` in `phoneTour`, right after `panel.locator('#add-btn').click()`. That one looks like a genuine race rather than an ambiguous selector, and is worth a separate look. Also note `HAS_PHONE` reports `no` even when the phone tour passes, because the sticky-comment step reports no media at all once the desktop capture fails — so that flag is not an independent signal today.

### CI on `ce12e3a`

Green: `ruff`, `vale`, `mypy`, `hacs`, `TypeScript`, `Python`, `changelog-release-gap`, `pytest and coverage report`, `release`. Still running: `test`, `integration`, `e2e`, `preview`, `walkthrough`. Nothing has failed.

---

prestomation:
/q review Pushed `8190b8d`, which adds an always-green rule to the instructions, makes the walkthrough capture a hard gate, and fixes the tour. Please be critical and skeptical, most serious issues first, and do not withhold minor ones.

**The hard gate (`.github/workflows/walkthrough-preview.yml`).** I deliberately KEPT `continue-on-error: true` on the capture step and added a final step that fails the job on `steps.capture.outcome == 'failure'`, because removing continue-on-error would skip every later step that is not `always()`/`!cancelled()` — the HA log dump, the "Note which tours were captured" step, and the sticky comment — so the run would go red with none of the diagnostics that say why. Please check:
- Is `if: ${{ !cancelled() && github.event.action != 'closed' && steps.capture.outcome == 'failure' }}` correct for every path — PR closed, run cancelled by the concurrency group, capture skipped, fork PRs where the job's `if` at the top already excludes it?
- Does placing it last actually guarantee the sticky comment posts before the job fails?
- Is there a path where the capture genuinely fails but `steps.capture.outcome` is not `'failure'`, so the gate silently passes? That would recreate the exact hole this is meant to close.

**The tour fixes (`tests/e2e/walkthrough.capture.ts`).** A notify row's `.hk-item-body` holds three `ha-form`s (identity, delivery, and the `notifyTriggerGroup` indent), so the bare descendant locator was ambiguous.
- Open gate is now `notifyRow.locator('> .hk-item-body ha-form').first()`. Is `.first()` the right readiness semantic here, or should it target a specific form?
- Collapse assertion is now `expect(notifyRow.locator('> .hk-item-body')).toBeHidden()`. I chose this because the header toggle sets `body.style.display = 'none'` on the single `.hk-item-body` per card (`panel-settings.ts:843`). Is the direct-child combinator `> .hk-item-body` reliable inside a Playwright chained locator, and is there any nesting case where a row contains a second `.hk-item-body`?
- The phone Add step now retries the click up to 3 times, 5s each. Does that change the recorded video's pacing in the passing case (I believe not — the loop exits immediately when the form appears)? Is a bounded `for` loop the right idiom here versus `expect.poll`/`toPass`?

**Verification honesty — please probe this.** I could NOT run the desktop tour to completion locally: `seedBuyReminder` does not produce the buy task in this sandbox, so the tour stops earlier at the shopping bucket (`walkthrough.capture.ts:173`). What I do have: the first local capture reproduced the CI failure exactly and, with the open-gate fix applied, progressed past it to the collapse assertion — which is direct evidence the first fix works and is how I found the second. The phone tour passed on first attempt in all five local runs. So the collapse fix is reasoned from source rather than executed. If you can spot a flaw in that reasoning, say so — the gate is now hard, so a wrong fix turns the check red.

**Instructions (`AGENTS.md`, `.amazonq/rules/testing-and-workflow.md`).** Do the new rules contradict anything already written there? I reworded the "Wait for CI" line and both "soft gate" passages; please check I did not leave a stale claim elsewhere that still says capture does not block.

---

prestomation:
Good catch on the stale message — that was the worst place to leave the old behaviour written down. Pushed `ccaa003`.

**1. Stale sticky-comment text (your top finding) — fixed.** The comment told the author "This is a soft gate and does not block the PR" while the check beside it now fails the run. It is the one sentence someone reads when the capture breaks, so it was actively misleading:

```diff
-`⚠️ The walkthrough capture failed for this push — see the [run logs](…). This is a soft gate and does not block the PR; pushing again re-runs it.`
+`⚠️ The walkthrough capture failed for this push — see the [run logs](…). This check is red until the tour runs clean: fix the tour, or push again if it was a flake.`
```

**2. The `'skipped'` hole — documented at the gate.** You are right that this is how the gate gets reopened, and a warning costs nothing:

```yaml
      # This reads `outcome == 'failure'`, so it only fires when the capture actually
      # ran and failed. Give the capture step an `if:` of its own and a skip makes
      # `outcome` 'skipped', which passes this gate silently — the exact hole the gate
      # exists to close. Gate on the skip too if the capture ever becomes conditional.
```

I did not widen the condition to `!= 'success'`, because that would fire on `'cancelled'` too and re-break the superseded-run case the sticky comment's own comment block explains. The warning is the right weight for a hazard that only appears if someone edits the capture step.

**3. Error swallowing in the retry loop — leaving as is, with reasoning.** `waitFor(...).catch(() => {})` is load-bearing rather than sloppy: a timeout there *means* "the sheet is not open yet, press again", which is the loop's whole purpose, so catching it is correct. A non-transient error is not lost either — the loop ends on `expect(form).toBeVisible()`, which surfaces it with a proper Playwright message. Logging inside a video capture would also put noise in a run whose output is a recording.

**On your probe of my collapse-fix reasoning** — your three failure modes are the right ones. #1 is settled (`panel-settings.ts:843`, and the body is created once per card at `:741`). For #3, the direct-child combinator is what rules out a wrapper appearing between the row and its body, which is why I used `> .hk-item-body` rather than the descendant form. #2 would need a CSS rule overriding an inline `display: none`, which nothing in `panel-styles.ts` does.

Worth being plain about one thing: I still have not executed the collapse assertion. `seedBuyReminder` does not produce the buy task in my sandbox, so the desktop tour stops at the shopping bucket well before it. This push re-runs the now-hard gate, which is the real check — if the reasoning is wrong, the check goes red and I fix it rather than it passing quietly.

---

PR #334, opened 2026-09-12
fix: counted wear item skips, card sections, and a snoozed fixed schedule

prestomation:
## What changed

Five defects found by reviewing the unreleased `0.24.0b1`–`b3` range
(`v0.23.0..main`), plus 4 smaller corrections found alongside them. Each fix has a
failing test written first.

### 1. Skip on a counted wear item bounced back within seconds

`recurrence.skip_occurrence` leaves `last_completed` alone by contract and sets
`next_due` to `None`. That is the exact state `reconcile.settle_use_tasks` arms
from, so the next coordinator refresh re-armed the replacement task and fired
`home_keeper_task_triggered` again. Skip was a no-op that sent the reminder back.

Reproduced against the running container:

```
rep.next_due before skip: 2026-09-12T01:10:45-04:00
skip -> ok
  +2s next_due=2026-09-12T01:11:40-04:00     *** RE-ARMED ***
```

`reconcile.cycle_start` now takes the later of the replacement task's
`last_completed` and its most recent skip, and both `uses_since_replacement` and
`replacement_backstop_due` measure from it. A skip starts the next cycle the way a
completion does. This is the same failure `store.skip_task` resets a usage meter
for (#268), and the same answer.

Same probe after the fix:

```
skip -> ok
  +3s .. +24s next_due=None    (the skip holds)
one use after the skip -> rep.next_due: 2026-09-12T01:21:03-04:00
```

### 2. A use task vanished from a `group_by: status` dashboard card

`statusBucket` gained a `'counted'` return in #321, but `STATUS_ORDER` in
`card-filter.ts` never gained the row. `groupTasks` builds its sections from that
table alone, so a use task matched no section and was dropped. The panel's own
table (`panel-controls.ts`) already had the entry. A card holding only counted
tasks rendered an empty body, with no empty-state message either.

### 3. `filter: counted` plus `horizon_days` returned an empty card

The horizon exempted `no_due` and `shopping`, the 2 dateless filters that existed
when it was written. `counted` selects use tasks, which are dateless for life.

### 4. Completing a snoozed fixed task skipped an occurrence

`_advance_fixed_schedule` reads `next_due` as "the occurrence the task is
showing". After a snooze that is false: `store.snooze_task` writes an arbitrary
instant. Snoozing a Monday task 10 days out and then marking it done advanced past
the snooze target rather than past the occurrence.

```
weekly Monday 09:00, snoozed to Jan 15, completed Jan 5 10:00
  before: next_due = 2026-01-19   (the Jan 12 occurrence is gone)
  after : next_due = 2026-01-12
  control, no snooze: 2026-01-12  (unchanged)
```

`_is_occurrence` now decides whether `next_due` is on the grid. An off-grid date
reads as "no occurrence on the board", which is the `now`-only branch. #331 stays
fixed, and an overdue task still collapses its missed occurrences in one step.
This one is a regression from #332: `v0.23.0` returns Jan 12.

### 5. A cleared time-limit box failed the whole appliance save

`mergePartForm` used `??`, which does not catch `0`, and `Number('')` is `0`. A
cleared or zeroed backstop box sent `replace_also_every.interval: 0`, which
`assets._normalize_replace_also_every` refuses. The entire appliance save failed
over one field.

### Smaller corrections

- `countedProgress` now excludes a manual part link, matching
  `reconcile.counted_part_pairs` and `reconcile_part_tasks`. `set_task_consumable`
  lets any task point at any part, so the panel could measure the count from a
  user-owned task and show a figure the reminder does not act on.
- `usesSinceReplacement` drops an unreadable timestamp once the cycle has a
  marker, matching `reconcile._sortable`. The comment claiming parity was wrong in
  the other direction.
- `settle_use_tasks` leaves a **disabled** replacement task alone, matching
  `sensor_watcher`. Arming one fired a device trigger nobody asked for.
- A use task's page reads `Counting` rather than a dash for its due date, which is
  what `dueLabel` already answers everywhere else.
- `ci/test-python-integration.sh` turns off the `homeassistant` pytest plugin. Its
  autouse fixture calls `pytest_socket.disable_socket` itself, so the whole
  integration tier stayed red locally on the 215 tests with only the 2 names the
  script already passed.

## Screenshots

The use task's page, where the due row reads `Counting` rather than a dash.

Use task page showing Counting on its next due row

The same page on a phone

The dashboard card grouped by status, with the Counted section that used to be
missing. The use task was dropped from this card entirely before.

Grouped dashboard card with a Counted section

The same card at phone width

## One-way doors

None. No service field, event payload, storage key or entity attribute changes
name, shape or format. `cycle_start` reads the `skips` list that `skip_task`
already writes, and `_is_occurrence` is private to `recurrence.py`.

One behaviour change worth naming, though it commits no surface: a skip on a
counted wear item's replacement task now restarts its count and its time backstop.
That is what makes Skip usable, and it matches what a skip already means on a
floating task and on a usage meter.

## Test evidence

```
pytest tests/unit           2397 passed, 1 skipped, 1 xfailed
bash ci/test-frontend.sh    45 files, 1447 tests passed
integration (215 tests)     215 passed in 479s
e2e (3 widths)              342 passed in 9.8m
ruff check / format         clean
mypy                        2 errors, both pre-existing on main
vale CHANGELOG.md           no new hits against origin/main
```

## Checklist

- [x] `CHANGELOG.md` updated with 5 bullets under the unreleased `0.24.0b3` section
- [x] Panel UI changed, so screenshots are committed under `docs/images/` and
      embedded above, desktop and phone
- [ ] New user-facing UI surface: not applicable, no new surface
- [ ] New user-facing feature: not applicable, bug fixes only

https://claude.ai/code/session_016qeVatfwqCXWUs6jbpoxhJ

---
_Generated by [Claude Code](https://claude.ai/code/session_016qeVatfwqCXWUs6jbpoxhJ)_


prestomation:
/q review Please be critical and skeptical. These are 5 bug fixes in the unreleased 0.24.0b1-b3 range. Surface the most serious problems first, and do not withhold the minor ones.

Scrutinize in this order:

1. **`reconcile.cycle_start` (new).** It takes the later of `last_completed` and the newest entry in `skips`, and both `uses_since_replacement` and `replacement_backstop_due` now measure from it. Is the semantic right, that a skip restarts a counted wear item's cycle? What breaks if a user deletes a skip from the history dialog, or edits one, or imports a document carrying skips? `trim_use_completions` protects entries above the marker, so a marker that moves forward shrinks the protected window: can that now drop an entry the count still needs? Is there a path where `skips` holds a timestamp in the future?

2. **`recurrence._is_occurrence` (new), used by `_advance_fixed_schedule`.** It decides whether `next_due` is on the schedule's grid by asking for the occurrence after `next_due` minus 1 microsecond. Is that sound for every freq, interval and season combination? Check the DST cases specifically: a fixed schedule anchored inside the repeated hour of an autumn transition, and one anchored in the spring-forward gap. Check `_clamp_season`, which I claim only ever returns a grid occurrence. What happens for a task whose anchor was edited after `next_due` was written, or whose `next_due` came from `set_due_today`? Is the extra `next_fixed_occurrence` call a cost worth worrying about on a large store?

3. **The card fixes in `card-filter.ts`.** `STATUS_ORDER` gained a `counted` row, placed after `shopping` and before `today`. Does that placement agree with the panel's own table, and does any other consumer of `STATUS_ORDER` or `StatusBucket` assume the old membership? The horizon now skips an array of 3 filter names rather than 2 inequality checks: is `counted` the last dateless filter, or is one missed?

4. **`mergePartForm`'s backstop guard.** It falls back through the typed value, then the stored value, then 1. Is falling back to the stored value right when a user deliberately clears the box, or does that write back a number they meant to remove? Is `Number.isInteger` the right test given what `ha-selector-number` emits?

5. **`countedProgress`'s new `!other?.manual` guard.** Does it now agree exactly with `reconcile.counted_part_pairs` and `reconcile_part_tasks`? Is there a third backend lookup it still disagrees with?

6. **Test quality.** The tests are written to fail first, but are they asserting on behaviour or on implementation? Would any of them still pass if I reverted the fix? Is the new integration coverage in `test_use_tasks.py` timing-dependent in a way that will flake on CI?

Also: HA best practices, anything in the diff that could be a performance problem on a store with thousands of completions, and any place where the panel and the backend still disagree about the same number.

---

prestomation:
/q review Second push, after the mutation gate. Same standing ask: be critical and skeptical, most serious first, and do not withhold the minor ones.

`5ac0a3ca` simplified 2 of the helpers the first push added, so please re-read them as code rather than as a diff:

1. **`cycleStart` in `utils.ts`.** It now maps every candidate through `new Date(value ?? '')` and filters NaN, then takes `Math.max`. Is that equivalent to the explicit loop it replaced for every input the panel can hand it? Specifically: `last_completed: null`, a `skips` array holding an entry with no `ts`, a mixed-offset history, and a timestamp far enough out that `Math.max` over a large spread loses precision. Does it still agree with `reconcile.cycle_start` on the Python side for all of those?

2. **`backstopInterval` in `forms.ts`.** I removed the `raw == null || raw === ''` guard on the grounds that `Number` reads null, `''` and `[]` as 0 and undefined as NaN, so every case it caught failed `Number.isInteger(n) && n >= 1` anyway. Is that reasoning airtight for everything `ha-selector-number` can emit? `Number(true)` is 1 and `Number(' 12 ')` is 12: can either reach this?

3. Two mutants survive and I judged both equivalent without annotating them: the `?? ''` string literal in `cycleStart`, and the optional chaining in `!other?.manual` (the preceding `other?.asset_id === src.asset_id` short-circuits whenever `other` is undefined, so the chain is unreachable). Do you agree, and should they carry `// Stryker disable` comments with a reason rather than being left silent?

Local gates on this head: Python mutation 91.16%, frontend mutation 95.83%, unit 2397, frontend 1457, integration 215, e2e 342.

---

prestomation:
`7e796660` cuts `0.24.0b4`. `v0.24.0b3` was tagged while this branch was in flight, so `changelog-release-gap` was right to refuse the entry: `release.yml` keys off an unchanged version and would have skipped re-publishing b3, leaving these fixes on the floor. The 5 bullets moved into their own `## [0.24.0b4]` section and `manifest.json` + `const.py` follow. No `preview-release` label, since this is bug fixes only.

/q review Third push. Review only this commit if the earlier 2 are already covered: it is a version bump plus a CHANGELOG section move.

- Are `manifest.json` and `const.py` (`PANEL_VERSION`) the complete set of version sources, or is there a third that will now disagree?
- Is `0.24.0b4` the right next number given `v0.24.0b3` is tagged, and does it sort correctly under PEP 440 against both `0.24.0b3` and the eventual `0.24.0`?
- The 5 bullets are `### Fixed` under a beta that carries no `### Added`. Does `ci/release-issues.py` handle a Fixed-only section, and does `notify-issues` do anything sensible with a section that references no issue?
- Do the bullets themselves hold to the house rules: bold lead as a 2 to 5 word noun-phrase heading with no articles, the whole bullet inside 3 sentences counting the lead, and nothing narrating the mechanism?

---

PR #335, opened 2026-09-12
fix: counted wear items keep their count, can be renewed early, and name their time limit

prestomation:
## What changed

The 3 counted-wear findings from the #334 review that needed a decision, plus 3 defects
found while fixing them — 2 of which were sitting on `main` behind green checks. No
CHANGELOG entry and no version bump: these are fixes to features that have not reached a
stable release, so they belong to the bullets `0.24.0` already carries. Apply
`preview-release` for an installable build to test against.

### 1. A counted wear item lost its count across an export and import

The count *is* the use task's completion log, and neither derived task is portable, so
an import rebuilt both halves empty. A jacket at 24 of 25 wears came back at 0 of 25,
silently — while a *months*-measured part survived, because `last_replaced` is an
ordinary part field. The feature was inconsistent with the one beside it.

`carried_uses` is a part field mirroring that `last_replaced` precedent: computed at
export from `reconcile.counted_uses`, stored by an import, never maintained during
ordinary use. It **self-expires** — `counted_uses` adds it only while `cycle_start` is
`None`, the state a freshly minted replacement task is in — so the first completion or
skip of the replacement half retires it with no write. That is why `store.py` is
untouched: a reset is a write path, and a write path can go out of step with the log it
tracks.

`uses_since_replacement` keeps its meaning and its callers. `trim_use_completions` still
reads it, deliberately: the carry is not in the log, and routing it through the trim
would inflate retention for entries that do not exist.

**Rejected: making the 2 derived tasks portable.** The better long-term idea and the
wrong change now — it has to un-withhold `source` in the one module whose design is
"name what you exclude", it needs a document-id to planned-id remap for
`source.part.asset_id` that does not exist (`asset_refs` is keyed on `external_id` and
`name` only), it must reopen a door `store.add_task` deliberately closes, and it breaks
silently if a hand-editor deletes either task record — in a document whose selling point
is being hand-editable. Written up in `IDEAS.md` as "portable source-owned tasks", to be
done once for all 4 reconciler namespaces, since its real prize is the renewal *history*
rather than an integer.

Two defects my own tests caught while being written:

- `_asset_out` stamped the count through `_strip`'s **shallow** copy, so the exporter
  wrote its figure into live storage. `_merge_parts` then read it back on a re-import
  and a household's own file doubled its own count.
- A stored carry on a part with no derived pair was dropped. A part mid-reconcile, or
  one in a hand-edited store, has no pair to count from, and losing the carry loses the
  cycle the field exists to keep.

### 2. A counted replacement task could not be completed while dormant

`docs/COUNTED_WEAR_ITEMS_PLAN.md:227` always said an early completion was allowed and
restarts the count, and `store.complete_task` always honoured it. Only the panel withheld
the button, so renewing the jacket's coating at 10 of 25 wears could not be recorded and
the count climbed to 31 of 25.

Two new integration tests prove the backend half: an early Done leaves the task dormant,
moves the marker, stamps the part, and a whole fresh cycle of uses is then needed to arm
it again.

The fix is one clause in `utils.isMonitoredDormant`, which
`.amazonq/rules/architecture-and-code.md` makes the single rule all 3 surfaces read —
spelling it out per surface is what shipped #231. The predicate already carried this
exception for a dormant usage meter, for the same stated reason, so the counted
replacement joins it rather than arriving as a special case.

The discriminator needs no asset lookup. `reconcile_part_tasks` is the only writer of a
non-manual `part` source and gives a time-measured part `floating` instead:

| reconciler task | recurrence | source namespace |
|---|---|---|
| use half | `REC_USE` | `part` (role `use`) |
| **counted replacement** | **`REC_TRIGGERED`** | **`part`, no role** |
| time-measured wear | `REC_FLOATING` | `part` |
| buy reminder | `REC_ONE_OFF` | `buy` |
| problem sensor | `REC_TRIGGERED` | `problem_sensor` |
| declarative companion | `REC_SENSOR` | `declarative_companion` |

The `manual` half is load-bearing: `set_task_consumable` puts that same shape on any task
the user owns, and that one really is waiting on its owner. Both cases are pinned on all
3 surfaces, each with a negative control.

### 3. The part row hid the time backstop

A part set to "every 25 uses, or every 12 months, whichever comes first" named only the
uses half. Once the backstop fired the row read "0 of 10 runs" against an empty meter
while its task sat in Overdue, with nothing to explain why.

`partSummaryLine` already rendered exactly this clause for the part editor, so the fix
extracts it as `partBackstopLabel` and gives the row the same chip. **No new locale key**
— `part.orEvery` is already translated in all 16 locales. The helper lives in `forms.ts`
rather than `panel-detail.ts` because `forms.ts` is on Stryker's mutate list and the other
is not.

A second chip rather than one longer label, agreed from a mockup of both at both widths:
an assist chip's text does not wrap, so a combined label is one unbreakable 33-character
pill in a row with ~250px of usable width at 360px.

### 4. Every exported part failed the schema Home Keeper publishes for it

Found because the **published-schema gate had been red on `main` since 0.24.0b3 while
reporting green**. `lint.yml` ran `pytest … | tee`, so the step took `tee`'s exit status
whatever pytest did, and the `grep -qE "[0-9]+ passed"` guard on the next line matched the
`23 passed` inside `1 failed, 23 passed`. Both safeguards had to fail together, and they
did. `set -o pipefail` closes it.

The defect underneath is older. `_strip` gave the appliance and the task their readable
shape; the parts list went past it untouched, so every part exported every key
`_normalize_part` writes — up to ten of them null. The published JSON Schema types
`stock`, `reorder_at`, `consume_quantity` and `restock_quantity` as `number`, so **an
export of an ordinary wear part fails the schema named on its own first line**, and an
editor flags 4 errors on a file Home Keeper just wrote. #321's counted wear part is what
put a null-bearing part into the maximal fixture and turned the gate red; the export was
already wrong.

The same pass exported `file_name`, `file_content_type` and `file_size` — the 3 keys
`_PART_SCHEMA` **refuses** — so `update_asset` answered 400 for an automation replaying an
exported appliance, and the document named an attached file it does not contain. Those
move to `EXCLUDED_PART_KEYS`, the exporter's third exclusion table, and
`count_file_documents` now counts a part's file so the envelope's `skipped` figure stops
reading low on exactly the files a restore has to re-upload.

`test_no_exported_value_is_null` is the part that matters: the only test that saw this
lives behind `HK_SCHEMA_GATE` and needs `jsonschema` on Python 3.14, so it runs in one job
and nowhere a contributor looks. The new test walks a maximal document in the plain unit
lane and names every null path. Reverting the fix makes it report the 4 paths the schema
gate reported, plus 7 more.

### 5. The walkthrough tour had run out of budget

The check went red on a commit that changes `_normalize_carried_uses` and nothing else.
Measured: desktop **204s** against a 240s cap documented as "the measurement plus ~40%"
for a **168–174s** tour.

No single PR did that. Six feature PRs edited the tour after #298 measured it — search +4
beats, notification icons +3, import/export +14, due today +4, counted wear +7, guide
split +0 — each correctly adding the beats its surface needed under the gate that requires
one, and none moved the number. 32 beats at 900ms is ~29s. The margin fell to ~15% and the
tour began timing out on luck; `retries: 2` buys nothing when every attempt is over the
cap. 240s → 300s, and the rule now sits in `AGENTS.md`: a PR that adds a beat owns the
re-measurement, and past ~360s shorten the tour instead.

It also closed a real hole. A `waitForTimeout` cannot hang, so anything that eats a whole
budget is a call that never returns — and while `actionTimeout` caps click and fill, it
does not cap navigation, which Playwright leaves uncapped too. `page.goto('/home-keeper')`,
which the tour calls several times, was the one call in it that could hang forever.
`navigationTimeout: 30_000` means a stuck navigation fails naming itself in 30s rather
than costing 15 minutes across 3 retries.

### Also fixed

`api.getAssets` returned `res.assets` unguarded. A resolved-but-empty payload is not a
rejection, so the card's `.catch(() => [])` never fired and `_assets` held `undefined`
behind an `Asset[]` type; `_resolvePartLink` then threw on the first task carrying a part
source — every half of every wear item — and the card painted nothing. Found by a test for
fix 2.

### A gate that was missing

A part's fields sit one level down, and every existing gate reads a record's **top level**
only: the round-trip test never touches voluptuous, and both structural tests in
`test_generate_schema.py` compare `record["properties"]`. So a field added to
`_normalize_part` and to `_maximal_asset` but not to `_PART_SCHEMA` travelled and
validated while staying invisible to an editor — and `update_asset` would refuse the parts
array `list_assets` had just returned. The new gate parses `_PART_SCHEMA` and compares it
with what the builder really writes, with an `EXCLUDED_PART_KEYS` table carrying the 3
upload-only file keys and their reasons. Verified it fails, and names the fix, when the
schema entry is removed.

## Screenshots

The part row, now naming both limits.

Appliance part row showing Every 25 uses beside or every 12 months

The same row on a phone, each chip wrapping to its own line

The replacement task's page, with the Done it used to withhold.

Replacement task page offering Done while still counting

The same page at phone width

Its list row. The 2 Battery Notes rows above are the control: monitored-dormant too, and
still no Done.

Task list with Done on the replacement row and none on the battery rows

The same rows at phone width

The dashboard card, where the same row gained its mark-done.

Grouped dashboard card with a complete button on the replacement row

The same card at phone width

The 2 phone shots above carry new filenames. Their old ones came back backtick-wrapped on
3 submissions of this body while their neighbours went through clean, which is the #298
case `AGENTS.md` records: the mangling follows the file, and only a rename clears it.

## One-way doors

Fix 1 commits one. Fixes 2 and 3 commit none — fix 2 is a UI affordance over an
already-permitted `complete_task` call, and fix 3 is internal markup reusing an existing
locale key. Fix 4 **changes what an export writes**, which is listed below as well.

| Door | Shape | Where it appears |
|---|---|---|
| `carried_uses` | integer, `0 <= n <= 500`, default 0, absent means 0 | Storage: `.storage/home_keeper` → `assets[].parts[].carried_uses` |
| | same | Portable document: `appliances[].parts[].carried_uses`, written only when non-zero |
| | same | Service input: `add_asset` / `update_asset` → `parts[]` (`_PART_SCHEMA`), documented in `services.yaml` |
| | same | Published JSON Schema, generated from `_PART_SCHEMA` |
| | same | Read surfaces: the `list_assets` service response, the `home_keeper/list_assets` websocket payload, the panel's `Part` type |
| **Semantics** | "uses counted before this record arrived; added to the use task's log **only while the replacement task has never been completed or skipped**" | The expensive half. An automation could come to rely on this meaning |
| Exported part shape | a part now states only what is true of it: no null-valued key, and never the 3 upload-only file keys | Portable document: `appliances[].parts[]`. A reader that indexed a key by position, or expected `stock: null` on a wear part, sees fewer keys. It was never valid against the published schema, so nothing correct depended on it |

**Ceiling rationale.** 500 (`MAX_COMPLETION_HISTORY`), not `MAX_USE_TARGET` (250): a
household can legitimately let a count run past its target, so the target is the wrong
ceiling — but `recurrence._record_entry` trims every log to 500 on every write, so a carry
above that is a figure this code could never have produced. A negative fraction is refused
too: `int` truncates toward zero, so `-0.5` reached the `>= 0` guard as `0` and passed
while `-1` was refused. `math.floor` for a real number, `int` for the text a quoted YAML
scalar gives.

**`TRANSFER_FORMAT` stays 1.** `const.py:94` bumps it only when the meaning of an existing
key changes, and this is additive. `ci/generate_schema.py:23-29` is the reason not to bump
casually: the deploy runs `clean: true`, so a bump that stops emitting
`home-keeper-1.schema.json` un-publishes every document written this year.

**One behaviour worth naming:** on an **update**-match the carry does *not* travel.
`_merge_parts` restores the stored value the way it restores the upload-only file keys.
That is correct — the local pair is live and authoritative — and it is also the
anti-double-count guard.

## Known non-goal

**Skip is still unreachable on a dormant counted replacement**, even though
`cycle_start` treats a skip as a cycle restart. Offering it means either a fourth
per-task condition in `deferVerbs` needing the same discriminator — putting the rule in
two places, which is the #231 failure again — or a separate "restart the count" action.
Both are product decisions, and Done already restarts the cycle.

## Test evidence

```
pytest tests/unit             2434 passed, 1 skipped, 1 xfailed
bash ci/test-frontend.sh      46 files, 1484 tests passed
integration                   green on the transfer/parts/uses/assets files, fresh container
e2e (3 widths)                342 passed
walkthrough tour              desktop 3.6m, phone 37.2s, both green at the new budget
mutation, Python              89.46%  (threshold 80%)
mutation, TypeScript         100.00%  (threshold 80%)
ruff check / format           clean
mypy                          2 errors, both pre-existing on main
vale                          no new hits against origin/main
```

The published-schema gate now reports **24 passed** in `lint.yml`'s mypy job — the first
run in which that number means anything, because until `set -o pipefail` the step reported
green through `1 failed, 23 passed`.

The full local integration tier is flaky in this container (a different 4–8 tests fail per
run, all in sensor-watcher / documents / companion-sync, none touching `transfer.py`),
while CI's integration job is green on every commit here. The targeted files pass 68/68 on
a fresh container.

## Checklist

- [x] `CHANGELOG.md` deliberately untouched: fixes to features not yet in a stable release
- [x] Panel UI changed, so screenshots are committed under `docs/images/`, desktop and phone
- [ ] New user-facing UI surface: not applicable, no new surface
- [x] `preview-release` applied instead of a beta bump, for testing before the stable cut

https://claude.ai/code/session_016qeVatfwqCXWUs6jbpoxhJ

---
_Generated by [Claude Code](https://claude.ai/code/session_016qeVatfwqCXWUs6jbpoxhJ)_


prestomation:
/q review Please be critical and skeptical. Surface the most serious problems first, and do not withhold the minor ones. Three fixes plus a new test gate; the first is the one that commits a one-way door, so weight it accordingly.

**1. `carried_uses` semantics (the expensive part).** It is computed at export, stored by an import, and never maintained during normal use. It self-expires: `reconcile.counted_uses` adds it only while `cycle_start(replace_task)` is None.
- Is the self-expiry airtight? Can a replacement task reach a state where `cycle_start` is None again *after* a cycle has started — a deleted completion through the history dialog, a deleted skip, `remove_completion`, `move_completion`, an `update_task` that clears `last_completed`? If so the carry silently revives and inflates the count. That is the failure mode I am least able to rule out.
- `_merge_parts` restores the stored value unconditionally, so on an update-match the carry does not travel. Is that right for an import onto a *different* install that matches by `external_id` or name, where the parts are the same but the tasks are that install's own?
- Is 500 the right ceiling, and is refusing an over-ceiling value better than clamping? An import of a hand-edited file now fails the whole appliance on one bad integer.
- `_counted_uses_by_part` falls back to the part's stored carry when `counted_part_pairs` finds no pair. Can that fall back to a *stale* carry that has in fact expired — a pair broken by a deleted task, where the replacement half really was completed?

**2. The export mutation I fixed.** `_asset_out` now deep-copies each part before stamping. Are there other places in `transfer.py` where `_strip`'s shallow copy lets the exporter write into live storage? `documents` is copied per-dict; `metadata` and `related_device_ids` are not obviously.

**3. `isMonitoredDormant`'s new clause.** It returns false for a `triggered` task with a non-manual `source.part`. I claim `reconcile_part_tasks` is the only writer of one and gives a time-measured part `floating` instead. Is there any other path — `store.update_task`, the importer, a websocket command, a migration from an older storage format, a hand-edited `.storage/home_keeper` — that produces a dormant `triggered` task with a non-manual `part` source that is *not* a counted replacement? And does offering Done on such a task do any harm if one exists?

**4. `partBackstopLabel`.** It reads `partCountsUses` rather than `replace_also_every` directly. Does that match every place the backend clears the field, and is there a window where the panel and the store disagree about whether a part still has a backstop?

**5. The new part-schema gate** in `test_transfer_coverage.py`. It parses `_PART_SCHEMA` with `ast`. Is the parse robust to the ways that schema could be written later — a key built from a constant, a nested `vol.Schema`, a spread from another dict? A gate that silently stops seeing keys is worse than no gate.

**6. Test quality.** Would any of the new tests still pass if I reverted the fix they cover? Is the new integration coverage timing-dependent in a way that will flake on CI? And is `counted-wear-actions.test.js` asserting behaviour rather than markup?

Also: HA best practices, and anything in the diff that is a performance problem on a store with thousands of completions — `_counted_uses_by_part` runs `counted_part_pairs` on every export.

---

prestomation:
Answering my own question 1 from the review request, since it was the thing I could not settle by reading and `/q` did not reach it.

**Yes, the carry can revive** — and it should. `counted_uses` adds it only while `cycle_start` is `None`, and deleting a replacement task's only completion through the history dialog puts `last_completed` back to `None`. So the carry returns.

Traced against the real modules:

```
imported, untouched     : 24
after completing here   : 0
after 30 more uses      : 30
cycle_start after delete: None
count after delete      : 54
```

54 is the honest answer. Deleting the completion says the replacement never happened, so the cycle it started never happened either, and the count is what the part arrived with plus everything recorded since. The alternative — a carry that expires on first contact and never comes back — would silently lose 24 wears the moment somebody corrected a mistaken Done.

`f2476b67` pins it both ways round, since a skip starts a cycle too and undoing one has to undo that cycle symmetrically.

Still open from that list and worth a reviewer's eye: whether `_counted_uses_by_part`'s fallback can reach a **stale** carry (a pair broken by a deleted task, where the replacement really was completed), and whether the `ast` parse of `_PART_SCHEMA` in the new gate is robust to how that schema might be written later — a gate that silently stops seeing keys is worse than no gate.

---

prestomation:
Correction on the self-expiry answer, so a later reader does not trust the wrong mechanism.

The review reply said `cycle_start` can only return to `None` if **all** completion and skip entries are deleted, and that no path clears a list to empty. That reasoning is wrong. Deleting a replacement task's only completion does return `cycle_start` to `None`, and I traced it live:

```
imported, untouched     : 24
after completing here   : 0
after 30 more uses      : 30
cycle_start after delete: None
count after delete      : 54
```

The **outcome** is right, and it is why the carry is safe: deleting the completion means the replacement never happened, so 24 carried plus 30 new is the honest count. The carry is not a stale figure that leaks back; it revives exactly when the event that retired it is undone.

`f2476b67` pins both directions — the completion path and the skip path — so the behaviour cannot change silently.

`c32f94ec` also closes the coverage gap the report named at `assets.py:546-551`: every path `_normalize_carried_uses` refuses (text, negative, above `MAX_COMPLETION_HISTORY`, and a boolean, which is an `int` in Python so only the boolean guard catches it), plus the 500 ceiling itself.

---

prestomation:
/q review Please be critical and skeptical about `c32f94ec`, which adds unit tests for `_normalize_carried_uses` in `custom_components/home_keeper/assets.py`. Surface the most serious issues first, and do not withhold minor ones.

Specifically:

1. **Correctness of the boundary claims.** The ceiling is `MAX_COMPLETION_HISTORY` (500), not `MAX_USE_TARGET` (250), on the argument that a count legitimately runs past its target while `recurrence._record_entry` trims every log to 500. Is 500 the right refusal point, or does some path let a legitimate figure exceed it?

2. **Test quality, not test count.** Does each new test assert something that would fail if the line it covers were wrong, or does any of them only execute the line? The boolean case in particular — is matching on the message enough, or should it assert which guard rejected it?

3. **Missing cases.** Is there an input shape `_normalize_carried_uses` accepts that no test states: a float, a numeric string, a very large int, `0` as a distinct case from absent?

4. **Interaction with `_merge_parts`.** The design turns on `_merge_parts` restoring the stored `carried_uses` unconditionally on an update-match. Is any path able to reach `_normalize_part` with an imported carry and have it survive onto a live part that already has a pair?

---

prestomation:
`925c22cd` fixes something the review turned up that is worse than what this PR set out to do: **the published-schema gate has been red on `main` since 0.24.0b3, under a green mypy check.**

`lint.yml` ran `python -m pytest … | tee /tmp/schema.log`. A pipeline's exit status is the last command's, so the step took `tee`'s 0 whatever pytest did. The guard on the next line, `grep -qE "[0-9]+ passed"`, then matched the `23 passed` inside `1 failed, 23 passed`. Both halves had to fail together, and they did. Same failure, same shape, on `1536444b` (main) and on this branch:

```
appliances/0/parts/1/consume_quantity: None is not of type 'number'
appliances/0/parts/1/reorder_at: None is not of type 'number'
appliances/0/parts/1/restock_quantity: None is not of type 'number'
appliances/0/parts/1/stock: None is not of type 'number'
```

The defect behind it is older than the gate's blindness. `_strip` gave the appliance and the task their readable shape, and the parts list went past it untouched, so every exported part carried every key `_normalize_part` writes — up to ten of them null. **An export of an ordinary wear part fails the JSON Schema Home Keeper publishes for it and names on the file's first line**, so an editor flags 4 errors on a file Home Keeper just wrote. #321's counted wear part is what put a null-bearing part into the maximal fixture and turned the gate red; the export was already wrong.

The same pass also exported `file_name`, `file_content_type` and `file_size` — the 3 keys `_PART_SCHEMA` **refuses** — so `home_keeper.update_asset` answered 400 for an automation replaying an exported appliance, and the document named an attached file it does not contain.

What changed:

- `EXCLUDED_PART_KEYS` in `transfer.py`, the exporter's third exclusion table, holding those 3 keys with reasons. `test_transfer_coverage.py` now reads that table instead of keeping its own copy, so the "what the document leaves out" and "what `_PART_SCHEMA` refuses" halves cannot drift.
- Each part exports through `_strip`, which also supplies the copy the counted-uses stamping needed.
- `count_file_documents` counts a part's attached file, so the envelope's `skipped` figure stops reading low on exactly the files a restore has to re-upload.
- `set -o pipefail` on the schema step.
- `test_no_exported_value_is_null` walks a maximal document and names every null path. **This is the part that matters**: the only test that saw this lived behind `HK_SCHEMA_GATE` and needs `jsonschema` on Python 3.14, so it runs in one job and nowhere a contributor looks. Reverting the fix makes the new plain-lane test report the 4 paths above plus 7 more.
- `ci/test-python-integration.sh` passes arguments through, so the 9-minute tier can be run one file at a time.

Ran locally: 2427 unit, mypy (the 2 pre-existing errors only), ruff clean, Python mutation 88.96%, and the transfer/parts/uses/assets integration files green on a fresh container. The full integration tier is running now and I will report it.

Note the one thing I could not verify here: the schema gate itself needs Python at Home Assistant's floor and this container is 3.12. Reading `lint.yml`'s mypy job on this push is what confirms it — and with `pipefail` in place, a green step now means something.

---

prestomation:
/q review Please be critical and skeptical about `925c22cd`, which changes how a part is exported in `custom_components/home_keeper/transfer.py`. Most serious issues first, and do not withhold minor ones.

1. **Correctness of stripping a part.** Each exported part now goes through `_strip`, which drops `None`, `""`, `[]` and `{}`. Is any of those a *stated* value on a part rather than an absent one — that is, does `_normalize_part` or `_merge_parts` treat an absent key differently from an empty one anywhere? `stock: 0` and `create_buy_task: False` must survive, and I have a test for both; is there a third case I have missed, such as `replace_unit: ""` versus absent, or `stock_unit`?

2. **The envelope figure.** `count_file_documents` now adds a part's `file_name` to the appliance's file-kind documents. Can one file be counted twice, or a part hold `file_name` with no real file behind it?

3. **`EXCLUDED_PART_KEYS` as a shared table.** `tests/unit/test_transfer_coverage.py` now reads the exporter's table for a second purpose — which keys `_PART_SCHEMA` refuses. Are those genuinely the same set for the same reason, or have I fused two ideas that will diverge the moment a part gains a field that travels but the schema does not name?

4. **Round-trip and idempotence.** Does dropping the empties change what a re-import of a household's own export does, in particular `_merge_parts` on an update-match and `test_re_running_the_same_import_changes_nothing`?

5. **The CI change.** `set -o pipefail` on the schema step in `lint.yml`. Are there other steps in this repository's workflows whose exit status is swallowed the same way, by a pipe or by a trailing command that always succeeds?

---

prestomation:
Triage of both reviews. `3ea3d80a` takes the three that were right.

**Taken — the missing input shapes.** `"24"`, `24.5` and a stated `0` all reach `_normalize_carried_uses` and nothing said what happens. Probed rather than assumed:

```
      24.5 -> 24
      '24' -> 24
         0 -> 0
    24.999 -> 24
```

Each is the right answer and each is now pinned. `"24"` is what YAML gives for a quoted number and is what `_PART_SCHEMA`'s `Coerce(int)` already implies. Truncation matches `replace_also_every.interval`, so a count never lands between 2 uses. A stated `0` is a different input from an absent key — it goes down the int path, not the `in (None, "")` guard — so it earns its own case. A hand-written or generated file is exactly what this field exists for, so these are decisions the tests should hold rather than ones `int()` makes quietly.

**Not taken, with reasons.**

- *999999 as a separate case.* It is refused by the same `> MAX_COMPLETION_HISTORY` comparison that 501 exercises, and 501 is the boundary. `MAX_INTERVAL` does not apply to this field. A second test of one branch adds a name, not a guarantee.
- *Asserting the exact message on the boolean case.* The review reasoned it out correctly: with the `_reject_boolean` call removed, `int(True)` succeeds and returns 1, so the test goes red. That is what makes it a real assertion, and matching on the message would pin the wording rather than the behaviour.
- *The concurrent import-during-reconcile hazard.* The scenario needs an import's part creation threaded through an update path, and the review's own walk of the code found that path does not exist. `_merge_parts` restoring unconditionally is already covered by a test where the incoming part states a different value. I would rather not add a test for a sequence nothing can produce — it would pass forever and say nothing.

On the `925c22cd` review: no findings, and its conclusion on question 5 matches what I found here — `grep -rn "| tee" .github/workflows/ ci/` returns that one step and nothing else.

Coverage is 100% of new statements on both pushes (`transfer.py` 39/39, `assets.py` 22/22, `reconcile.py` 14/14).

---

prestomation:
/q review One more push, `3ea3d80a`, which only adds tests to `tests/unit/test_assets_part_uses.py` for the input shapes your last review named. Be critical and skeptical, most serious first, and do not withhold minor ones.

1. **Is the parametrized truncation case stating the right rule?** `24.5` and `24.999` both give 24. I justified that as matching `replace_also_every.interval`. Check that claim against the code rather than taking it from me, and say whether any other whole-number part field disagrees.

2. **Negative fractions.** `-0.5` truncates toward zero to `0`, which the `< 0` guard then accepts rather than refuses. Is that a real hole in `_normalize_carried_uses`, and does it matter given where the value comes from?

3. **Did I reject anything I should have taken?** I declined a separate 999999 case, an exact-message assertion on the boolean guard, and a test for an import running during reconciliation. Push back if any of those is a real gap.

---

prestomation:
Answering my own question 2 before the review gets to it: the negative-fraction hole is real, and `183e3e2d` closes it.

```
      -0.5 -> ACCEPTED as 0
    -0.999 -> ACCEPTED as 0
        -1 -> refused: carried_uses must be >= 0
      -1.5 -> refused: carried_uses must be >= 0
    '-0.5' -> refused: carried_uses must be an integer
```

`int` truncates toward zero, so a fraction below zero reached the `< 0` guard as `0` and passed. One wrong file, answered 2 ways — and the half that passed said nothing, which is the opposite of what the refusal is for. `replace_also_every.interval` escapes this only by accident: its floor is 1, not 0, so `int(-0.5)` fails that check anyway.

Fixed with `math.floor` for a real number and `int` for the text a quoted YAML scalar gives. No legitimate input changes — `24.5` still gives 24 and `"24"` still gives 24 — and the negative cases are parametrized over both an integer and a fraction, so the `isinstance` branch has a killing assertion on each side.

Python mutation gate 89.46% on the changed surface (threshold 80). 2434 unit tests, ruff clean, mypy showing the 2 pre-existing errors only.

The earlier `/q review` request still stands for `3ea3d80a`; question 2 there is now answered by this commit rather than open.

---

prestomation:
The walkthrough check went red on `183e3e2d`, a commit that changes `_normalize_carried_uses` validation and nothing else — no panel code, no tour edit. `9076f387` fixes it, and the cause is worth writing down because the red check was pointing at the wrong thing.

Reproduced locally on a fresh container, then measured with the cap raised so the run reports its own length instead of dying at it:

```
desktop  3.4m (204s)    cap 240s
phone    36.7s
```

`walkthrough.config.ts` sets its budget as "the measurement plus ~40%", against a tour measured at **168–174s**. #321 added the counted wear item beats — which that gate required it to — and left the number alone. The margin fell from ~40% to ~15%, and the tour has been passing on luck ever since: green at 08:11 on `f2476b67`, red at 13:59 on `183e3e2d`, with nothing between them touching the panel. `retries: 2` buys nothing when every attempt is over the cap, which is why all three failed.

240s → 300s, which is 204s plus ~45%.

Two things rule out a hung tour rather than a long one. `actionTimeout: 20_000` means a bad selector fails in 20s naming its step — it did not fire, so nothing was stuck on an element. And with the cap raised the tour passes end to end, twice, at the same length.

The rule this needed now exists in `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md`: the desktop walk is a fixed sequence of pauses, so its wall clock only ever grows, every PR that adds a beat owns the re-measurement, and a timeout here means suspect the margin before suspecting CI. Measure with `--timeout=600000 --reporter=list` and read the duration reported, never the cap it died at.

Also verified locally on this branch while the container was up: **342 e2e tests pass** across all three widths.

---

prestomation:
/q review `9076f387` raises the walkthrough tour's per-test timeout from 240s to 300s in `tests/e2e/walkthrough.config.ts` and records the reasoning in `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md`. Be critical and skeptical — most serious first, and do not withhold minor ones. I am specifically worried that I have papered over a real slowdown.

1. **Is raising the cap the right call, or am I hiding a regression?** I measured 204s against a documented 168–174s and attributed the ~35s to the counted wear item beats #321 added. Check that attribution against the diff of `walkthrough.capture.ts` in `e3be2a6b`: do the beats it added plausibly account for ~35s, or is something else in this PR's range making the panel slower?

2. **Is 300s the right number, or should the tour be trimmed instead?** The tour only ever grows. At what point does raising the cap stop being the answer, and is there a beat in the desktop walk that is redundant with the phone walk?

3. **Does the guard still guard?** The cap is the only thing bounding a hung tour, since `actionTimeout` bounds actions and the tour is mostly `page.waitForTimeout`. At 300s with `retries: 2`, a genuinely hung tour now costs 15 minutes of CI. Is there a better shape — a per-section assertion, a wall-clock check inside the tour — that would fail fast on a hang while still allowing a long healthy run?

4. **Anything else in the repository with the same shape:** a timeout or threshold documented as "measurement plus N%" that a later change has quietly eaten into.

---

prestomation:
The review was right to make me check the attribution, and **I had it wrong**. `7b0d3eb1` corrects it.

**#321 added 7 beats, not ~38.** That is 6.3s, not the ~35s I attributed to it. The review said that if the count came in well under ~30 beats then something else was contributing — so I counted every tour edit since #298 took the 168–174s measurement:

```
#303 search             +4      #318 due today      +4
#302 notification icons +3      #321 counted wear   +7
#309 import/export     +14      #333 guide split    +0
```

32 beats at 900ms is ~29s, and the interactions those PRs added alongside their beats cover the rest of the ~33s. So the attribution holds, but not to one PR: six of them each added the beats their surface needed, correctly, under the gate that requires a new surface to appear in the tour, and none moved the budget. It is nobody's regression and everybody's — a different lesson from the one I wrote down, and the one the rules now carry.

**On the margin math, I disagree.** The comment reads 204s + 45% ≈ 296s as leaving "~4 seconds of actual margin". That compares the computed number to the cap. The margin is cap minus tour: 300 − 204 = **96s, about 47%**. This run, with the navigation cap added, measured 216s — still 84s of headroom, and a useful data point that 240s was genuinely marginal rather than unlucky.

**The hang guard was the good catch, and there is a sharper fix than per-section timeouts.** A `waitForTimeout` cannot hang — it is a fixed duration — so anything that eats a whole budget is a call that never returns. `actionTimeout: 20_000` already caps click and fill. What it does not cap is **navigation**, which Playwright also leaves off by default, and `page.goto('/home-keeper')` — which the tour calls several times — was the one call in it that could hang forever. That is the hole the per-test budget was quietly covering.

`navigationTimeout: 30_000` closes it: a stuck navigation now fails naming itself in 30s, rather than costing 15 minutes across 3 retries and reporting as a timeout on the whole tour. That gets the fail-fast property without wrapping a 1,000-line tour in `test.step` blocks whose own timeouts would need re-measuring every time a beat moves — which is the maintenance trap this whole thread is about.

Also taken: a documented ceiling. Past ~360s, shorten the tour rather than raise the cap again — drop a beat the phone walk already covers, or split the desktop walk.

On question 4, I checked the other capture configs rather than grepping for the phrase. `screenshots.config.ts` is also at 240s and its comment says it has been raised twice for the same reason ("the capture keeps growing"), so it has the same shape — but it carries no measurement to have gone stale against, so there is nothing to correct, only to measure next time it goes red. `screenshots-shopping-status` (180s) and `screenshots-companion-filter` (120s) are single-surface captures and are not growing.

Verified locally with the cap in place: desktop 3.6m, phone 37.2s, both green.

---

prestomation:
/q review `7b0d3eb1` corrects the tour budget attribution and adds `navigationTimeout: 30_000` to `tests/e2e/walkthrough.config.ts`. Be critical and skeptical, most serious first, and do not withhold minor ones.

1. **Is `navigationTimeout` actually the right guard, and is 30s right?** My claim is that `waitForTimeout` cannot hang, `actionTimeout` covers click and fill, and `page.goto` was the only remaining call in the tour that could hang forever. Check that against `walkthrough.capture.ts`: is there a call in the tour that neither cap covers — a `waitForLoadState`, a `waitForFunction`, a `waitForResponse`, an `expect.poll`, a `browser.newContext`, or the video flush at the end?

2. **Does `navigationTimeout` risk a new flake?** The seeded HA container is slow on a cold start and the tour navigates several times. Is 30s comfortably above the worst realistic panel load on a loaded CI runner, or have I traded a rare hang for a recurring flake?

3. **Check my beat arithmetic.** I counted 32 beats added across six PRs since #298 by grepping added `waitForTimeout` lines in each commit's diff of `walkthrough.capture.ts`. That method misses a beat moved rather than added, and counts one inside a helper only once regardless of how often the helper runs. Does the tour have such a helper, and would that change the total?

4. **The margin disagreement.** Your last review read 204s + 45% ≈ 296s as "~4 seconds of actual margin". I read the margin as cap minus tour: 300 − 204 = 96s. Say plainly which is right, because the rule written into AGENTS.md depends on it.

---

prestomation:
/q review `9e13fc22` is a rename only: 2 phone screenshots and the 2 lines in `tests/e2e/screenshots-counted.capture.ts` that write them. Short review, but be skeptical rather than agreeable.

1. **Did I miss a reference?** I grepped `.ts` and `.md` for the old names and found only the capture script. Check for any other reader — `website/scripts/`, a test asserting on a filename, a docs page embedding it by relative path.

2. **Do the new names fit the convention?** `66c-panel-mobile-renew-task-page.png` and `67c-panel-mobile-renew-task-row.png`. `AGENTS.md` wants the pair to sort together with a `-mobile-` segment on the phone one, but the desktop siblings are still `66-panel-counted-replacement-page.png` and `67-panel-counted-replacement-row.png`, so the pair no longer shares a stem. Is that worse than the mangling it fixes, and should the desktop two be renamed to match?

3. **Anything else in `docs/images/` at risk?** The mangling follows the file and I have no theory for which names trigger it. Is there a pattern you can see across the 4 that mangled and the many that did not?

---

prestomation:
Merged `main` (`c5d9263b`) in — `f99959d8`. One conflict, in `tests/e2e/walkthrough.config.ts`, because #337 diagnosed the same worn-out tour budget independently.

**Main's number wins.** Mine was 300s from one local measurement of 204s. Main's is 360s from three: 204s green on CI, **234s on CI on the very next push of the same branch**, and 210s locally — with the 234s run then failing inside a 40s wait and its first retry hitting 240s outright. My 300s would have left 28% over that 234s sample, which is the margin that had just failed. Same diagnosis from both ends; main simply had the better evidence, and its reading is right that the spread is the runner rather than the tour.

**Two things from this branch survive the merge**, because main has neither:

- `navigationTimeout: 30_000` merged cleanly and makes main's sentence *"this cap is the only thing that bounds a hung tour"* no longer true. `page.goto` was the one call in the tour that could hang forever; it now fails naming itself in 30s instead of eating 18 minutes over 3 attempts. I rewrote that paragraph to say so rather than dropping it.
- The per-PR beat attribution — 32 beats over 6 PRs since #298 — which is the evidence for *why* the tour lengthened, not just that it did.

Main's *"the job's own 30-minute cap gives next"* and my *"past ~360s, shorten the tour"* were the same conclusion reached from opposite directions, so they are now one sentence: **360s is the last free raise**, and the next tour that outgrows it is paid for by shortening the walk. `AGENTS.md` and `.amazonq/rules/` follow.

Re-verified on the merged tree: 2460 unit tests pass (main added 26), ruff clean, mypy at its 2 pre-existing errors, and the merged config lists its 2 tours.

One thing worth flagging from the other side of the merge: #337 touches `sensor_watcher.py` and adds 92 lines to `tests/integration/test_sensor_watcher.py`. That is the file whose local failures I reported here as container flake. Some of those may have been the real defects #337 fixed rather than noise, so treat my "flaky locally" note on this PR as less certain than I wrote it.

---

prestomation:
/q review `f99959d8` merges `main` into this branch and resolves one conflict in `tests/e2e/walkthrough.config.ts`. Be critical and skeptical, most serious first, and do not withhold minor ones.

1. **Did I resolve the conflict correctly, or lose something?** I took main's `timeout: 360_000` and its measurement paragraph, kept this branch's `navigationTimeout: 30_000` and beat-attribution paragraph, and rewrote main's "this cap is the only thing that bounds a hung tour" because `navigationTimeout` makes it false. Read the merged file against both parents and say whether any claim in it is now wrong, duplicated, or contradicts a neighbour.

2. **Is 360s still right given `navigationTimeout`?** With a 30s navigation cap in place, a hang now fails early, so the per-test budget only ever has to cover a healthy-but-slow run. Does that argue for a *lower* cap than 360s, or is the 234s CI sample still the number that decides it?

3. **Semantic conflicts the merge did not flag.** `.amazonq/rules/architecture-and-code.md` auto-merged, and both sides edited it. Check that my "a record inside a record gets the same treatment" addition and whatever #337 added still read as one coherent set of rules rather than two.

4. **Did #337 change anything my export fix now contradicts?** It touches `store.py`, `models.py` and `sensor_tasks.py`. My change alters `transfer.py`'s exported part shape and `assets.py` validation. Check for any interaction — a new field on a part or an asset that now fails to travel, or a sensor task that the null-stripping affects.

---

prestomation:
The review asked for more context rather than answering, so here are the 2 merge questions I could settle myself, with the evidence.

**Do the merged rules read as one set?** Yes. #337's additions to `.amazonq/rules/architecture-and-code.md` land in the sensor/reconciler section (holds, carried edge state, recipe pausing, the reconcile debouncer); mine is in Data portability, ~300 lines further down. Different sections, no overlapping claim, nothing to reconcile.

**Did #337 add anything my export change now breaks?** No, and the reason is worth recording. Its `models.py` change adds `EDGE_ONLY_SENSOR_FIELDS`, which is a validation tuple that *refuses* `for_seconds` and `clear_on_recover` on a usage-mode sensor task. Both fields already existed, so nothing new has to travel, and the round-trip gate passes. The one interaction I checked closely: it refuses only a **truthy** value, so an exported `for_seconds: 0` still imports — which matters because my null-stripping does not reach into a task's nested `sensor` block.

**That last point turned up something worth a follow-up.** My new `test_no_exported_value_is_null` walks the whole document, but the maximal fixture's task is `floating` with no `sensor` block at all, so the gate never exercises one. I probed all 4 modes directly to see whether the parts defect exists there too:

```
        state: nulls=[]  empties=[]  keys=[attribute, clear_on_recover, entity_id, for_seconds, mode, state]
    threshold: nulls=[]  empties=[]  keys=[comparison, entity_id, mode, value]
        usage: nulls=[]  empties=[]  keys=[entity_id, mode, target, unit]
 availability: nulls=[]  empties=[]  keys=[clear_on_recover, entity_id, mode]
```

Clean in every mode, and structurally so: `normalize_sensor` builds its result key by key, where `_normalize_part` writes a fixed shape with `None` in the slots that do not apply. That is the whole difference between the two, and it is why parts had the defect and sensor blocks never did.

So there is no bug here — but the **gate has no teeth on that block**, and a future change that made `normalize_sensor` write a null would ship the same way parts did. Giving the maximal fixture a sensor task would close it. I am deliberately **not** doing that in this PR: `_maximal_task` is shared by the round-trip and coverage gates, #337 has just changed sensor validation, and this PR is already ready for review. Better as its own small change than bolted onto this one.

---

prestomation:
/q review Re-asking with the context the last reply wanted. The change under review is commit `f99959d8` on branch `fix/counted-wear-count-portability-and-early-done`: a merge of `main` (`c5d9263b`) into this PR, resolving one conflict. The only hand-edited files are `tests/e2e/walkthrough.config.ts`, `AGENTS.md` and `.amazonq/rules/testing-and-workflow.md`. Be critical and skeptical; most serious first, and do not withhold minor ones.

**The conflict and how I resolved it.** Both sides changed the walkthrough tour's per-test `timeout` after diagnosing the same problem. Main set `360_000` with 3 measurements (204s CI green, 234s CI on the next push of the same branch, 210s local, and a documented failure at 240s). This branch had set `300_000` from a single 204s local measurement. I took main's 360s and main's measurement paragraph, and kept 2 things only this branch has: `navigationTimeout: 30_000` in the `use` block, and a paragraph attributing the tour's growth to 32 beats across 6 PRs since #298.

I also rewrote one sentence of main's text. Main says *"This cap is the only thing that bounds a hung tour."* With `navigationTimeout` set that is false — `page.goto` was the only call in the tour that could hang indefinitely — so the merged file says the cap is the *last* thing that bounds it and that each hanging call now has its own shorter cap.

Please check, reading `tests/e2e/walkthrough.config.ts` on this branch against both parents:

1. **Is any claim in the merged comment block now wrong, duplicated, or contradicting a neighbour?** It is a long comment assembled from two authors and that is exactly where a stale sentence hides.

2. **Is 360s still the right number now that `navigationTimeout: 30_000` exists?** A hang fails early, so the per-test budget only has to cover a healthy-but-slow run. Does that argue for a lower cap, or does the 234s CI sample still decide it?

3. **Is `navigationTimeout: 30_000` safe for the tour specifically?** It applies to every `page.goto` in `walkthrough.capture.ts` against a freshly seeded HA container.

4. **AGENTS.md and `.amazonq/rules/testing-and-workflow.md`**: I merged main's "the job's own 30-minute cap gives next" with my "past ~360s, shorten the tour" into one rule — 360s is the last free raise, and the next tour that outgrows it is paid for by shortening the walk. Does that rule now contradict anything else in either file?

---

PR #337, opened 2026-09-12
fix: sensor holds, and the six defects a review of the feature turned up

### LINKED ISSUE #336: Declarative companions: hold time setting is not respected
I configured a declarative companion for the Device Pulse integration, and I kept the default 3600s hold time. Yet, whenever the monitored entity goes to the configured state, even for a minute, the task is opened and then resolved shortly after. 

This is my config:
image

prestomation:
Fixes #336

### 1. A hold no longer banks time the entity spent silent (#336)

`for_seconds` is meant to measure how long the condition has been **continuously**
true. An indeterminate reading — a missing, `unavailable` or `unknown` entity — kept
the pending crossing alive, so a device that dropped off the network banked the whole
blackout and armed its task the instant it came back in the trigger state. The
reporter kept the Device Pulse recipe's 3600 s hold and got a task for every
one-minute visit to the trigger condition.

Reproduced against the real container before the fix (a 20 s hold):

```
crossed  meter='100.0'       next_due=None                  <- crossing, hold pending
gap      meter='unavailable' next_due=None                  <- 28 s with no reading
back     meter='100.0'       next_due=2026-09-12T09:18:17   <- armed on the return
```

`sensor_tasks._evaluate_indeterminate` is now the one place the rule lives, and the
`threshold`, `state` and `availability` evaluators all delegate to it: decide nothing,
and end a pending hold so the next true reading starts a fresh one. A task with **no**
pending crossing keeps its carried `condition_met`, which is what still stops a dropout
from re-arming a task the user dealt with, or completing a `clear_on_recover` one.
`evaluate_threshold` takes `reading=None` instead of the watcher skipping the task —
the skip left the crossing *and* its hold timer in place.

### 2. Six more defects, from reviewing the rest of the feature

Each one has a test written first.

- **A recipe that meters a sensor lost its meter anchor.** The reconcile pass rewrites
  a materialized task's whole `sensor` block from the recipe, and `sensor.baseline` is
  the one key on it no recipe writes — the watcher stamps it from the first live
  reading. A pass runs on *any* entity, device or area registry event, so a `usage`
  recipe restarted its meter whenever anything in Home Assistant changed and could
  never reach its target. `merge_sensor_binding` carries a stored anchor forward; a
  recipe's own `baseline` now only seeds a task that has none.
- **Disabling a recipe deleted its tasks**, and the completions recorded on them went
  too. `pause_spec_tasks` switches them off and marks the provenance `paused`; the
  enabled pass clears the marker with a `"resumed"` op, which the store reports as a
  freshly-made id so the watcher arms a condition that became true while the recipe was
  off. A task switched off by hand carries no marker, so that choice survives.
- **Carried edge state outlived the condition it was decided against.** A task moved
  from "below 20%" to "below 50%" kept a `condition_met` measured against the old limit,
  so a battery already at 30% stayed dormant until it rose above 50% and fell back
  through it. The edge is stamped with `sensor_tasks.condition_fingerprint` and retired
  when the binding's condition changes — not for a hold or `clear_on_recover` change,
  which decide nothing about whether the condition is true.
- **Nothing debounced the reconciler.** Home Assistant fires one entity-registry event
  per entity and each one ran a full pass: every spec over every entity, a Jinja render
  per match, a store write. The docstring claimed the dispatcher already coalesced them,
  and no `Debouncer` existed anywhere in the component. One does now (`immediate=True`,
  so a saved recipe still shows its tasks at once), shut down with the listeners.
- **`task_template.category` and `task_template.priority` did nothing.** Validated,
  stored, exported and advertised by the service, and a task has no such field. Dropped
  from the normalized shape and from `services.yaml` — but not refused, so a spec that
  carries them still loads instead of being thrown away.
- **A usage binding silently dropped `for_seconds` and `clear_on_recover`**, which is
  the thing `_reject_fields` exists to stop. Refused now, and only when truthy: a form
  that always sends `for_seconds: 0` is saying nothing.

### 3. Two CI budgets that had run out

Neither is a test failure. Both are caps that a growing suite had caught up with, and
both bit this PR.

- **`integration.yml` cancelled a run whose 216 tests had all passed.** The tier takes
  ~9.5 minutes at main HEAD (9:26–9:54 on the last 4 runs) against a 10-minute job cap,
  so any PR adding a test was one cancelled run away. Raised to 20, and the #336 test
  waits a little less.
- **The walkthrough tour outgrew its 240 s test budget.** The desktop tour took 3.4m on
  CI on this branch's previous push and 3.9m on the next one, whose first retry hit the
  cap outright — the same tour, and `tests/e2e/` and the panel are identical to main on
  this branch, so the spread is the runner. The config's own note says to suspect the
  margin first and to move the number against a fresh measurement: 204 s and 234 s on
  CI, 210 s in the dev container, so 360 s is the highest of those plus ~54%. The tour
  passes locally, desktop and phone, and the sticky comment on this PR now renders both.

## Tests

- `tests/unit/test_sensor_tasks.py` — the hold restarts after a gap in all 3 edge
  modes; a missing threshold reading never arms and never clears an armed
  `clear_on_recover` task; the fingerprint's own value, and what does and does not
  change it. Two tests that pinned the old carry-through-a-dropout behaviour are
  rewritten.
- `tests/unit/test_declarative_companions.py` — the meter anchor survives a pass,
  is dropped on a mode change, and a recipe's seed loses to a stored one; pause and
  resume, including a task switched off by hand and one that never stated `enabled`.
- `tests/unit/test_models_sensor.py` — a usage binding refuses a hold and an auto-clear
  switch, accepts an empty one, and reads an empty box as an absent one.
- `tests/integration/test_sensor_watcher.py` — the reporter's sequence against the real
  container (failed on `main` with `armed on the return: the blackout was banked as
  hold time`), and editing a condition to one the meter already meets arms the task.
- `tests/integration/test_declarative_companion_sync.py` — disabling a recipe keeps its
  task and the completion recorded on it, and enabling brings it back.

Run locally: `tests/unit` (2424 passed), `tests/integration/test_sensor_watcher.py` +
`test_declarative_companion_sync.py` (38 passed), `ci/test-frontend.sh` (1457 passed),
the walkthrough capture (desktop and phone), ruff, vale on the changed prose, and
`ci/test-mutation-python.sh` — **93.74 %**, threshold 80 %.

## One-way doors

- **`source.declarative_companion.paused`** — a boolean marker on a managed task's
  provenance block, written when the recipe that owns the task is switched off and
  cleared when it comes back. It is stored, and it travels in an export like the rest
  of `source`. Nothing outside Home Keeper is expected to write it, but an automation
  reading a task's `source` will see it.
- **`task_template.category` / `task_template.priority`** stop being part of the
  `add_declarative_companion` payload. A caller that sends them is not refused; the
  values are dropped, as they effectively were before.
- **A usage-mode `sensor` binding now refuses a truthy `for_seconds` or
  `clear_on_recover`.** A service call that used to be accepted (and silently
  stripped) now returns an error.

Two of these are behaviour changes rather than pure fixes — a disabled recipe keeping
its tasks, and the usage binding refusing what it used to strip. Both are easy to drop
from this PR if you would rather they landed on their own.

## Checklist

- [x] Tests run locally (`pytest tests/unit -v`, integration, frontend, mutation)
- [x] `CHANGELOG.md` updated with `(Fixes #336)`; version bumped to `0.24.0b5`
      because `0.24.0b4` already shipped
- [x] Panel UI unchanged — no screenshots needed
- [x] No new user-facing UI surface — the walkthrough tour is unchanged
- [x] Conventions written into `.amazonq/rules/architecture-and-code.md`
- [x] Bug fixes only — no `preview-release` label

https://claude.ai/code/session_017WaVi7EecopC2ogFKnyZEW

### DIFF
--- a/.amazonq/rules/architecture-and-code.md
+++ b/.amazonq/rules/architecture-and-code.md
@@ -196,6 +196,43 @@ command for admins; Home Keeper follows that rather than inventing a weaker line
   unrelated reload. Which tasks even have an edge to skip is the pure
   `sensor_tasks.holds_edge_state` — a `usage` meter is still anchored, because its
   baseline is a persisted reading and not an edge.
+- **A hold measures unbroken truth, and carried edge state belongs to one condition.**
+  Two rules the watcher and the pure evaluators share, both learned from #336:
+  - An **indeterminate reading** — missing / `unavailable` / `unknown`, and `missing`
+    for the availability mode — decides nothing *and* ends a pending hold
+    (`sensor_tasks._evaluate_indeterminate`). Carrying the crossing through a blackout
+    banked time the condition was not true, so a device that dropped off the network
+    armed its task the moment it came back in the trigger state. A task with no pending
+    crossing keeps its carried `condition_met`, which is what still stops a dropout
+    from re-arming a task the user dealt with, or clearing a `clear_on_recover` one.
+    The watcher therefore **routes a missing reading through the evaluator** rather than
+    skipping the task: skipping left the crossing and its hold timer in place.
+  - The carried edge is stamped with `sensor_tasks.condition_fingerprint(task)` — the
+    entity, attribute, mode and condition, and deliberately not the hold or
+    `clear_on_recover`. When the fingerprint changes the state is retired, so editing a
+    task (or the recipe that owns it) reads the entity against the new condition and
+    arms on a standing match instead of waiting for it to recur.
+- **The reconciler owns every key on a materialized task's `sensor` block except the
+  meter anchor.** `sensor.baseline` is written by the watcher, not by any recipe, so
+  `declarative_companions.merge_sensor_binding` carries a stored one forward and the
+  recipe's own `baseline` only seeds a task that has none. Rewriting the block
+  wholesale reset a `usage` recipe's meter on every registry event, which no user could
+  see and which no target could survive. A stored baseline is carried only while both
+  bindings stay in `usage` mode — it is a usage-only field, so it would fail validation
+  in any other.
+- **A disabled recipe pauses its tasks; it does not delete them.**
+  `declarative_companions.pause_spec_tasks` switches each task off and marks its
+  provenance `paused`; the enabled pass clears the marker and switches the task back on
+  with a `"resumed"` op, which the store reports as a freshly-made id so the watcher
+  arms it on a condition that became true while the recipe was off. Deleting them threw
+  away the completions recorded on each one, and a task switched off **by hand** carries
+  no marker, so that choice survives the recipe coming back.
+- **Every registry event goes through the reconcile debouncer.** Home Assistant fires
+  one entity-registry event per entity, and a pass walks every spec over every entity,
+  renders Jinja per match and writes the store, so an integration loading 50 entities
+  once cost 50 full passes. `DeclarativeCompanionSync._reconcile_debouncer`
+  (`immediate=True`, `RECONCILE_DEBOUNCE_SECONDS`) keeps the first pass prompt and folds
+  the rest of a burst into one trailing pass; it is shut down with the listeners.
 - **A declarative companion's notes are re-rendered when the task arms.** The reconcile
   pass renders name/notes from live state, but it runs on *registry* changes, so a
   template that quotes the reading (`{{ state }} h left`) froze at whatever the entity
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,7 +9,12 @@ on:
 jobs:
   integration:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    # The tier runs a real Home Assistant container and several tests that wait out
+    # a hold or a coordinator tick, so it takes ~9.5 minutes at main HEAD. A 10
+    # minute cap cancelled a run whose 216 tests had all passed, and a cancelled run
+    # reads as a red check. 20 gives the tier room to grow before the cap is the
+    # thing that fails a PR again.
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v7
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ All notable changes to Home Keeper are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/) and the project uses semantic
 versioning, with PEP 440 pre-release suffixes (`bN`/`aN`/`rcN`) for betas.
 
+## [0.24.0b5]
+
+### Fixed
+
+- **Sensor hold time.** A hold now counts only the time an entity reported the
+  trigger condition. An entity that stopped reporting part way through kept its
+  hold running, so the task opened as soon as the entity came back. (Fixes #336)
+
 ## [0.24.0b4]
 
 ### Fixed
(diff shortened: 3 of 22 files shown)


prestomation:
/q review Be critical and skeptical, and rank the most serious problems first — do not withhold minor ones.

This changes when a `for_seconds` hold on a sensor task completes. A reading that says nothing (missing / `unavailable` / `unknown` entity, or `missing` availability status) now ends a pending hold instead of carrying the crossing through the gap.

Please scrutinise:

1. **Correctness of the new rule.** `_evaluate_indeterminate` drops an unconsumed crossing and sets `condition_met=False`, but keeps a carried `condition_met` when there is no crossing. Is that split right in every combination of armed / dormant / baselined-met / `clear_on_recover`? Can it fabricate an arm, lose an arm forever, or complete a task on a dropout? Look hard at the `availability` mode, where "no reading" is the arm signal and only `missing` is indeterminate.
2. **The watcher change.** `evaluate_threshold` now takes `reading=None` rather than the watcher skipping the task. Does routing a missing reading through `_apply_edge` change anything else — the hold timer bookkeeping in `_schedule_hold`, the arm-time notes refresh, the usage path, or the count of coordinator refreshes?
3. **Races and timers.** `async_evaluate` awaits inside its loop and runs from 3 places (state change, hold timer, coordinator tick). Can 2 overlapping passes double-arm a task or leave a stale timer?
4. **Declarative companions.** These materialize ordinary sensor tasks, and the reconcile pass rewrites the whole `sensor` block from the recipe. Does anything in that interaction still let a hold be skipped or a crossing be replayed?
5. **HA best practices** for `async_track_point_in_time` use and teardown, and anything in the tests that asserts the mechanism rather than the behaviour.

---

prestomation:
/q review This push adds 6 fixes from a review of the whole sensor-task and declarative-companion surface, on top of the #336 hold fix you already looked at. Be critical and skeptical, rank the most serious problems first, and do not withhold minor ones.

Please scrutinise, in this order:

1. **Pausing a disabled recipe.** `pause_spec_tasks` switches tasks off and writes `source.declarative_companion.paused`; the enabled pass clears it and emits a `"resumed"` op that the store reports as a freshly-made id, so the watcher arms a condition that became true while the recipe was off. Can that marker leak, survive a delete, be written to a task the recipe does not own, or resurrect a task a person deliberately switched off? What happens to a device-backed task's per-task entities and the reload the coordinator triggers on the enabled/disabled signature change? Does an import of an exported paused task behave?
2. **The meter anchor carry.** `merge_sensor_binding` keeps a stored `sensor.baseline` and lets the recipe's own value seed only a fresh task. Is there a case where the recipe should win — a recipe edited to a new starting reading, a mode switched away and back, an entity re-matched under a new registry id?
3. **The condition fingerprint.** Retiring carried edge state on a binding change: does it retire too much (a reconcile pass that rewrites `sensor` identically, a `for_seconds` edit) or too little? Could it turn a restart into an arm, or lose a pending hold that should have stood?
4. **The debouncer.** `immediate=True` with a 5 s cooldown on registry events and the spec-changed signal. Can a trailing pass be dropped, run after the entry unloads, or race the entry reload that an entity-set change schedules? Is `async_shutdown` on `async_on_unload` enough?
5. **The tightened usage validation** and the dropped `task_template.category` / `priority`: any stored spec or task shape that now fails to load, or any caller this breaks that I have not called out as a one-way door in the PR body?

Also: HA best practices for `Debouncer` use, and anything in the new tests that asserts the mechanism rather than the behaviour.

---

PR #339, opened 2026-09-12
Appliance report: rename the inventory export and move it beside the backup

prestomation:
## What changed

Home Keeper had 2 exports and nothing that told a user them apart.

`export_data` (#309) writes the portable YAML document. `export_inventory` predates it and writes a CSV for insurance. An audit of the 2 confirms the older one holds **no data the newer one lacks**: both read the same `coord.store.list_assets()`, and `transfer.py` excludes only `id`/`created`/`identifiers`/`connections`/`device_id`/`task_history` from an asset and the 3 file-blob keys from a part. Every raw field the CSV prints already travels in the portable document.

What the older export adds is **format and arithmetic**, not facts:

- a CSV a spreadsheet opens, with formula injection neutralized by `_cell()`
- derived money: per-asset `spares_value` (each part's `cost x stock`), `part_count`, and the totals block. `export_data` emits the inputs and never multiplies them.
- a narrowed, localized column set, and the `Details` column, which prints custom fields as `label: value` in one cell

So it earns its place as a **report** beside the **backup**. This renames it to **Appliance report**, moves its button into Settings → Import and export, and gives it a blurb that states the difference.

### Desktop

Settings, Import and export, with the Appliance report section and its button below Save to a file and Read from a file

The Appliances tab control row, with Add appliance now closing the row

### Phone

The same card at phone width, with the Appliance report button on a row of its own

The Appliances tab at phone width, with only the floating Add appliance button

The phone layout is the better argument for the move. Below 700px the spacer collapses and Add floats away as a FAB, which left `#export-btn` as the only thing inline at the end of a wrapped control row, with no `--hk-tap` rule of its own.

## One-way doors

Clean break, no alias, as agreed.

| Surface | Was | Is |
| --- | --- | --- |
| Service | `home_keeper.export_inventory` | `home_keeper.export_appliance_report` |
| Websocket | `home_keeper/export_inventory` | `home_keeper/export_appliance_report` |
| Response key | `inventory` | `report` |
| Download | `home-keeper-inventory-<date>.csv` | `home-keeper-appliance-report-<date>.csv` |

Any automation, script, or voice intent calling the old service breaks on upgrade with no deprecation window. The CHANGELOG bullet says so. The response key is the half integrators actually parse, so it moves with the name rather than leaving the rename half done.

Internal names follow (`inventory.py` → `appliance_report.py`, `build_inventory` → `build_report`, `inventory_to_csv` → `report_to_csv`, `inventory.csv.*` → `report.csv.*`, `Inventory`/`InventoryRow` → `ApplianceReport`/`ApplianceReportRow`).

## Also fixed, found on the way

- **`services.yaml` promised fields the report stopped printing.** It advertised "purchase + warranty dates", which became free-form metadata and now appear in the Details column.
- **`error.exportFailed` said "Inventory export failed" in all 16 locales** and is shared by *both* exports, so it was already wrong for the YAML one. Now neutral.
- **`ApplianceReportRow` had no `serial_number`**, though `build_report` has emitted one for some time.
- **The guide documented the report on the wrong page.** `import-export.md` never mentioned it; the one paragraph describing it was welded to the end of an appliance-detail-page section.

## Mutation gate: 80% required, 99.07% achieved

The rename put `build_report`/`report_to_csv` in scope. It passed at 95.26% first time, but the survivors named real assertion gaps rather than noise, so they are fixed:

- the case-insensitive sort used `"zebra"`/`"Apple"`, which sort identically with or without the `.lower()`
- nothing distinguished `round(total_cost, 2)` from any other precision, because every cost in the file was already whole cents
- `report_to_csv` defaulted `assets` and `totals` and nothing handed it a report missing either
- the TOTAL row's blank cells were never asserted
- the localization test used Spanish, which spells TOTAL the same as English, so it could not tell a localized total row from one resolved against the wrong language. German spells both apart.
- **the formula-injection guard names 6 lead characters and asserted on 2**, and nothing asserted the other half at all: that an ordinary cell is left alone. A widened set would put an apostrophe in front of every model number starting with the added letter.

Each was confirmed by mutating the line and watching the test go red. The 2 remaining survivors are the `lang="en"` default, which cannot be killed because `resolve_string` falls back to English for any language it cannot resolve; annotated in place. Worth recording: `# pragma: no mutate` suppresses a mutant on a statement but **not** on an argument default.

**The gate runs as `.github/workflows/mutation.yml`, whose jobs are named `Python` and `TypeScript`** — there is no check called "mutation" in the list. Worth knowing, because scanning for the workflow name finds nothing and reads exactly like a gate that silently skipped.

## Verification

- `pytest tests/unit` — 2475 passed
- `bash ci/test-frontend.sh` — 1484 passed, including the 63 i18n parity checks
- `bash ci/test-mutation-python.sh` — 99.07%, 214 mutants scored
- `ruff check` / `ruff format --check` clean, `tsc --noEmit` clean
- `ci/e2e-up.sh` — 342 e2e passed
- `ci/capture-video.sh` — both tours clean, desktop 3.4m against the 360s cap
- vale — hit set compared against `origin/main` per AGENTS.md: 78 before, 78 after, **0 new**
- docusaurus build — 0 broken anchors, so `#appliance-report` resolves

## Note: an unrelated defect surfaced

Hypothesis found a **pre-existing** bug in `recurrence.py` with a fresh seed, unrelated to this change and reproducible on `main`:

```
anchor 2024-09-29T03:00 Pacific/Chatham, DAILY/1
  -> 2024-09-29T04:00+13:45  local 04:00
  -> 2024-09-30T04:00+13:45  local 04:00
```

A daily task anchored at 03:00 on Chatham's DST-start date moves to 04:00 and stays there for every later occurrence. CI will not see it (the `ci` profile derandomizes and `.hypothesis/` is not cached), and it is out of scope here, so it is left for its own change.

## Checklist

- [x] `CHANGELOG.md` updated — no `(Fixes #N)`, this fixes no filed issue
- [x] Panel UI changed → current screenshots committed under `docs/images/` and embedded above with an HTML `` tag, desktop and phone for both changed surfaces
- [x] New user-facing UI surface → not applicable. This moves and renames an existing surface, and the tour already stops at the transfer card, so it picks the new row up with no new beat. Verified by pulling a frame from the published tour: it renders the Appliance report section and its blurb.
- [x] New user-facing feature → bumped to `0.24.0b6` (`v0.24.0b5` is already a published tag, so `changelog-release-gap` required it), and the `preview-release` label is applied.

https://claude.ai/code/session_018phRDVTZ2KdntLCfdKMYk6


prestomation:
/q review Please be critical and skeptical, and rank the most serious issues first without withholding the minor ones. This is a deliberate breaking rename with no alias, so I especially want scrutiny on:

**Correctness of the rename's completeness.** `home_keeper.export_inventory` → `export_appliance_report`, the websocket command, the response key `inventory` → `report`, the module and its 2 functions, the `inventory.csv.*` backend string keys, and the TypeScript types. Is there a call site, a docstring, a test, a rule file, or a generated artifact that still names the old thing? Is there any runtime path that reads the old response key? Check `api_surface.py` against the actual registrations, since `test_api_surface.py` parses source and I want to know if it can be fooled.

**Privilege model.** The report exposes every serial number, purchase cost and value total. `docs/SECURITY.md` calls this out and the gate must hold in *both* the service and the websocket twin, or `call_service` walks around it. Confirm both still carry their admin check after the rename, and that `card_projection` still withholds the same fields.

**Is keeping this feature the right call at all?** The PR argues the report holds no field `export_data` lacks, and earns its place only through the arithmetic (`spares_value`, the totals) and the CSV shape. Push back on that reasoning if it does not hold — in particular, check whether `transfer.py` really exports every field `build_report` reads, since that claim is the whole justification.

**Maintainability.** Is the 3-section transfer card the right home for this, or does one card now do 2 unrelated jobs? Is the blurb carrying weight the UI should carry structurally?

**Test quality, not just coverage.** The mutation gate went 95.26% → 99.07% by fixing assertion gaps the survivors named (a case-fold test that would pass unfolded, an unpinned `round(…, 2)`, a localization test using a language that spells TOTAL like English, and a formula-injection guard asserting 2 of its 6 lead characters with nothing asserting ordinary text is left alone). Are those fixes actually pinning the behaviour, or did I just make the mutants go away? Are there gaps left that mutmut cannot see?

**HA best practices.** A response-only admin service, a breaking service rename with no deprecation path, and `strings.json`/`services.yaml`/`translations` parity across 16 locales.

---

prestomation:
Thanks — I checked all 6 against the tree. **1 was worth a change and is pushed (cf095c60). 4 are already satisfied. 1 is a decision the maintainer already made.** Detail below, because several of these were checked against the diff rather than the repository.

### 1. "The test file wasn't renamed" — not the case

`tests/unit/test_inventory.py` does not exist. It was moved with `git mv` in the same commit, which is why the diff shows the new path:

```
$ git ls-files tests/unit/ | grep -i 'inventory\|appliance_report'
tests/unit/test_appliance_report.py
```

`git show --numstat --find-renames 95dc5895` records it as a rename, not an add:

```
30  18  custom_components/home_keeper/{inventory.py => appliance_report.py}
45  28  tests/unit/{test_inventory.py => test_appliance_report.py}
```

That mattered for more than tidiness: `ci/mutation_scope.py` runs `git diff --diff-filter=ACMR` and keys off the new path, so a rename git could not detect would have put every function in the file into mutation scope.

`git grep export_inventory` now returns only `CHANGELOG.md` (the new bullet, which names the old action deliberately so upgraders can search for it, plus historical entries) and `docs/REVIEW_FIX_PLAN.md`, a scratch plan doc recording a past review.

### 2. "Integration test coverage missing" — it is there, and it is the REST path

`tests/integration/test_admin_gates.py` covers the renamed service on both sides, and `_call` posts to `/api/services/home_keeper/<service>` — the REST service path, not the websocket:

```python
("export_appliance_report", {}),      # in test_admin_only_services_refuse_a_non_admin
...
def test_export_appliance_report_still_works_for_an_admin(ha):
    resp = call_service(ha, "home_keeper", "export_appliance_report", {}, return_response=True)
    payload = resp.get("service_response", resp)
    assert "report" in payload and "csv" in payload
```

So a non-admin gets 401 and an admin gets the report, asserted against a real HA container.

**"Does `test_api_surface.py` actually catch a websocket missing `@require_admin`?"** Yes, and it catches the service side too. Two separate guards, both **bidirectional** — they fail if the gate is missing *and* if the gate exists but the model doesn't declare it:

- `test_admin_only_services_verify_admin` parses each handler body for `_verify_admin`
- `test_websocket_admin_flags_match_source` parses each command for `require_admin`

### 3. "Prove `call_service` cannot bypass `@require_admin`" — this has it backwards

`call_service` **does** bypass the websocket decorator. That is exactly why the service handler carries its own gate. The docstring on `_verify_admin` says so in the code:

> The websocket twins of these services are `@websocket_api.require_admin`, and a service call is the same operation over the same authenticated connection — **without this, `call_service` is a hole straight through that decorator.**

So there are 2 independent gates, not 1 decorator being trusted from 2 directions — and `test_admin_gates.py` above exercises the service one over REST, which is the hole you are asking about. Both remain in place after the rename, as you confirmed.

### 4. The `Details` column is a transformation — fair, and now documented ✅

This is the one worth a change. `_metadata_details` flattens custom fields into `label: value; label: value`, and `export_data` does not do that, so "holds no field the export leaves out" was true of the *fields* and quiet about the *shape*. Pushed in cf095c60 — the guide now says what each side does and why:

> The export keeps each custom field separate, which is what an import needs to read them back. The report prints them in one cell, because a spreadsheet column cannot hold a list.

I'd argue this strengthens rather than weakens the justification: transforming for a reader is what makes it a report rather than a second backup. A user who wants the raw fields uses the export; that is the split.

### 5. "Did mutmut re-run and confirm death?" — yes, 3 times

Not asserted, measured. The gate ran after each round: **95.26% → 98.58% → 99.07%** (212/214 killed). The formula-injection fix is visible in that progression — `x__cell__mutmut_8` (which widens the lead-character set) survived my first attempt because none of my ordinary-text cases started with the added letter, so I added the fixture's own model number `XE50` and re-ran to confirm it died.

**On the case-fold mutant: it is not a survivor.** It is suppressed by `# pragma: no mutate` and does not appear in the final list. The behaviour is tested — `test_rows_sorted_by_name_case_insensitive`, which I rewrote from `"zebra"`/`"Apple"` to `"apple"`/`"Zebra"` precisely because the original pair sorts identically with or without the fold. The 2 remaining survivors are both the `lang="en"` default, unkillable because `resolve_string` falls back to English for any language it cannot resolve — and `test_csv_headers_and_total_are_localized` proves that fallback directly.

### 6. The 3-section card — the maintainer chose this shape

You are right that these serve different audiences, and that was weighed before implementation: a separate settings section was one of the options considered and the maintainer picked the row-inside-the-card. The reasoning is that a user looking for "give me a file of my data" looks in one place, and the difference between the 2 only reads when they are adjacent — a separate Reports area would restore exactly the discoverability problem this change set out to fix, since the button's old home on the Appliances tab was already "somewhere else".

Not mine to overturn on a review comment, so I'll leave it. If the maintainer wants to revisit, a Reports area is a reasonable follow-up once there is a second report to put in it.

---

On the "more than 30 files" warning: 101 of the 195 are recaptured screenshots. The panel renders its own version in the settings rail, so the bump to `0.24.0b6` changes every shot. The reviewable surface is much smaller than the file count suggests.

---

prestomation:
/q review Narrow scope this time — `cf095c60` is a docs-only commit, one paragraph in `docs/guide/automation/import-export.md`, written in response to your previous finding that `_metadata_details` flattens custom fields into a single cell.

Please be critical and skeptical, and check only:

**Is the new paragraph accurate?** Read `build_report` and `_metadata_details` in `custom_components/home_keeper/appliance_report.py` against `_asset_out` and `EXCLUDED_ASSET_KEYS` in `transfer.py`. The paragraph now claims 3 things:
1. the export holds every custom field as a separate label and value
2. the report prints them as `label: value; label: value` in one cell
3. the export's separate form is what an import reads back

**Does it now overclaim in the other direction?** I was careful not to trade one inaccuracy for another. In particular: is "a spreadsheet column cannot hold a list" a fair statement of *why*, or is it a rationalisation of an implementation choice that could have been made differently (one column per custom field, say)? If the real reason is that the field set is user-defined and varies per appliance, the doc should say that instead.

**Any other place the same claim is now stale.** `docs/guide/appliances/appliances.md`, `docs/DESIGN.md`, the `services.yaml` description for `export_appliance_report`, and the module docstring all describe what the report holds. If any of them still implies the report and the export carry identical shapes, name it.

Do not re-review the rename mechanics — that was your previous pass, and I replied to all 6 items in the comment above this one.

---

PR #347, opened 2026-09-20
Writing style: cut the empty subject and verb

prestomation:
### 1. A new writing-style rule

One rule in `.amazonq/rules/writing-style.md`, under Tone:

> **Cut the empty subject and verb. Write the fact as a phrase.** Write "Due every 6 months", not "It comes due every 6 months". Write "Two tasks", not "This creates two tasks". Write "No spare in stock", not "There is no spare in stock". A subject that names nothing, such as "it", "this", or "there", adds a word and no fact.

The rule names its own limit. It covers UI text, help text, and a short line that states a value. A full sentence stays correct in a paragraph of documentation, where the subject names a real thing. The checklist gets a matching step.

The rule came out of a draft for a preview box on appliance wear items. Every generated line read as a narrated sentence.

### 2. The one shipped string that broke it

`form.summary.label` moves from "When it comes due" to "When due". Only `en.json` changes. The rule is an English rule, and the 16 translations are already terse or idiomatic in their own languages. `i18n-parity.test.js` checks keys and untranslated leaks, so nothing breaks.

`docs/guide/tasks/sensor-tasks.md` follows the label. The CHANGELOG keeps the old wording, because a released entry records what shipped.

### 3. Screenshots for the changed surface

No shot framed the rule box. Every task-form capture centres on a field, which leaves the box below the fold, so this label had no coverage in the corpus. Two steps are added, and both assert the label as well as photograph it.

**Desktop**

The task form with the rule box centred, labelled WHEN DUE, reading Every 100 of use, or every 6 months

**Phone**

The task form on a phone with the rule box labelled WHEN DUE, reading Every month after completion

The other 87 shots are left alone. A capture run rewrites every one of them, but the differences are the seeded dates and the notification counts, not the label.

### 4. The plan for the change this came out of

`docs/WEAR_ITEM_PREVIEW_PLAN.md` holds the implementation plan for the wear item preview. No code for it here.

## One-way doors

**None.** A rules file and a panel label. No service, event, entity attribute, storage field or websocket command changes. The label is read by people, never by an automation.

## Notes for the reviewer

Two things found while running the gates, neither caused by this change:

1. **The capture harness fails at the Duplicate drawer** (`screenshots.capture.ts:199`) on a container that is not freshly wiped. It fails the same way on `main`'s own files. A full `down -v` plus a clean config directory gets a pass in about 2 minutes.
2. **Committed shots render task rows one character per line.** `docs/images/30b-panel-sensor-backstop.png` on `main` shows a vertical column of letters where each task name should be, and a fresh capture reproduces it. Several shots in the corpus have it. Worth its own issue.

## Checklist

- [x] Tests run locally: `bash ci/test-frontend.sh` (1484 passed), full local e2e (342 passed), `vale` on the changed prose
- [x] `CHANGELOG.md` updated — not needed, no user-visible behaviour changes
- [x] Panel UI changed → screenshots committed and embedded above, both widths
- [x] New user-facing feature → none

https://claude.ai/code/session_01EHeuTg5LMDbwR1UcQG5iJ4


prestomation:
/q review Be critical and skeptical of this one new writing-style rule. Do not withhold minor findings, and put the most serious first.

- **Correctness of the rule itself.** Does "cut the empty subject and verb" conflict with any rule already in `.amazonq/rules/writing-style.md`? The file says "Do not omit articles" and "Use the simple present tense for descriptions", and a bare phrase such as "Due every 6 months" has no verb at all. Is the stated limit ("a full sentence is still correct in a paragraph of documentation") a clear enough line, or will a reader over-apply it to `README.md` and `docs/guide/**`?
- **Conflict with ASD-STE100.** STE100 requires a complete sentence with a subject and a verb in procedural text. Does this rule break that where it matters, and should it be scoped to UI strings and help text only?
- **Conflict with the vale `ai-tells` style.** Does a stack of short verbless lines trip any rule in `styles/ai-tells/` — `StackedAnaphora` on repeated "Due …" openings, for example?
- **Enforceability.** The rule has no Vale token. Is a review-only rule worth adding, or does it rot? If it is worth a token in `styles/HomeKeeper/`, what regex would catch the real cases without firing on legitimate prose?
- **Existing corpus.** Does the rule invalidate text already shipped in `strings.json`, `services.yaml` or `locales/en.json`, and if so is a follow-up cleanup needed?
- **Maintainability.** Is the rule in the right section (Tone, next to "Do not write for effect"), or does it belong under Sentences?

---

prestomation:
/q review A second push landed. Be critical and skeptical, most serious first, and do not withhold minor findings.

What changed: the shipped label `form.summary.label` moves from "When it comes due" to "When due", a phone capture step is added for that box, and `docs/WEAR_ITEM_PREVIEW_PLAN.md` is added.

- **Translation policy.** Only `en.json` changed. The reasoning: the new rule is an English STE rule, and the 16 translations are already terse or idiomatic in their own languages ("Quando vence", 何时到期). Is that right, or do "Wann sie fällig wird" and "Quand elle sera due" now say something the English label no longer says? Name any locale whose value is now out of step with the English one in register or length.
- **Missed references.** Does any other file still say "When it comes due" where it names the live label rather than a released entry? I left the CHANGELOG line at 1182 alone on purpose, and updated `docs/guide/tasks/sensor-tasks.md`. Check the alt text of committed screenshots and any test fixture.
- **Screenshot staleness.** Several committed PNGs show the old label (30b, 31, 43). The capture run that refreshes them is pending. Is a per-shot refresh enough, or does the guide text around those images need a look as well?
- **The new capture step.** It opens the form with `#add-btn`, asserts `.hk-form-summary-label`, then cancels with `#f-cancel`. Does it leave the phone walk in the state the next step expects? Is asserting a 2-word label worth the coupling, or should the assertion live in a `tests/e2e/tests/` spec instead of the capture harness?
- **Is a CHANGELOG entry owed?** A shipped UI label changed. I judged it too small for a bullet under the three-sentence budget. Say if that is wrong.
- **The plan document.** Read `docs/WEAR_ITEM_PREVIEW_PLAN.md` as a reviewer, not as prose. The parity test in section 5 is the part I least trust. Is reading `const.ACTION_TASK_NAME_TEMPLATES["en"]` against `locales/en.json` the right shape, or should the panel take the name from the backend at runtime so no second copy exists? Name anything the plan misses: a gate, a test lane, a surface it does not mention.

---

prestomation:
/q review Third push, and the PR now has its final shape. Review the whole diff at `e307185`, most serious first, and do not withhold minor findings. Read the diff itself rather than the PR body.

Note for accuracy: an earlier answer in this thread said "the wear item preview implementation follows the plan". There is no implementation in this PR. It carries a style rule, a label rename, 2 screenshots and a plan document. Please re-check against the real diff.

- **The capture steps.** `screenshots.capture.ts` gains 2 steps. The desktop one at the sensor-backstop block asserts `.hk-form-summary-label` has text "When due", then centres the box. The phone one opens the form with `#add-btn` and cancels with `#f-cancel`. Does either leave the walk in a state the next step does not expect? The desktop one runs before `setBackstop(panel, false)`, so check that the extra assertion and scroll do not disturb the threshold shot that follows.
- **Asserting a label in a capture harness.** The harness is for documentation, and the repo's own rule says a screenshot is not coverage. Is a `toHaveText('When due')` assertion in 2 capture steps the right place, or does it belong in `tests/e2e/tests/` where a failure names a behaviour rather than a missing photo? Argue it either way, but pick one.
- **Only 2 screenshots refreshed.** A capture run rewrites all 89. I committed the 2 new ones and reverted the other 87, on the grounds that their deltas are seeded dates and notification counts rather than the label. Verify that claim: is there any committed shot that frames the rule box and now shows a stale label?
- **The plan document.** Section 5 proposes `tests/unit/test_part_preview_parity.py`, reading `const.ACTION_TASK_NAME_TEMPLATES["en"]` against `locales/en.json`, so the panel's copy of the task-name templates cannot drift from the backend's. Is a parity test the right answer, or should the panel read the name from the backend so no second copy exists at all? What does the plan miss?
- **The rule's reach.** Does the new rule, as written, invalidate any line in `docs/WEAR_ITEM_PREVIEW_PLAN.md` itself, or in the PR's own prose?

---

prestomation:
Checked this against the file, and the finding does not hold.

**"The plan contains several instances of 'it comes due'"** — it contains none. The grep over `docs/WEAR_ITEM_PREVIEW_PLAN.md` for that phrasing returns nothing. The 8 lines that use "it" all have a real referent in a documentation paragraph: the box, the builder function, the form, the guide section. That is the case the rule explicitly permits. So there is no follow-up commit to make.

Two further notes, so the thread does not mislead a later reader:

1. This review answers 3 questions that were not asked. My comment asked about the 2 new capture steps and whether they disturb the threshold shot, about whether a `toHaveText` assertion belongs in a capture harness or in `tests/e2e/tests/`, about the decision to refresh 2 screenshots and revert 87, and about the parity test in section 5 of the plan. None of those is addressed.
2. An earlier review in this thread said "the wear item preview implementation follows the plan in all checked areas". No implementation exists in this PR. The diff is a style rule, a label rename, 2 screenshots and a plan document.

The checklist renumbering is right, which is the one point above that matches the diff: 6 items became 7, the new step is 5, and the old 5 and 6 became 6 and 7.

Separately, the plan is a scratch `*_PLAN.md`. `AGENTS.md` holds user-facing text to the style rules and names `CHANGELOG.md`, `README.md`, `docs/guide/**`, the canonical `docs/*.md`, `strings.json`, `services.yaml` and the frontend locale. A plan document is not in that set, and `.vale.ini` excludes it too.

---

PR #348, opened 2026-09-20
Wear item preview, and stop task names collapsing to one letter per line

prestomation:
Implements `docs/WEAR_ITEM_PREVIEW_PLAN.md`, which merged in #347.

## 1. The preview

A wear item has 3 fields whose result the form never stated. **Action** picks the task name from a table of 7, **Use task name** names a second task, and **Count uses as** captions the count. A counted wear item creates **2** tasks, and nothing in the editor said so, so a user found out on save.

The foot of each wear item now carries a box titled **What this creates**. It names each task, gives its schedule, and adds a quieter block for the stock a completion takes.

**Desktop**

The part editor for a counted wear item, with a What this creates box naming the use task and the renewal task

**Phone**

The same box on a phone, under the Attach file button and above Remove part

`partPreview` is pure and takes the appliance name as an argument, so it stays on the mutation-tested surface in `forms.ts`. The box redraws from the same `merge()` the collapsed row summary already uses, so no keystroke calls `_render()`.

The old wear hint is gone. It said a wear item creates a maintenance task without naming it, and never mentioned the second task.

## 2. Task names no longer collapse to one letter per line

Found while capturing the screenshots above, and present in the committed ones on `main` — I flagged it in #347 and this fixes it.

Above 1151px a task row is a grid, and its tracks were a zero-floor name track beside a fixed chip track. That holds while the panel has the window. Open the drawer and the panel is ~515px against ~728px of tracks, so the whole shortfall came off the only track that could give. The name fell to **17px wide and 128px tall**: every letter of "Wear rain jacket" on its own line.

The name track now has a `6rem` floor and the chip track absorbs the squeeze, which is what the rule below it already says should happen. The text uses `break-word`, which wraps identically to `anywhere` but leaves min-content at the longest word rather than at one character.

## 3. One design change the plan got wrong

The plan specified a line reading `Count reads 0 of 20 wears.` The capture put it beside a part row reading **17 of 25 wears**. The box previews what a part creates, so a fixed `0` is false on any part that has been used, and threading the live count in would make a pure builder depend on the task list. The noun it carried is already in the due line, so the line is dropped. `docs/WEAR_ITEM_PREVIEW_PLAN.md` records this and 3 smaller corrections.

## One-way doors

**None.** The box is panel text built from fields that already exist. No service, event, entity attribute, storage field or websocket command changes.

Two internal surfaces do change, and both are reversible:

- **The task-name templates now exist twice**, in `const.ACTION_TASK_NAME_TEMPLATES` and as `part.taskName.*` locale keys. Only the browser knows the viewer's language and the preview redraws per keystroke, so the panel cannot ask the backend. The locale values are **copied from the backend tables**, not translated by hand, and `tests/unit/test_part_preview_parity.py` fails if the two ever disagree — every action, all 16 languages, placeholders included.
- **`part.wearHint` is removed** from all 16 locales. No automation reads a locale string.

## Testing

| Lane | Result |
| --- | --- |
| `ci/test-frontend.sh` | 1528 passed |
| `pytest tests/unit` | 2644 passed, 1 skipped, 1 xfailed |
| Full local e2e | 344 of 345; the 1 failure was `settings.spec.ts` state pollution from repeated capture runs on a reused container, and it passes 9/9 on a clean one |
| `ci/test-mutation-frontend.sh` | **100.00%** (146/146), threshold 80% |
| `ci/test-mutation-python.sh` | **100.00%** (92/92), threshold 80% |
| `vale` | clean on both added blocks |

`task-name-wrap.spec.ts` measures rendered boxes with the drawer open. It fails on the old CSS at all 3 widths and passes on the new. Nothing below a browser can catch it: it is the grid algorithm resolving track sizes against content.

One mutant is annotated equivalent at the source with its reason, rather than left surviving.

## Checklist

- [x] Tests run locally — see the table above
- [x] `CHANGELOG.md` updated — `## [0.24.0b8]`, an `### Added` bullet linking the guide and a `### Fixed` bullet for the collapse
- [x] Panel UI changed → screenshots committed and embedded above, both widths, both inspected
- [x] New user-facing UI surface → `screenshots-counted.capture.ts` gains `62d` and `62e`, both asserting the task names as well as photographing them
- [x] New user-facing feature → bumped to `0.24.0b8` in `manifest.json` and `const.PANEL_VERSION`; please add the `preview-release` label
- [x] Documented in `docs/guide/appliances/appliances.md`

https://claude.ai/code/session_01EHeuTg5LMDbwR1UcQG5iJ4

---
_Generated by [Claude Code](https://claude.ai/code/session_01EHeuTg5LMDbwR1UcQG5iJ4)_


prestomation:
/q review Be critical and skeptical. Put the most serious first and do not withhold minor findings. Read the diff, not this description.

This implements the plan that merged in #347, plus a layout fix found while capturing its screenshots.

- **Correctness of `partPreview` in `forms.ts`.** It claims to predict what `reconcile.py` will create. Walk the two against each other: the gate on a truthy `replace_interval`, the action template fallback to `replace`, the use task naming the appliance and never the part, the appliance fallback word, and the part name used verbatim with no tidying. Name any input where the preview and the reconciler would disagree.
- **`partFirstDue`.** It adds days, weeks and months to a last-replaced date and clamps a short month the way `recurrence.add_months` does. Is the clamp right for every case, including a leap year and a December rollover? Should this arithmetic exist in the panel at all, given the backend owns the recurrence engine?
- **The duplicated templates.** `part.taskName.*` mirrors `const.ACTION_TASK_NAME_TEMPLATES`, guarded by `tests/unit/test_part_preview_parity.py`. Is a parity test the right guard, or is there a way to have one copy? Does the test actually fail on the drift it claims to catch?
- **The grid fix in `panel-styles.ts`.** The name track gains a `6rem` floor and the chip track becomes `minmax(0, …)`. What does that do to a very long single-word task name, to a row carrying many chips, and at exactly 1151px? Is `6rem` a defensible floor or a magic number?
- **`task-name-wrap.spec.ts`.** It asserts `height < width` per row. Is that sound for a legitimately long name that wraps to 3 lines in a narrow column, or will it go flaky on a future fixture?
- **The removed `part.wearHint`.** Its second sentence explained that the last-replaced date seeds the schedule. Is that guidance now lost to a user, and does the field's own label carry enough?
- **Security and HA practice.** The preview builds HTML with `innerHTML`. Check every interpolation is escaped — the part name, the use task name and the noun are all free user text.

---

prestomation:
Thanks — this one engaged with the diff, and I checked each answer against the code rather than taking it. I agree with all of it but point 5, which I have acted on anyway.

**Point 5, the `height < width` assertion.** The review says there is no flakiness risk and that future fixtures are fine. I do not think that holds. The assertion compares the box against the **column**, and the column now has a `6rem` floor. Where that floor binds, roughly 96px, a name of about 60 characters wraps honestly to 7 lines at 112px and fails a correct render. No current fixture reaches it, which is why it passes today, and that is exactly the kind of assertion that goes red a year from now on a change that broke nothing.

Fixed in `2f1bd69` by measuring against the text instead of the column. A letter-per-line stack is one line per character at about 16px each, so the budget is 4px per character: that permits a wrap as tight as 4 characters to the line, and still fails by 2x on the real bug, where "Wear rain jacket" measured 128px against a 64px budget.

Re-verified both directions: red on the pre-fix CSS, green on the fix.

**On point 3**, one correction to the arithmetic. The parity test is 7 actions x 16 languages = 112 cases, plus 16 use-template cases and 16 placeholder cases, not 140 in the shape described. The conclusion stands.

**On point 2**, agreed that the panel has to own the arithmetic, and worth being explicit about what guards it: the parity test covers the **names**, not the dates. What guards `partFirstDue` is its own unit tests, including the Jan 31 clamp, which is the case a naive `setMonth` gets wrong by rolling forward to Mar 3.

The `ruff` and `vale` failures in this thread were on `d59e9e0` and are fixed in `2516748`. The vale one is worth recording: the local Vale 3.9.1 reported zero hits on that file, so I checked the rewrite against every regex in `styles/ai-tells/VerbTricolon.yml` by hand instead of trusting the local binary. That discrepancy is the one `AGENTS.md` already documents.

---

PR #354, opened 2026-09-21
Update the dependencies and the checkout action

prestomation:
## What changed

This applies the open Dependabot PR, plus the dependency drift that Dependabot has
not filed yet, and closes three gaps in the maintenance setup. All of it is
developer-only, so there is no CHANGELOG entry and no version bump.

### Dependencies

| Project | Package | From | To |
| --- | --- | --- | --- |
| `custom_components/home_keeper/frontend` | rollup | 4.62.5 | 4.63.4 |
| `custom_components/home_keeper/frontend` | vitest (lock file) | 5.0.0 | 5.0.1 |
| `/` | jsdom | 30.0.1 | 30.1.0 |
| `website` | @docusaurus/* | 3.10.1 | 3.10.2 |
| `website` | react, react-dom (lock file) | 19.2.7 | 19.3.0 |

The rollup bump supersedes **#345**, which asked for 4.63.3. 4.63.4 is the newest
patch.

### Maintenance setup

- **Dependabot did not cover `/website`.** The site is the fourth npm project in the
  repository and its build is a gate on every PR, but no ecosystem entry named it, so
  nothing updated its dependencies. This adds the entry. The Docusaurus and React
  updates above are the backlog that gap created.
- **The new `/website` entry must not propose a TypeScript major.** No CI job runs the
  site's `npm run typecheck`. `docs-preview.yml` and `docs-deploy.yml` run
  `npm run build`, and a Docusaurus build does not call `tsc`. So a major TypeScript
  bump would pass every check, `dependabot-auto-merge.yml` would merge it, and no job
  would have typechecked the site. The entry carries an ignore rule for the major, and
  its comment names the condition to remove it. The rule does not block the patch
  updates that the `~5.6.2` range allows. See **Follow-up** below.
- **Three workflows still used `actions/checkout@v4`** while every other workflow used
  `@v7`: `docs-deploy.yml`, `docs-preview.yml` and `preview-release.yml`. They now use
  `@v7`.

### Left out on purpose

- **vitest 4 -> 5 at the root, and typescript 6 -> 7 in the panel frontend.** Both are
  major versions. The frontend typescript major is already in the Dependabot ignore
  list, and the root vitest major needs its own change.
- **The `vale-ai-tells` pin in `.vale.ini`** is at v1.26.0 and the newest release is
  v1.37.0. I tried the bump and measured it: the newer style adds 43 rules, and the
  hits on the current corpus go from 94 to 406. The vale gate is diff-scoped, but a
  rule applies per block, so after this bump a one-line edit can report a hit on a
  paragraph the PR did not write. `AGENTS.md` says to bump this pin with a full-corpus
  cleanup pass, so it needs its own change. I reverted it here.

## Follow-up

**No lane runs `npm run typecheck` in `website/`.** The ignore rule above is a guard,
not the fix. A lane that runs it would catch a TypeScript break and would let the
ignore rule come out. That adds a new gate, so it belongs in its own change and is the
maintainer's call.

## One-way doors

None. No service input, event payload, storage field or entity attribute changes.

## Tests

- **Root vitest**, which is what `ci/test-frontend.sh` runs: 47 files, 1533 tests, all
  pass. (Running vitest from inside the frontend directory fails, but it fails the same
  way on `main`: the fixture paths, the jsdom environment and the `panel-version` alias
  all come from the root `vitest.config.js`.)
- **rollup build**: both bundles build with 4.63.4.
- **Docusaurus build**: 3.10.2 builds every page. My local build stops at
  `ci/generate_schema.py`, which needs a Python at Home Assistant's floor that this
  machine does not have, and the one broken link it reports is the schema file that the
  skipped generator writes. CI has that Python, and `Docs site — PR preview` passed.
- `.github/dependabot.yml` parses, with 6 update entries.
- No Python file changed.

## Checklist

- [x] Tests run locally (root vitest, rollup build, Docusaurus build)
- [x] `CHANGELOG.md` updated — not needed, this change is developer-only
- [x] Panel UI changed → no panel source changed, so no screenshots
- [x] New user-facing feature → none, so no beta bump and no `preview-release` label

Supersedes #345.

https://claude.ai/code/session_01DzuKFiC429dkPeQ6DRyL7j


prestomation:
/q review Please review this maintenance change critically and skeptically. Put the most serious issues first, and do not hold back the minor ones.

Topics I want scrutinized:

- **Correctness of the dependency updates.** Does the `rollup` 4.62.5 -> 4.63.4 bump change the built bundle in any way that matters (tree shaking, output format, sourcemaps)? Is `jsdom` 30.0.1 -> 30.1.0 safe for the vitest jsdom environment the root suite uses? Do the Docusaurus 3.10.1 -> 3.10.2 and React 19.2.7 -> 19.3.0 updates break the site build, the `sync-docs.mjs` output, or the `pr-preview` deploy?
- **Lock file integrity.** The lock files were regenerated with `--package-lock-only`. Check that every lock file still agrees with its `package.json`, that no transitive dependency was dropped or downgraded, and that no unexpected package entered.
- **The `actions/checkout@v4` -> `@v7` move.** Does any of the three workflows depend on v4 behaviour? `docs-deploy.yml` passes a `ref`, `docs-preview.yml` sets `fetch-depth: 0`, and `preview-release.yml` checks out the PR head SHA. Does v7 change the default token, the submodule handling, or the credential persistence in a way that breaks those?
- **The new Dependabot entry for `/website`.** Is the directory correct, and will it conflict with the existing `/` npm entry? Should it carry an ignore rule, as the frontend entry does for typescript?
- **Anything I missed.** Is there other maintenance drift in this repository: a stale action pin, a Python dependency, or a CI configuration that no PR lane reads?


---

prestomation:
Thanks — I took the one actionable finding and pushed it in bfbf1dd.

**The `/website` ignore rule: agreed, but for a different reason.** The frontend rule exists because `@rollup/plugin-typescript@12.3.0` crashes on TypeScript 7. The site does not use rollup at all, so that reason does not transfer.

A different reason does, and it is worse. **No CI job runs the site's `npm run typecheck`.** `docs-preview.yml` runs `npm run build`, and a Docusaurus build does not call `tsc`. So a major TypeScript bump on `/website` would pass every check, `dependabot-auto-merge.yml` would merge it, and no job would have typechecked the site. The rule now blocks that, and its comment names the condition to remove it: a lane that runs `npm run typecheck` in `website/`.

Wiring that lane in is the real fix, but it adds a new gate and belongs in its own change, not in a dependency update.

**On the other points**, agreed and no action:

- The `checkout@v4` -> `@v7` read matches mine. None of the three workflows sets `token`, `submodules` or `persist-credentials`, so the defaults carry over.
- The `/website` directory does not conflict with the `/` npm entry. Dependabot scopes each directory on its own.
- The lock files: `Test` and `E2E (browser)` both passed on 6ff4993, and `Docs site — PR preview` built the whole site with Docusaurus 3.10.2 on a Python that meets Home Assistant's floor. That is the check my local run could not finish.


---

prestomation:
/q review Please review the new commit bfbf1dd critically and skeptically. It adds a TypeScript major-version ignore rule to the `/website` Dependabot entry.

Check these in particular, and put the most serious first:

- **Is my premise right?** I claim no CI job runs the site's `npm run typecheck`. Search every workflow and every `ci/*.sh` script and confirm or refute it. If some lane does typecheck the site, the rule is wrong and I will drop it.
- **Does a Docusaurus build typecheck?** I claim `npm run build` does not call `tsc`, so the site's `prebuild` sync hook and the build itself would not catch a TypeScript break. Confirm or refute.
- **Is the ignore rule the right shape?** `update-types: ["version-update:semver-major"]` on `dependency-name: "typescript"`. Does this also block the minor and patch updates that the `~5.6.2` range allows, which I do want?
- **Is the whole idea wrong?** An argument against: `dependabot-auto-merge.yml` deliberately trusts CI for any bump, majors included. Adding an ignore rule works against that policy. Is wiring `npm run typecheck` into a lane the better fix, and should this rule wait for it?
- **Anything else in the full diff** that the earlier review passes missed.


---

